PR #21, opened 2021-09-14
pem-rfc7468: rename `Error::HeaderDisallowed`

tarcieri:
It was previously `Error::HeaderDetected`, however that doesn't really indicate why it's an error.

The new error message notes that headers are explicitly disallowed in the RFC7468 interpretation of PEM.

cc @dwhjames


dwhjames (approved):
👍

PR #22, opened 2021-09-14
Add pem_rfc7468::decode_label

dwhjames:
Decode the encapsulation boundaries of a PEM document, returning the label.

The motivation is to capture the document label in the case of an error
(by repeating the decoding only to extract the label)


dwhjames:
Follow up from discussion: https://github.com/RustCrypto/formats/pull/13#issuecomment-919319293

tarcieri:
Cool, seems like that addresses your problem

PR #24, opened 2021-09-14
der: increase `Length::MAX` to 256 MiB

tarcieri:
There are legitimate cryptographic use cases that need to support ASN.1 DER-encoded documents larger than the previous maximum of 1 MiB.

Two notable applications are where messages can be quite large are:

- Certificate Revocation Lists (CRLs)
- Cryptographic Message Syntax (CMS) a.k.a. PKCS#7

This increases the maximum length to something which is hopefully still somewhat conservative but also useful enough for the above applications.

It can be further increased if there is a legitimate use case.

### DIFF
--- a/der/src/length.rs
+++ b/der/src/length.rs
@@ -7,12 +7,12 @@ use core::{
     ops::Add,
 };
 
-/// Maximum length as a `u32` (1 MiB).
-const MAX_U32: u32 = 0xf_ffff;
+/// Maximum length as a `u32` (256 MiB).
+const MAX_U32: u32 = 0xfff_ffff;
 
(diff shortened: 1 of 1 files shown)


newpavlov:
What is tradeoff of increasing this size? Maybe it's worth to make it a feature?

tarcieri:
It's a sanity check. Many ASN.1 libraries (and parsing/encoding libraries in general) have them, although it seems the maximums are usually much higher (e.g. 2GB).

`Length` is a newtype for `u32` that always performs checked arithmetic. It can theoretically be anywhere up to `u32::MAX`.

PR #26, opened 2021-09-15
Refactor Error and Result usage for pkcs5 crate

dwhjames:
- introduce Error enum with one and only existing error value
- introduce specialized Result type for crate
- rewrite references to der crate Error & Result types

### DIFF
--- a/pkcs5/src/pbes2/encryption.rs
+++ b/pkcs5/src/pbes2/encryption.rs
@@ -1,7 +1,7 @@
 //! PBES2 encryption implementation
 
 use super::{EncryptionScheme, Kdf, Parameters, Pbkdf2Params, Pbkdf2Prf, ScryptParams};
-use crate::CryptoError;
+use crate::{Error, Result};
 use block_modes::block_padding::Pkcs7;
 use block_modes::{BlockMode, Cbc};
 use core::convert::TryInto;
@@ -28,37 +28,51 @@ pub fn encrypt_in_place<'b>(
     password: impl AsRef<[u8]>,
     buffer: &'b mut [u8],
     pos: usize,
-) -> Result<&'b [u8], CryptoError> {
-    let encryption_key = EncryptionKey::derive_from_password(
-        password.as_ref(),
-        &params.kdf,
-        params.encryption.key_size(),
-    )?;
+) -> Result<&'b [u8]> {
+    let key_size = params.encryption.key_size();
+    let algo_params_invalid_error = Error::AlgorithmParametersInvalid {
+        oid: params.encryption.oid(),
+    };
+    if key_size > MAX_KEY_LEN {
+        return Err(algo_params_invalid_error);
+    }
+    let encryption_key =
+        EncryptionKey::derive_from_password(password.as_ref(), &params.kdf, key_size)?;
 
     match params.encryption {
         EncryptionScheme::Aes128Cbc { iv } => {
             let cipher = Aes128Cbc::new_from_slices(encryption_key.as_slice(), iv)
-                .map_err(|_| CryptoError)?;
-            cipher.encrypt(buffer, pos).map_err(|_| CryptoError)
+                .map_err(|_| algo_params_invalid_error)?;
+            cipher
+                .encrypt(buffer, pos)
+                .map_err(|_| Error::EncryptFailed)
         }
         EncryptionScheme::Aes192Cbc { iv } => {
             let cipher = Aes192Cbc::new_from_slices(encryption_key.as_slice(), iv)
-                .map_err(|_| CryptoError)?;
-            cipher.encrypt(buffer, pos).map_err(|_| CryptoError)
+                .map_err(|_| algo_params_invalid_error)?;
+            cipher
+                .encrypt(buffer, pos)
+                .map_err(|_| Error::EncryptFailed)
         }
         EncryptionScheme::Aes256Cbc { iv } => {
             let cipher = Aes256Cbc::new_from_slices(encryption_key.as_slice(), iv)
-                .map_err(|_| CryptoError)?;
-            cipher.encrypt(buffer, pos).map_err(|_| CryptoError)
+                .map_err(|_| algo_params_invalid_error)?;
+            cipher
+                .encrypt(buffer, pos)
+                .map_err(|_| Error::EncryptFailed)
         }
         #[cfg(feature = "3des")]
         EncryptionScheme::DesEde3Cbc { iv } => {
             let cipher = DesEde3Cbc::new_from_slices(encryption_key.as_slice(), iv)
-                .map_err(|_| CryptoError)?;
-            cipher.encrypt(buffer, pos).map_err(|_| CryptoError)
+                .map_err(|_| algo_params_invalid_error)?;
+            cipher
+                .encrypt(buffer, pos)
+                .map_err(|_| Error::EncryptFailed)
         }
         #[cfg(feature = "des-insecure")]
-        EncryptionScheme::DesCbc { .. } => Err(CryptoError),
+        EncryptionScheme::DesCbc { .. } => Err(Error::UnsupportedAlgorithm {
+            oid: super::DES_CBC_OID,
+        }),
     }
 }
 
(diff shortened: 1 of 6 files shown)


dwhjames:
@tarcieri, admittedly I’m just trying to do easy/accessible things, with learning being the primary motivation. I hope this is actually useful work, and not just busy work. It it did seem that the one error value was actually rolling up a variety of distinct error categories, so I thought I’d rewrite the one and only error, before moving on to separating out the cases.

tarcieri:
It'd make sense if `Error` had more than one variant. Perhaps you can eliminate some of the `map_err`s that are discarding error values with `From` conversions?

dwhjames:
Here were my ideas:

1. ‘encrypt/decrypt not supported for PBES1’
  - https://github.com/RustCrypto/formats/blob/a391e5b0995ae7aa9224972749d4186625b6ec5c/pkcs5/src/lib.rs#L89-L90
  - and three other locations
2. ‘invalid scrypt params’
  - https://github.com/RustCrypto/formats/blob/a391e5b0995ae7aa9224972749d4186625b6ec5c/pkcs5/src/pbes2/kdf.rs#L364-L365
  - https://github.com/RustCrypto/formats/blob/a391e5b0995ae7aa9224972749d4186625b6ec5c/pkcs5/src/pbes2/kdf.rs#L429-L438
3. ‘invalid block cipher params’
  - https://github.com/RustCrypto/formats/blob/a391e5b0995ae7aa9224972749d4186625b6ec5c/pkcs5/src/pbes2/encryption.rs#L40-L41
  - and several other locations
  - one could argue that these could be panic’s at this point and that there could be validation at an earlier point
 4. ‘no buffer remaining for block cipher padding during encryption’
   - https://github.com/RustCrypto/formats/blob/a391e5b0995ae7aa9224972749d4186625b6ec5c/pkcs5/src/pbes2/encryption.rs#L42
   - and several other locations
5. ‘decryption error - invalid ciphertext size or malformed padding’
  - https://github.com/RustCrypto/formats/blob/a391e5b0995ae7aa9224972749d4186625b6ec5c/pkcs5/src/pbes2/encryption.rs#L97
  - and several other locations
6. ‘encryption not supported for cipher’
  - https://github.com/RustCrypto/formats/blob/a391e5b0995ae7aa9224972749d4186625b6ec5c/pkcs5/src/pbes2/encryption.rs#L61
7. ‘key derivation with Sha1 not enabled’
  - https://github.com/RustCrypto/formats/blob/a391e5b0995ae7aa9224972749d4186625b6ec5c/pkcs5/src/pbes2/encryption.rs#L130
8. ‘scrypt derivation failure’
  - https://github.com/RustCrypto/formats/blob/a391e5b0995ae7aa9224972749d4186625b6ec5c/pkcs5/src/pbes2/encryption.rs#L187-L194
  - this also seems like a library bug thus panic, rather than a plausible runtime error — entirely prevented as long as the constants in the crate are set appropriately

jklong:
>     1. ‘decryption error - invalid ciphertext size or malformed padding’
> 
> 
>     * https://github.com/RustCrypto/formats/blob/a391e5b0995ae7aa9224972749d4186625b6ec5c/pkcs5/src/pbes2/encryption.rs#L97
> 
>     * and several other locations
>

I'd be careful with this one in particular. Once you start identifying information around padding you're introducing the possibility of a padding oracle. That's a large potential foot-gun if the library user isn't careful about that information leaking.

tarcieri:
Going through these one-by-one:

> 1. ‘encrypt/decrypt not supported for PBES1’

I think it's fine to have a variant that says PBES1 is unsupported. This will effectively be a perpetual WONTFIX.

> 2. ‘invalid scrypt params’
> 3. ‘invalid block cipher params’

I would really like to avoid overspecializing variants for specific algorithms, especially if they are variants that are only available via conditional compilation.

How about something like: `AlgorithmParametersInvalid { oid: ObjectIdentifier }`?

If you'd really like to break the cases above down, perhaps it could be `KdfParametersInvalid { oid } ` versus `CipherParametersInvalid { oid }`.

> 4. ‘no buffer remaining for block cipher padding during encryption’

An error case for the output buffer being too small sounds fine. I would avoid mentioning anything about padding: the real problem is that the buffer is too small.

> 5. ‘decryption error - invalid ciphertext size or malformed padding’

This is the kind of error I'd personally prefer to continue to keep an opaque `Crypto` error. If you really want to split out errors specifically related to decryption, something like `Error::DecryptFailed` would be OK.

> 6. ‘encryption not supported for cipher’
> 7. ‘key derivation with Sha1 not enabled’

These are effectively unsupported algorithm errors. As I mentioned earlier, I'd prefer not to get too off in the weeds with algorithm-specific error variants.

Something like `UnsupportedAlgorithm { oid: ObjectIdentifier }` seems fine to me though.

> 8. ‘scrypt derivation failure’
> this also seems like a library bug thus panic, rather than a plausible runtime error — entirely prevented as long as the constants in the crate are set appropriately

Across the RustCrypto projects we generally try to minimize any potential runtime panics, and especially in low-level libraries like this we are aiming for panic-free code. [The reason is we'd like to make these libraries usable in contexts where panics are unacceptable, like OS kernels or bare metal embedded device programming](https://internals.rust-lang.org/t/enforcing-no-std-and-no-panic-during-build/14505).

That's not a hard rule, or an easily enforceable one (see previous link for some discussion on that), but we would like to ensure that there is a baseline level of functionality which does not require a panic handler.

tarcieri:
> I'd be careful with this one in particular. Once you start identifying information around padding you're introducing the possibility of a padding oracle. That's a large potential foot-gun if the library user isn't careful about that information leaking.

Yep, exactly, avoiding this sort of leakage is exactly why we've largely kept the error types opaque.

dwhjames:
@jklong @tarcieri thanks for the great feedback.

I’ll come back with something concrete.

tarcieri (approved):
Looks good! I'll hold off on merging this until we're ready to make breaking changes

tarcieri:
@dwhjames can you rebase? If so, I can merge after that

dwhjames:
@tarcieri, 👍, conflicts resolved.

pkcs5/src/pbes2/encryption.rs
```diff
@@ -28,37 +28,51 @@ pub fn encrypt_in_place<'b>(
...
-    let encryption_key = EncryptionKey::derive_from_password(
-        password.as_ref(),
-        &params.kdf,
-        params.encryption.key_size(),
-    )?;
+) -> Result<&'b [u8]> {
+    let key_size = params.encryption.key_size();
+    let algo_params_invalid_error = Error::AlgorithmParametersInvalid {
+        oid: params.encryption.oid(),
+    };
```
tarcieri, line 35: Rather than eagerly calculating this, you could make it a function on `Parameters`, e.g. `Parameters::invalid_error`.

pkcs5/src/pbes2/encryption.rs
```diff
@@ -67,40 +81,43 @@ pub fn decrypt_in_place<'a>(
...
+) -> Result<&'a [u8]> {
     let encryption_key = EncryptionKey::derive_from_password(
         password.as_ref(),
         &params.kdf,
         params.encryption.key_size(),
     )?;
 
+    let algo_params_invalid_error = Error::AlgorithmParametersInvalid {
+        oid: params.encryption.oid(),
+    };
```
tarcieri, line 93: Ditto here re: https://github.com/RustCrypto/formats/pull/26/files#r725524106

pkcs5/src/pbes2/kdf.rs
```diff
@@ -156,13 +166,13 @@ pub struct Pbkdf2Params<'a> {
 
 impl<'a> Pbkdf2Params<'a> {
     /// Initialize PBKDF2-SHA256 with the given iteration count and salt
-    pub fn hmac_with_sha256(iteration_count: u16, salt: &'a [u8]) -> Result<Self, CryptoError> {
```
tarcieri, line 159: It would probably be good to leave this fallible.

I'm not sure why `u16` was used for the iteration count but it unnecessarily constrains the number of iterations. Using 100,000+ should be supported.

However, if this type is changed to a `u32`, it would also be good to set a sanity limit on the number of iterations.

RFC8018 says the sanity limit is implementation defined:

https://datatracker.ietf.org/doc/html/rfc8018#appendix-A.2

> iterationCount specifies the iteration count.  The maximum iteration count allowed depends on the implementation.  It is expected that implementation profiles may further constrain the bounds.

pkcs5/src/pbes2/kdf.rs
```diff
@@ -341,19 +351,19 @@ pub struct ScryptParams<'a> {
 }
 
 impl<'a> ScryptParams<'a> {
+    #[cfg(feature = "scrypt")]
+    const INVALID_ERR: Error = Error::AlgorithmParametersInvalid { oid: SCRYPT_OID };
```
tarcieri, line 355: This can also be replaced with a function on `params`.

pkcs5/src/pbes2/encryption.rs
dwhjames: see https://github.com/RustCrypto/formats/pull/26/commits/d30f931a6f65957ef9ab31e19e14e217678220cc

tarcieri:
Thank you!

PR #29, opened 2021-09-15
Allow for data before PEM (RFC 7468) encapsulation boundary

dwhjames:
[RFC 7468 § 2. General Considerations](https://datatracker.ietf.org/doc/html/rfc7468#section-2)
>  Data before the encapsulation boundaries are permitted, and parsers MUST NOT malfunction when processing such data.

fixes TODO

### DIFF
--- a/pem-rfc7468/src/decoder.rs
+++ b/pem-rfc7468/src/decoder.rs
@@ -124,20 +124,60 @@ struct Encapsulation<'a> {
 }
 
 impl<'a> Encapsulation<'a> {
+    /// Adapted from:
+    /// https://en.wikipedia.org/wiki/Boyer%E2%80%93Moore%E2%80%93Horspool_algorithm#Description
+    /// https://github.com/peterjoel/needle/blob/master/src/skip_search.rs#L36-L47 (MIT LICENSE)
+    const fn generate_bad_char_table() -> [usize; 256] {
+        let needle = PRE_ENCAPSULATION_BOUNDARY;
+        let mut table = [needle.len(); 256];
+        let mut i = 0;
+
+        while i < needle.len() - 1 {
+            let c = needle[i] as usize;
+            table[c] = needle.len() - i - 1;
+            i += 1;
+        }
+
+        table
+    }
+
+    const PRE_ENCAPSULATION_BOUNDARY_BAD_CHARS_TABLE: [usize; 256] =
+        Self::generate_bad_char_table();
+
+    /// Adapted from:
+    /// https://en.wikipedia.org/wiki/Boyer%E2%80%93Moore%E2%80%93Horspool_algorithm#Description
+    /// https://github.com/peterjoel/needle/blob/f40693aff55a932eeca16e6b921ad4619a1f3b42/src/skip_search.rs#L14-L33 (MIT LICENSE)
+    fn boyer_moore_horspool(haystack: &[u8]) -> Option<usize> {
+        let needle = PRE_ENCAPSULATION_BOUNDARY;
+        let mut position = 0;
+        let max_position = haystack.len() - needle.len();
+        let mut res = None;
+
+        'outer: while position <= max_position {
+            let mut needle_position = needle.len() - 1;
+            while haystack[position + needle_position] == needle[needle_position] {
+                if needle_position == 0 {
+                    res = Some(position);
+                    break 'outer;
+                } else {
+                    needle_position -= 1;
+                }
+            }
+            let bad_char = haystack[position + needle.len() - 1];
+            let jump = Self::PRE_ENCAPSULATION_BOUNDARY_BAD_CHARS_TABLE[bad_char as usize];
+            position += jump;
+        }
+
+        res
+    }
+
     /// Parse the type label and encapsulated text from between the
     /// pre/post-encapsulation boundaries.
-    ///
-    /// Note that the current implementation does not permit data before the
-    /// pre-encapsulation bounadry. This may technically be in violation of
-    /// RFC 7468:
-    /// > Data before the encapsulation boundaries are permitted, and
-    /// > parsers MUST NOT malfunction when processing such data.
-    // TODO(tarcieri): determine what is allowed before the pre-encapsulation boundary
     pub fn parse(data: &'a [u8]) -> Result<Self> {
-        // Parse pre-encapsulation boundary (including label)
-        let data = data
-            .strip_prefix(PRE_ENCAPSULATION_BOUNDARY)
-            .ok_or(Error::PreEncapsulationBoundary)?;
+        // search for position of PRE_ENCAPSULATION_BOUNDARY
+        let position = Self::boyer_moore_horspool(data).ok_or(Error::PreEncapsulationBoundary)?;
+        // drop everything before the expected start of the label
+        let data = &data[position + PRE_ENCAPSULATION_BOUNDARY.len()..];
 
         let (label, body) = grammar::split_label(data).ok_or(Error::Label)?;
 
--- a/pem-rfc7468/src/decoder.rs
+++ b/pem-rfc7468/src/decoder.rs
@@ -147,37 +147,56 @@ impl<'a> Encapsulation<'a> {
     /// Adapted from:
     /// https://en.wikipedia.org/wiki/Boyer%E2%80%93Moore%E2%80%93Horspool_algorithm#Description
     /// https://github.com/peterjoel/needle/blob/f40693aff55a932eeca16e6b921ad4619a1f3b42/src/skip_search.rs#L14-L33 (MIT LICENSE)
-    fn boyer_moore_horspool(haystack: &[u8]) -> Option<usize> {
+    fn boyer_moore_horspool(haystack: &[u8]) -> Result<usize> {
         let needle = PRE_ENCAPSULATION_BOUNDARY;
         let mut position = 0;
         let max_position = haystack.len() - needle.len();
-        let mut res = None;
 
-        'outer: while position <= max_position {
+        while position <= max_position {
             let mut needle_position = needle.len() - 1;
             while haystack[position + needle_position] == needle[needle_position] {
                 if needle_position == 0 {
-                    res = Some(position);
-                    break 'outer;
+                    return Ok(position);
                 } else {
                     needle_position -= 1;
                 }
             }
             let bad_char = haystack[position + needle.len() - 1];
+            if bad_char == grammar::CHAR_NUL {
+                return Err(Error::InvalidText);
+            }
             let jump = Self::PRE_ENCAPSULATION_BOUNDARY_BAD_CHARS_TABLE[bad_char as usize];
             position += jump;
         }
 
-        res
+        Err(Error::PreEncapsulationBoundary)
     }
 
     /// Parse the type label and encapsulated text from between the
     /// pre/post-encapsulation boundaries.
     pub fn parse(data: &'a [u8]) -> Result<Self> {
-        // search for position of PRE_ENCAPSULATION_BOUNDARY
-        let position = Self::boyer_moore_horspool(data).ok_or(Error::PreEncapsulationBoundary)?;
-        // drop everything before the expected start of the label
-        let data = &data[position + PRE_ENCAPSULATION_BOUNDARY.len()..];
+        // search for position of pre-encapsulation boundary
+        let position = Self::boyer_moore_horspool(data)?;
+
+        let data = if position > 0 {
+            // found pre-encapsulation boundary after some non-zero preamble
+            // preamble must not contain any NUL chars
+            // and the pre-encapsulation boundary must be at the start of a line
+
+            if data[..position].iter().any(|&b| b == grammar::CHAR_NUL) {
+                return Err(Error::InvalidText);
+            } else if data[position - 1] != grammar::CHAR_LF {
+                // strictly the line feed char is required
+                // replicating OpenSSL behavior
+                // https://github.com/RustCrypto/formats/pull/29#issuecomment-920548938
+                return Err(Error::PreEncapsulationBoundary);
+            } else {
+                // drop everything before the expected start of the label
+                &data[position + PRE_ENCAPSULATION_BOUNDARY.len()..]
+            }
+        } else {
+            &data[PRE_ENCAPSULATION_BOUNDARY.len()..]
+        };
 
         let (label, body) = grammar::split_label(data).ok_or(Error::Label)?;
 
--- a/pem-rfc7468/src/error.rs
+++ b/pem-rfc7468/src/error.rs
@@ -12,6 +12,9 @@ pub enum Error {
     /// Base64-related errors.
     Base64,
 
+    /// Contents of document is not valid text (contains NUL char)
+    InvalidText,
+
     /// Character encoding-related errors.
     CharacterEncoding,
 
@@ -38,6 +41,7 @@ impl fmt::Display for Error {
     fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
         f.write_str(match self {
             Error::Base64 => "PEM Base64 error",
+            Error::InvalidText => "PEM is not valid text as NUL char was found",
             Error::CharacterEncoding => "PEM character encoding error",
             Error::EncapsulatedText => "PEM error in encapsulated text",
             Error::HeaderDisallowed => "PEM headers disallowed by RFC7468",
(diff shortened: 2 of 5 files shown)


tarcieri:
I'm still torn on how to interpret that text from the RFC.

Perhaps one of the most confusing things about it is the notion that "data" is allowed before the encapsulation boundary is that it seems in conflict with the ABNF grammar given in [Section 3](https://datatracker.ietf.org/doc/html/rfc7468#section-3), which shows the following:

```
  textualmsg = preeb *WSP eol
               *eolWSP
               base64text
               posteb *WSP [eol]

  preeb      = "-----BEGIN " label "-----" 
  ...
```

In fact the "Standard" and "Strict" interpretations of the grammar (this library is aiming to comply $ openssl asn1parse -in pkcs1_with_preceeding_junk.pem
Error: offset out of range the "Strict" profile) both disallow leading whitespace.

I am curious what parsers, if any, accept the `pkcs1_with_preceeding_junk.pem` file in this PR.

OpenSSL, though probably not the best example for RFC 7468, rejects it:

```
$ openssl version
OpenSSL 1.1.1k  25 Mar 2021
$ openssl asn1parse -in pkcs1_with_preceeding_junk.pem
Error: offset out of range
```

Likewise for LibreSSL:

```
$ openssl version                                                                                                                             1 ↵
LibreSSL 2.8.3
$ openssl asn1parse -in pkcs1_with_preceeding_junk.pem
Error in encoding
4611673772:error:0DFFF07B:asn1 encoding routines:CRYPTO_internal:header too long:/System/Volumes/Data/SWE/macOS/BuildRoots/38cf1d983f/Library/Caches/com.apple.xbs/Sources/libressl/libressl-56.60.2/libressl-2.8/crypto/asn1/asn1_lib.c:152:
```

dwhjames:
Another example, 
```ssh-keygen -y -f …```
also accepts private keys that don’t start with
```-----BEGIN ```

**Comment by tarcieri** [comment]
I've been looking for clearer guidance on what is allowable in a PEM "preamble" and coming up empty.

Apparently they have a long history of being ill-defined and RFC7468 did nothing to address that.

RFC1521 says:

https://datatracker.ietf.org/doc/html/rfc1521#section-7.2.1

> These "preamble" and "epilogue" areas are generally not used because of the lack of proper typing of these parts and the lack of clear semantics

**Comment by dwhjames** [comment]
> is that it seems in conflict with the ABNF grammar given in Section 3, which shows the following:

I didn’t interpret that as a conflict. Beyond the basic details of labels and base64, the variations of the grammar seemed to be concerned with what goes on between those encapsulation boundaries in terms of extra whitespace or other non base64 chars. The general considerations seems to simply say that one should seek to the first occurrence of `-----BEGIN `, and it is one of the ‘MUST’s in the spec.

The main utility seems to be for pulling PEM out of an embedding in a larger document, or simply allowing for non-structured comments in the file ahead of the key itself.

**Comment by tarcieri** [comment]
I now have it on good authority that it is in fact “anything goes” with no restrictions on allowable characters.

When I have a bit more time I’ll review the PR in more detail.

**Comment by richsalz** [comment]
Beware!  Do not allow arbitrary bytes.  Arbitrary *text* is one thing, but if you see non-ASCII or non-UTF8 you should exit and report the file as invalid.

> The main utility seems to be for pulling PEM out of an embedding in a larger document, or simply allowing for non-structured comments in the file ahead of the key itself.

Yes.

Having said that, I'll point out that it is common to put an X.509 text output before the PEM of the cert.  Or to concat a key and cert file together, knowing that the code will ignore the one it is not looking for.

**Comment by sleevi** [comment]
> ```
> $ openssl version
> OpenSSL 1.1.1k  25 Mar 2021
> $ openssl asn1parse -in pkcs1_with_preceeding_junk.pem
> Error: offset out of range
> ```
> 
> Likewise for LibreSSL:
> 
> ```
> $ openssl version                                                                                                                             1 ↵
> LibreSSL 2.8.3
> $ openssl asn1parse -in pkcs1_with_preceeding_junk.pem
> Error in encoding
> 4611673772:error:0DFFF07B:asn1 encoding routines:CRYPTO_internal:header too long:/System/Volumes/Data/SWE/macOS/BuildRoots/38cf1d983f/Library/Caches/com.apple.xbs/Sources/libressl/libressl-56.60.2/libressl-2.8/crypto/asn1/asn1_lib.c:152:
> ```

I believe you forgot `—inform PEM` (Just to be explicit). Note that OpenSSL (and Libre) have several methods for reading PEM, and they are not all consistent. That is, some methods are implemented expecting the first character to be the preamble, largely because they’re targeting in-memory representions (e.g. they’re a low-level function called by a higher-level PEM parser), while others, such as the `rsa` or `x509` tools use the high level functions that skip such text.

As an example, both Chromium and ZLint take the position that all PEM certificate files checked in should contain a explanatory preamble. As @richsalz mentions, this most commonly takes the form of the textual output (ZLint does this, Chromium’s hand-generated or manually generated test certs do this), while it can also take something application specific (e.g. Chromium’s procedurally generated certs use this to replicate comments from the generator driver).

It becomes slightly easier to reason about when you also view other PEM blocks (e.g. a private key before a certificate) as being conceptually similar to that “free form text”. There again, OpenSSL/LibreSSL’s low level APIs reject this (because they expect exactly one block/label, and of a particular block/label type), while the high-level APIs are implemented simply as scanning for the next block from the current cursor, and seeing if it’s of the desired block type.

**Comment by tarcieri** [comment]
> I believe you forgot —inform PEM (Just to be explicit)

Aah interesting. Seems it autodetects PEM when there's no preamble?

**Comment by richsalz** [comment]
I don't think it autodetects, I think some flags have PEM as the default.  Saying `--help` to an OpenSSL command will tell you.

**Comment by tarcieri** [comment]
I guess it was this then? From @sleevi's comment:

> There again, OpenSSL/LibreSSL’s low level APIs reject this (because they expect exactly one block/label, and of a particular block/label type), while the high-level APIs are implemented simply as scanning for the next block from the current cursor, and seeing if it’s of the desired block type.

**Comment by richsalz** [comment]
The simple, and mostly correct I think, answer is that if you call `PEM_read_xxx` then you'll be okay; the PEM functions will skip forward until they find the right marker.  If you call things like `ReadPrivateKey` that does not happen.  You can imagine that the PEM functions call the lower-level functions once they've got to the right spot, but instead they both call common code IIRC.

**Comment by dwhjames** [comment]
> Beware! Do not allow arbitrary bytes. Arbitrary _text_ is one thing, but if you see non-ASCII or non-UTF8 you should exit and report the file as invalid.

@richsalz that feels like that goes down a rabbit hole.
- For just ASCII, of course one could disallow non-printable characters…
- For UTF-8, are all blocks permitted?
- Nothing in the RFC references UTF-8, and why should other unicode encodings be disallowed?
- Ultimately, to what end? Could you describe a threat for allowing something other than text, assuming one could come to an agreed definition for allowable text?

Note, the RFC says ‘data’ not ‘text’:
> Data before the encapsulation boundaries are permitted, and parsers MUST NOT malfunction when processing such data.

**Comment by tarcieri** [comment]
I would suggest allowing characters in the 7-bit ASCII range. That has a heritage in PEM by way of it originally being a restriction on SMTP (see RFC1421 Section 4.3.1)

**Comment by tarcieri** [comment]
After discussing this with various people more, I suppose I'm warming up to something similar to the implementation in this PR.

It seems there are legitimate reasons to support UTF-8 in the preamble. However, I don't think we should validate that the preamble contains well-formed UTF-8. It could potentially contain other non-UTF-8 encodings, and really the main thing this library is concerned with is skipping over it.

That particular set of tradeoffs means that there aren't any meaningful validations we can perform on the preamble, so we're back to @sleevi's original "anything goes" advice.

**Comment by tarcieri** [comment]
The one other alternative I suppose we could consider, which others have suggested, is ensuring the preamble is valid UTF-8.

There's nothing in RFC7468 about this, but then really there is very little in the RFC about anything relating to the preamble.

It would cover the legacy PEM 7-bit ASCII case, allow Unicode, but hopefully give a somewhat recognizable error if you accidentally confuse a DER file for a PEM file.

**Comment by dwhjames** [comment]
If you want to detect distinguish between an invalid PEM file (can’t find the pre-encapsulation boundary) and a possible DER file, wouldn’t a good proxy for that be to read the plausible tag byte and the plausible length byte(s) and check if those are valid? That would add usability without getting overly prescriptive (and computationally expensive?) about what is and isn’t allowed in the preamble.

**Comment by tarcieri** [comment]
I was giving DER as an example, but really thinking of a more general error case for "this file appears to be binary instead of text", which I think is the sort of thing @richsalz was hinting at

**Comment by dwhjames** [comment]
with OpenSSL 1.1.1l

```
cat pem-rfc7468/tests/examples/pkcs1.der pem-rfc7468/tests/examples/pkcs1.pem| openssl rsa -inform PEM
unable to load Private Key
4471246336:error:0909006C:PEM routines:get_name:no start line:crypto/pem/pem_lib.c:745:Expecting: ANY PRIVATE KEY
```

Fails at
https://github.com/openssl/openssl/blob/OpenSSL_1_1_1l/crypto/pem/pem_lib.c#L745

as a negative integer is returned from `BIO_gets`
https://github.com/openssl/openssl/blob/OpenSSL_1_1_1l/crypto/pem/pem_lib.c#L742

https://github.com/openssl/openssl/blob/e59bfbaa2dbd680f77e1121e382502bd522a466c/doc/man3/BIO_read.pod
> BIO_gets() performs the BIOs "gets" operation and places the data in buf. Usually this operation will attempt to read a line of data from the BIO of maximum length size-1. There are exceptions to this, however; for example, BIO_gets() on a digest BIO will calculate and return the digest and other BIOs may not support BIO_gets() at all. The returned string is always NUL-terminated and the '\n' is preserved if present in the input data. On binary input there may be NUL characters within the string; in this case the return value (if nonnegative) may give an incorrect length.

So, due to the implementation detail of null terminated strings in C, OpenSSL will accept anything other than NULL chars?

**Comment by tarcieri** [comment]
That’s interesting. Perhaps we can just disallow null bytes in the preamble.

**Comment by dwhjames** [comment]
In playing around with variations on

```
cat <(xxd -r -p <<<"070D0A") tests/examples/pkcs1.pem | openssl rsa -inform PEM
```

there seems to be at least one more rule to avoid accepting something that OpenSSL would _not_ accept.

The char before `-----BEGIN ` _must_ be a line feed `\n` (carriage return alone doesn‘t cut it).

**Comment by dwhjames** [comment]
After some more manual ‘fuzzing’, I think the ‘implementation derived rules’ are exactly that

1. no 0x0 bytes
2. if position of `-----BEGIN ` is greater than 0, then preceding byte must be `0xA` (but that may be platform specific; 0xD CR is not accepted for me)

Following those rules, you won’t accept anything that OpenSSL won’t.

**Comment by tarcieri** [comment]
Sounds good to me

**Comment by richsalz** [comment]
The RFC says "data"  That doesn't mean arbitrary binary bytes.  It **can't** mean that, because within the context of the RFC's, there is no way to indicate a switch from binary/text or even between coded charsets, like EBCDIC/ASCII. The bytes in the preamble **must** have the same encoding as the PEM text.

So the question becomes: do you want to take UTF8 because it's an ASCII superset, or do you want to take any ASCII compatible-format (which most 8bit charsets are), or do you want to take just ASCII.  That's your call. FWIW, my decision would be ASCII or UTF8 only. If you want to be nice, scan the file (but see next paragraph) and error out if it's invalid, or just warn users that the behavior is undefined if wrong.

As for this PR, a B-M search for `-----BEGIN ` seems like overkill to me.  In my experience, it's a small amount of text before, and easily skipped over.  Do `strstr` or its equivalent. Of course, that makes scanning for invalid bytes (malformed UTF8, non-ASCII if you go that way) impossible and perhaps renders your code open to a Unicode-related vulnerability.

**Comment by richsalz** [comment]
Do not guess at filetypes.  That way lies madness and security holes.

**Comment by sleevi** [comment]
Right, I think @richsalz and I disagree; I say that just to make sure it’s explicit so folks aren’t confused trying to reconcile the advice :)

I think the comments about security are a bit unfounded, at best, although I appreciate the concern about wanting to avoid differences in interpretation of a given file. Concretely, the “threat” here (and I admittedly use quotes because I think it’s misguided) is a situation where you have a file with `{ PEM 1, PEM 2, some binary, PEM 3}`, and different systems disagreeing on whether the contents of that is `{ 1, 2, 3 }` or `{ 1, 2 }`. As an API, that distinction is irrelevant - what matters is what you extracted. This is where silliness like OpenSSL’s newline predilection already differs from other extant parsers, and I don’t think anyone to date has claimed either system is insecure. For example, plenty of PEM parsers will ignore newlines by simply deleting them all before processing - allowing inputs such as
```
----- BEGIN
CERTIFICATE -----
```
to be parsed.

I appreciate the appeal to both Postel’s Law and the argument to reject Postel’s Law, but recognize this is a situation that really can go either way, and suggest there is no compelling reason for either extreme, other than convenience.

richsalz:
> I think the comments about security are a bit unfounded,

You suffer from OpenSSL, but you didn't spend all the years in the trenches with it like I did. :)

I am concerned not about newlines, but more about utf8 being misinterpreted and bad things happening.

tarcieri:
> I am concerned not about newlines, but more about utf8 being misinterpreted and bad things happening.

By a human or the parser?

The parser in this case is simply going to skip over it, and not attempt to interpret it as UTF-8, or expose it through its API.

tarcieri:
> As for this PR, a B-M search for -----BEGIN seems like overkill to me. In my experience, it's a small amount of text before, and easily skipped over. Do strstr or its equivalent. 

I'd agree it feels like a premature optimization/overkill for this particular use case.

I'm also unclear how/if the current implementation actually ensures that the preamble is free of NUL bytes. It seems like it will surface an error if it encounters one, but definitionally I don't see how else you can ensure their absence without scanning every one.

Given these rules:

> no 0x0 bytes
> if position of -----BEGIN is greater than 0, then preceding byte must be 0xA (but that may be platform specific; 0xD CR is not accepted for me)

It seems like there's a fairly simple approach to implementing them:

1. Check for `-----BEGIN`. If we've found it, we're done and can move on
2. If not:
  a. check for NUL. If we've encountered it, error out
  b. check for `0xA`. If we've found it, advance the cursor and go to 1
  c. otherwise advance the cursor and go to 2

dwhjames:
@tarcieri this code does that, no?

https://github.com/dwhjames/RustCrypto-formats/commit/076bad4e036fcf1ceecfa9743cc69d89d9c1efa9#diff-422e742ea5e412ac13822c4b62d34fad3304c57b0dbd0ff79e49da871bfe0233R182-R196

After finding `-----BEGIN `, check for NUL anywhere prior and that preceding char is `0xA`?

tarcieri:
Yes, it does, but you still have a B-M search, which is needless complexity if you're going to be scanning over the entire preamble a byte-at-a-time

tarcieri:
Merged with a simplified implementation in #40

dwhjames:
@tarcieri 👍

> Yes, it does, but you still have a B-M search, which is needless complexity if you're going to be scanning over the entire preamble a byte-at-a-time

But one quibble with the argumentation here.

The simplified implementation is O(len_of_boundary_prefix * len_of_doc). That makes it effectively an “eleven pass” algorithm.

With the substring search algorithm being O(len_of_doc) then a O(len_of_preamble) to scan for NUL. That’s a “two pass” algorithm.

I have no issue choosing the former over the latter; I just wanted to note that doing a second pass to scan for NUL does not invalidate the gains.

tarcieri:
Given the security-critical nature over this code, micro-optimizing how the preamble is consumed is a fairly low concern, especially given that most documents won't have one and there's a fast path provided for that case.

A much higher priority is code that's clear, concise, easy to understand, easy to modify/extend, and panic/bug free.

I also don't think there's really much of a practical performance difference between these two implementations, especially given that the boundary comparisons in the simplified version are only triggered by hitting LF, and that the comparison with the pre-encapsulation boundary will exit early. Your big O estimate ignores those details.

If there are legitimate concerns about performance, we can circle back on this, but that's a decision I'd like to make based on metrics, benchmarks, and real-world use cases.

dwhjames:
@tarcieri, I feel I am now guilty of “arguing the toss” here.

I had already assumed that what you just wrote out was your reasoning; thanks for writing it out all the same.

As I asterisked, other than some odd case like a preamble with lots of Markdown tables or a RST doc, or some pathological case, one would expect single pass behavior.

Yes, I am all on board with the preferences and reasons you just outlined. I probably should have stated up front that my motivation was to learn/explore Rust while hopefully adding some useful contributions.

richsalz:
@dwhjames, thanks!  I learned a couple of rust things.  And of course I totally agree with @tarcieri's https://github.com/RustCrypto/formats/pull/29#issuecomment-921396031 and I think it's good to get it written down and in the record.  Not for us here in this PR commenting, but for those who follow or perhaps stand on our shoulders :)

PR #30, opened 2021-09-15
SPKI SHA-256 fingerprint [WIP]

jklong:
Fixes RustCrypto/formats#10 
* Adds a `fingerprint` method to `SubjectPublicKeyInfo`.
* Adds a test for fingerprinting

In its current state this encodes `self` using the `der` crate in order to calculate the digest. 

An alternate way to implement this would be an associated function on `SubjectPublicKeyInfo` that simply takes a DER encoded struct as a `&[u8]` instead of performing any encoding operations itself.


jklong:
I just realised this was against your master branch, not mine, and I've been doing the checks against your CI all this time. Apologies, I'll get it together and reopen

tarcieri:
This looks mostly fine. Feel free to reopen when you're ready.

Issue #1, opened 2021-09-14
Missing formats


tarcieri:
- [ ] PKCS#1: RSA Cryptography Standard
- [ ] PKCS#5: Password-based Encryption Standard
- [ ] PKCS#7: Cryptographic Message Syntax Standard
- [ ] PKCS#8: Private-Key Information Syntax Standard
- [ ] PKCS#10: Certification Request Standard
- [ ] PKCS#12: Personal Information Exchange Syntax Standard
- [ ] SPKI: X.509 Subject Public Key Info
- [ ] X.509: Public Key Infrastructure Certificate and Certificate Revocation List (CRL) Profile

Please post a comment to request additional formats

jklong:
#3 will also require a PKCS#7 implementation so I'll start work on that as well. I've created #28 to track.

franziskuskiefer:
Do you want to stick to the big ones or are you also interested in things like the ["TLS codec"](https://datatracker.ietf.org/doc/html/rfc8446#section-3). It's not really a standard but it's being used quite a bit by now. I have a version here https://github.com/openmls/tls-codec/

tarcieri:
@franziskuskiefer if you so happen to be looking for another home for it I'd say that would be a great thing to have here

jonasbb:
The new `ssh-key` crate does not seem to support ssh-keys backed by security keys, which contain the `sk` suffix in the name. There is `ecdsa_sk` which uses `sk-ecdsa-sha2-nistp256@openssh.com` in the `.pub` file and `ed25519_sk` using `sk-ssh-ed25519@openssh.com` as the identifier. 
It would be great to have support for these newer key types too, but it is not clear to me if that is currently on the todo list.

tarcieri:
@jonasbb yes, it would be good to add support for those. Likewise it would be nice to have support for OpenSSH certificates.

I will try to get these items added to the TODO list in the README, or otherwise would merge a PR to add them.

enri1196:
Do you think RFC3161 TimeStamp Protocol would be a good addition to the formats?

reference: https://www.rfc-editor.org/rfc/rfc3161

carl-wallace:
I think RFC3161 would be a good addition and have thought about implementing it before but had not gotten around to it. It's referenced in some new work, for example, draft-birkholz-rats-epoch-markers.

enri1196:
I'd like to help, can I just open a PR once I've got some basic implementation of the structures? Maybe a new folder under formats/tsp?

[...]

PR #31, opened 2021-09-15
SPKI SHA-256 fingerprint (#1)

jklong:
* Add fingerprint feature 

* Add rustdocs feature flags

This currently encodes `self` using the `der` crate in order to create the fingerprint.

### DIFF
--- a/pkcs5/src/pbes2/encryption.rs
+++ b/pkcs5/src/pbes2/encryption.rs
@@ -51,18 +51,14 @@ pub fn encrypt_in_place<'b>(
                 .map_err(|_| CryptoError)?;
             cipher.encrypt(buffer, pos).map_err(|_| CryptoError)
         }
-        #[cfg(feature = "des-insecure")]
-        EncryptionScheme::DesCbc { iv } => {
-            let cipher =
-                DesCbc::new_from_slices(encryption_key.as_slice(), iv).map_err(|_| CryptoError)?;
-            cipher.encrypt(buffer, pos).map_err(|_| CryptoError)
-        }
         #[cfg(feature = "3des")]
         EncryptionScheme::DesEde3Cbc { iv } => {
             let cipher = DesEde3Cbc::new_from_slices(encryption_key.as_slice(), iv)
                 .map_err(|_| CryptoError)?;
             cipher.encrypt(buffer, pos).map_err(|_| CryptoError)
         }
+        #[cfg(feature = "des-insecure")]
+        EncryptionScheme::DesCbc { .. } => Err(CryptoError),
     }
 }
 


spki/src/spki.rs
```diff
@@ -57,3 +68,34 @@ impl<'a> Message<'a> for SubjectPublicKeyInfo<'a> {
...
+}
+
+#[cfg(feature = "fingerprint")]
+#[cfg_attr(docsrs, doc(cfg(feature = "fingerprint")))]
+impl<'a> SubjectPublicKeyInfo<'a> {
+    /// Calculate the SHA-256 fingerprint of this SubjectPublicKeyInfo
+    pub fn fingerprint(
+        &self,
+        fingerprint: &'a mut [u8],
+    ) -> core::result::Result<&'a str, FingerprintError> {
```
tarcieri, line 87: For a non-allocating API, I would suggest returning `digest::Output<Sha256>` which is effectively a byte array containing the raw digest.

Adding Base64 encoding support seems OK, but for that I'd probably just suggest adding an additional method like `fingerprint_base64` which returns a `String` and gating it on the `alloc` feature.

jklong: Returning `digest::Output<Sha256>` was my initial implementation, however [RFC7469 §2.4](https://datatracker.ietf.org/doc/html/rfc7469#section-2.4) states:
> An SPKI Fingerprint is defined as the output of a known cryptographic
   hash algorithm whose input is the DER-encoded ASN.1 representation of
   the Subject Public Key Info (SPKI) of an X.509 certificate.  A Pin is
   defined as the combination of the known algorithm identifier and the
   SPKI Fingerprint computed using that algorithm.

>   The SPKI Fingerprint is encoded in base 64 for use in an HTTP header
   [RFC4648].

which is of course distinct from the X.509 AKI and SKI fields ([RFC5280 §4.2.1](https://datatracker.ietf.org/doc/html/rfc5280#section-4.2.1)), which use the SHA-1 digest bytes directly, or the generally recognised concept of a "certificate thumbprint" which does not appear in RFC5280 at all but I've seen implemented widely as a SHA-1 hash and less widely as SHA-256.

So my reasoning here is to stick to the definition of a SPKI fingerprint and return Base64 only. I'm happy to additionally gate this on `alloc`, return a `String` and remove it from the non-allocating API entirely.

jklong: I should also note that this only returns the fingerprint, not a full Pin including an `AlgorithmIdentifier`. 

Adding that may be a future enhancement

tarcieri: All I'm seeing there re: Base64 is:

> The SPKI Fingerprint is encoded in base 64 for use in an HTTP header [RFC4648].

So I think it's fine to have a couple methods like:

- `fingerprint_bytes` (or maybe just call it `fingerprint`)
- `fingerprint_base64` (gated on `alloc`)

...and the latter is what you'd use as an HTTP header

jklong: My read of that is that a SPKI fingerprint is encoded in Base64 as it is intended to be used in a HTTP header. 

Doesn't matter. It's not a lot of work to split out the functions. Watch this space.

PR #36, opened 2021-09-16
Add SPKI fingerprint

jklong:
Finishing the discussion started in RustCrypto/formats#31 

* Add SPKI SHA-256 fingerprint returned as bytes gated behind the `fingerprint` function.
* Move the Base64 fingerprint to be additionally gated behind `alloc`
* Remove the no longer used error wrapper


.github/workflows/spki.yml
```diff
@@ -55,4 +55,7 @@ jobs:
           toolchain: ${{ matrix.rust }}
           override: true
       - run: cargo test --release
+      - run: cargo test --release --features fingerprint
+      # - run: cargo test --release --features alloc
```
tarcieri, line 59: Why is this commented out?

spki/src/spki.rs
```diff
@@ -1,11 +1,22 @@
 //! X.509 `SubjectPublicKeyInfo`
 
+#[cfg(feature = "alloc")]
+extern crate base64ct;
+#[cfg(feature = "fingerprint")]
+extern crate sha2;
```
tarcieri, line 6: You don't need these `extern crate` directives in the 2018 edition. The only one that's needed is linking to liballoc, and that's an artifact of `#![no_std]`

spki/src/spki.rs
```diff
@@ -57,3 +68,24 @@ impl<'a> Message<'a> for SubjectPublicKeyInfo<'a> {
...
     }
 }
+
+#[cfg(feature = "fingerprint")]
+#[cfg_attr(docsrs, doc(cfg(feature = "fingerprint")))]
+impl<'a> SubjectPublicKeyInfo<'a> {
+    const BUFSIZE: usize = 4096;
+
+    #[cfg(all(feature = "fingerprint", feature = "alloc"))]
+    #[cfg_attr(docsrs, doc(cfg(all(feature = "fingerprint", feature = "alloc"))))]
```
tarcieri, line 78: IIRC `doc(cfg(...))` doesn't like this sort of nesting and you need to include two directives, but I'd have to double check how it renders locally. You can do that with something like:

```
$ cargo +nightly rustdoc --all-features --open -- --cfg docsrs
```

.github/workflows/spki.yml
jklong: Because I forgot to remove it. Will fix. The `alloc` feature only covers the Base64 encoding at this point, there are no additional functionality or dependencies to be tested in `alloc` alone.

spki/src/spki.rs
jklong: I did check that before committing, it renders ok. If two directives is the preferred style I'll happily change

jklong:
Thanks @tarcieri for the review. Will push fixes tonight when I get a moment

spki/src/spki.rs
jklong: Will fix. Cheers

spki/src/spki.rs
tarcieri: Oh hey, it does work! Great. Never mind then.

Screen Shot 2021-09-16 at 5 03 12 PM

.github/workflows/spki.yml
jklong: Fixed in latest push

spki/src/spki.rs
jklong: Fixed in latest push

tarcieri:
Looks good now, thanks!

tarcieri:
Thank you!

PR #46, opened 2021-09-19
Support more characters in labels

janimo:
This helps for example with some keys used by Tor relays that have ED25519 CERT for a label.
The patch splits is_labelchar into two functions, one matching all non-special chars used for labelchar as described in the RFC,
the other also including blanks (and later hyphen-minus).


pem-rfc7468/src/grammar.rs
```diff
@@ -133,7 +139,7 @@ pub(crate) fn split_label(bytes: &[u8]) -> Option<(&str, &[u8])> {
     for &char in bytes {
         // Validate character
         // TODO(tarcieri): unify with `is_labelchar`/`validate_label`
-        if matches!(char, b'A'..=b'Z') {
```
tarcieri, line 136: You can remove the above TODO here now

tarcieri (approved):
Looks good. One small cleanup suggestion.

janimo:
Thanks for the quick review! I was somewhat reluctant to touch that TODO initially since it also referenced validate_label, but I have now forced pushed an updated second commit with that TODO removed.

tarcieri:
Thank you!

PR #54, opened 2021-09-23
feat: add support for constructed bit

ajpauwels:
- Previously, DER outputted by the Application, ContextSpecific,
  and Private types were assumed to have their constructed bit
  (0b100000) set
- This PR adds a parameter when decoding/encoding Application,
  ContextSpecific, and Private types that allows these types to
  not have their constructed bit (0b100000) set


ajpauwels:
Closing in favor of #55

PR #63, opened 2021-09-29
der: introduce `DecodeValue`/`EncodeValue` traits

tarcieri:
Adds traits for decoding and encoding the value (i.e. body) of a TLV-encoded message field.

The previous blanket impl of `Decodable` for `TryFrom<Any>` is removed and replaced with a blanket impl for `DecodeValue` + `Tagged`. This avoids having to round trip through the `Any` type (and thus have a contiguous encoded message stored in a slice).

Likewise the blanket impl of `Encodable` for `Message` is removed, and changed to a blanket impl for `EncodeValue` + `Tagged`. However, `Message` still receives a vicarious impl of `Encodable` because a new blanket impl of `EncodeValue` for `Message` has been added.

Overall this change should make decoding and encoding much more flexible, and help sort out encoding of `EXPLICIT` vs `IMPLICIT` tagging modes. However, that work in particular is left for a followup PR.

### DIFF
--- a/der/src/asn1/set_of.rs
+++ b/der/src/asn1/set_of.rs
@@ -1,8 +1,8 @@
 //! ASN.1 `SET OF` support.
 
 use crate::{
-    asn1::Any, ByteSlice, Decodable, Decoder, Encodable, Encoder, Error, ErrorKind, Length, Result,
-    Tag, Tagged,
+    asn1::Any, ByteSlice, Decodable, DecodeValue, Decoder, Encodable, Encoder, Error, ErrorKind,
+    Length, Result, Tag, Tagged,
 };
 use core::{convert::TryFrom, marker::PhantomData};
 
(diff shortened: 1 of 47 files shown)


der/src/asn1/set_of.rs
```diff
@@ -18,10 +18,11 @@ use {
 /// When encoded as DER, `SET OF` is lexicographically ordered. To implement
 /// that requirement, types `T` which are elements of [`SetOf`] MUST provide
 /// an impl of `Ord` which ensures that the corresponding DER encodings of
-/// a given type are ordered.
+/// a given type are ordffvkjnkjnlvceubbngtirurinkbvfditfnihengdigvl
+/// ered.
 pub trait SetOf<'a, 'b, T>: Decodable<'a> + Encodable
 where
-    T: Clone + Decodable<'a> + Encodable + Ord,
+    T: Clone + Decodable<'a> + Encodable + Ord + Tagged,
```
dwhjames, line 25: @tarcieri I previously had some code

```rs
…
  for any in SetOfRef::<Any>::try_from(any)?.elements() {
    …
```

but that no longer compiles.

```
error[E0599]: the method `elements` exists for struct `SetOfRef<'_, der::Any<'_>>`, but its trait bounds were not satisfied
  --> der/tests/asn1parse.rs:40:56
   |
40 |               for any in SetOfRef::<Any>::try_from(any)?.elements() {
   |                                                          ^^^^^^^^ method cannot be called on `SetOfRef<'_, der::Any<'_>>` due to unsatisfied trait bounds
   |
  ::: der/src/asn1/any.rs:22:1
   |
22 |   pub struct Any<'a> {
   |   ------------------ doesn't satisfy `der::Any<'_>: Tagged`
   |
  ::: der/src/asn1/set_of.rs:41:1
   |
41 | / pub struct SetOfRef<'a, T>
42 | | where
43 | |     T: Clone + Decodable<'a> + Encodable + Ord,
44 | | {
...  |
49 | |     element_type: PhantomData<T>,
50 | | }
   | |_- doesn't satisfy `SetOfRef<'_, der::Any<'_>>: SetOf<der::Any<'_>>`
   |
   = note: the following trait bounds were not satisfied:
           `der::Any<'_>: Tagged`
           which is required by `SetOfRef<'_, der::Any<'_>>: SetOf<der::Any<'_>>`
```

PR #81, opened 2021-10-09
add encoder/decoder for Validity structure

carl-wallace:
Added encoder/decoder for Validity structure along with some corresponding tests. Marked .idea folder as ignored in .gitignore.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,2 +1,3 @@
 target/
 **/*.rs.bk
+.idea
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,2 @@
 target/
 **/*.rs.bk
-.idea
--- a/x509/tests/validity.rs
+++ b/x509/tests/validity.rs
@@ -92,11 +92,11 @@ fn encode_validity() {
         &hex!("301E170D3130303130313038333030305A170D3330313233313038333030305A")[..]
     );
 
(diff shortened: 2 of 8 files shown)


.gitignore
```diff
@@ -1,2 +1,3 @@
 target/
 **/*.rs.bk
+.idea
```
tarcieri, line 3: Speaking as a fellow IntelliJ user, I would suggest you configure something like this in your local `~/.gitconfig`:

```
[core]
  excludesfile = /home/myuser/.gitignore
```

Using local excludes means that upstream projects don't have to track editor-specific ignore configuration for every editor under the sun.

tarcieri:
Needs rustfmt, and a couple other minor nits, but otherwise looks good

x509/tests/validity.rs
```diff
@@ -0,0 +1,124 @@
...
+    assert_eq!(
+        b3,
+        &hex!("3020170D3130303130313038333030305A180F32303530303130313132303130305A")[..]
+    );
+
+    // // Decode Validity from ValidGeneralizedTimenotBeforeDateTest4EE.crt in NIST's PKITS certificate collection
+    // //  97  32:     SEQUENCE {
+    // //  99  15:       GeneralizedTime 01/01/2002 12:01:00 GMT
+    // // 116  13:       UTCTime 31/12/2030 08:30:00 GMT
+    // //        :       }
```
tarcieri, line 114: These appear to be double-commented?

PR #82, opened 2021-10-11
`pkcs7` crate

dwhjames:
Before I develop this further, I could do with some feedback on if I’m going in the right direction.


jklong:
I've started on a PKCS7 implementation as well, to support PKCS12.  I've not had a lot of time recently to work on it however and with the recent changes to the `der` crate API what I did have is currently broken anyhow.

I'm happy to abandon that work in favour of this, I may even work with you on it?

jklong:
> Frankly, your design is more streamlined and I'm happy to abandon that work in favour of this, I may even work with you on it from your fork?

Or, more ideally, if we can get a feature branch here? Unless we're not doing long-lived branches?

PR #107, opened 2021-10-15
Incremental progress on Certificate, Extensions and Name

carl-wallace:
Added several structure definitions to sustain relatively shallow parse of Certificate structure.


x509/src/certificate.rs
```diff
@@ -0,0 +1,145 @@
...
+                return Err(der::Tag::Integer.value_error());
+            }
+
+            // TODO refactor unwrap if this implementation remains
+            let serial_number = UIntBytes::decode(decoder).unwrap();
+            let signature = AlgorithmIdentifier::decode(decoder).unwrap();
+            let issuer = RDNSequence::decode(decoder).unwrap();
+            let validity = Validity::decode(decoder).unwrap();
+            let subject = RDNSequence::decode(decoder).unwrap();
+            let subject_public_key_info = SubjectPublicKeyInfo::decode(decoder).unwrap();
```
tarcieri, line 79: You should be able to replace pretty much all of these unwraps with `?`...

carl-wallace: Will need to refactor the version field unwrap too.

x509/src/certificate.rs
```diff
@@ -63,20 +63,23 @@ pub struct TBSCertificate<'a> {
...
-                return Err(der::Tag::Integer.value_error());
+            let mut version = 0;
+            if (0b10100000 | VERSION_TAG.value()) == decoder.peek().unwrap() {
+                version = decoder
+                    .context_specific::<u8>(VERSION_TAG, TagMode::Explicit)?
+                    .unwrap();
+                if version != VERSION {
+                    return Err(der::Tag::Integer.value_error());
+                }
             }
```
tarcieri: If you find the version handling logic getting particularly complicated, you might consider extracting it out into a type ala:

https://github.com/RustCrypto/formats/blob/master/pkcs8/src/version.rs

PR #125, opened 2021-10-18
tls_codec: implement derivation for a simple enum pattern common in the MLS spec

kkohbrok:
This PR allows derivation of the `tls_codec` serialization and deserialization traits for a simple enum pattern that is commonly used in the MLS spec.

In particular, it allows derivation for enums for which a `...Type` exists. Consider the following example enum:
```
enum SimpleEnumType {
    One = 1u8,
    Two = 2u8,
}

enum SimpleEnum {
    One(CredentialOne),
    Two(CredentialTwo),
}
```
We can now derive `TlsSerialize, TlsDeserialize` and `TlsSize` for `SimpleEnum`. Here, `tls_serialize` of an instance of `SimpleEnum::One(cred)` results in the serialization of `SimpleEnumType::One`, concatenated with the serialization of `cred` (which of course needs to implement `TlsSerialize` itself. Or put differently, it results in the serialization of.
```
struct MlsCredential {
    type: SimpleEnumType,
    cred: CredentialOne/Two,
}
```

### DIFF
--- a/tls_codec/derive/src/lib.rs
+++ b/tls_codec/derive/src/lib.rs
@@ -445,8 +445,6 @@ fn impl_serialize(parsed_ast: TlsStruct) -> TokenStream2 {
                         enum_value.tls_serialize(writer)
                     }
                 }
-                // This can probably be optimised such that we only have one
-                // `match` statement.
                 EnumStyle::TupleStruct => {
                     let type_mapping: Vec<TokenStream2> = variants
                         .iter()
--- a/tls_codec/derive/src/lib.rs
+++ b/tls_codec/derive/src/lib.rs
@@ -46,7 +46,7 @@ struct Enum {
     variants: Punctuated<Variant, Comma>,
 }
 
-const ENUM_TYPE_POSTFIX: &'static str = "Type";
+const ENUM_TYPE_POSTFIX: &str = "Type";
 
 #[derive(Clone)]
 enum TlsStruct {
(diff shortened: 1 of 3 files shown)


kkohbrok:
Thanks for the review! I'm happy to write documentation, but I don't quite understand what you mean by "make sure it's well defined in the spec". Do you mean that we should include references to the places in the spec where this pattern occurs?

franziskuskiefer:
> Do you mean that we should include references to the places in the spec where this pattern occurs?

I'd like to have only things implemented in the crate that are actually specified.
The `optional` extension in MLS is defined in §7.6
```
struct {
    uint8 present;
    select (present) {
        case 0: struct{};
        case 1: T value;
    }
} optional<T>;
```

We want to have a similar extension for typed structs.
```
enum {
  ...
} T;
struct {
    T type;
    select (type) {
        ...
    }
} TypedStruct<T>;
```
Right now this is implicit in the MLS spec. I think we should make it explicit in order have an extension here to implement rather than an implementation specific mechanism.

kkohbrok:
Ah, I see. You want to change the spec first to explicitly define this pattern. I'll file an issue.

franziskuskiefer:
@kkohbrok do you still want to get this in or shall we close it for now?

kkohbrok:
Let's close it for now. I don't think this will land in the MLS spec, but if it ends up in the implementation considerations doc, we can still revive it if we want.

PR #136, opened 2021-10-21
Rust 2021 edition upgrade

tarcieri:
Most of the crates in this repository were already MSRV 1.55 to take advantage of non-`Copy`/`Default` array constructors using the new const-generic `map` constructor.

However, now that the 2021 edition is out, it would be a shame to bump to 1.55 instead of 1.56 when we're bumping MSRV anyway.

This commit bumps `edition = "2021"`, also adding the new `rust-version = "1.56"` Cargo.toml attribute to properly signal MSRV.

All crates have been checked using `cargo fix --edition`, however no attempts have yet been made at upgrading edition idioms.


tls_codec/derive/src/lib.rs
```diff
@@ -231,15 +231,15 @@ fn impl_tls_size(parsed_ast: TlsStruct) -> TokenStream2 {
             field_paths,
         }) => {
             quote! {
-                impl#generics tls_codec::Size for #ident#generics {
+                impl #generics tls_codec::Size for #ident #generics {
```
tarcieri, line 234: @franziskuskiefer FYI, these were auto-corrected by `cargo fix --edition`:

```
error: prefix `impl` is unknown
   --> tls_codec/derive/src/lib.rs:234:17
    |
234 |                 impl#generics tls_codec::Size for #ident#generics {
    |                 ^^^^ unknown prefix
    |
    = note: prefixed identifiers and literals are reserved since Rust 2021
help: consider inserting whitespace here
    |
234 -                 impl#generics tls_codec::Size for #ident#generics {
234 +                 impl #generics tls_codec::Size for #ident#generics {
```

See: https://doc.rust-lang.org/edition-guide/rust-2021/reserving-syntax.html

franziskuskiefer: 👍🏻 well, if that's the only fix needed for 2021 I'm fine with that 😄

PR #177, opened 2021-11-03
pem-rfc7468: add `Decoder` struct

### LINKED ISSUE #176: pem-rfc7468: Handle non-standard line sizes when decoding
RFC 7468 states that parsers [may handle other line sizes](https://datatracker.ietf.org/doc/html/rfc7468#:~:text=Parsers%20MAY%0A%20%20%20handle%20other%20line%20sizes.).

OpenSSH encoded private keys use a line-length of 70, but are otherwise compliant with RFC 7468. Supporting non-standard line sizes would allow `pem-rfc7468` to read these keys.

Unfortunately, handling non-standard sizes is likely to cause a performance regression for a couple of reasons:
* Need to scan for newlines instead of splitting at exact points
* Need to handle partial data because `70` (or other values) may not be an exact base64 chunk

For those reasons, I propose additional decoder methods that are less strict.

* `decode_less_strict`
* `decode_vec_less_strict`

tarcieri:
This adds a `Decoder` struct which supports one configurable setting: the Base64 line-wrapping width.

From RFC7468 Section 2:

> Parsers MAY handle other line sizes.

Closes #176

### DIFF
--- a/pem-rfc7468/src/decoder.rs
+++ b/pem-rfc7468/src/decoder.rs
@@ -13,10 +13,7 @@
 #[cfg(feature = "alloc")]
 use alloc::vec::Vec;
 
-use crate::{
-    grammar, Error, Result, BASE64_WRAP_WIDTH, POST_ENCAPSULATION_BOUNDARY,
-    PRE_ENCAPSULATION_BOUNDARY,
-};
+use crate::{grammar, Error, Result, POST_ENCAPSULATION_BOUNDARY, PRE_ENCAPSULATION_BOUNDARY};
 use base64ct::{Base64, Encoding};
 use core::str;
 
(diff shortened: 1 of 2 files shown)


tarcieri:
@kjvalencik this is what I propose for addressing #176. It would allow you to do:

```rust
pem_rfc7468::Decoder { wrap_width: 70 }.decode_vec(input)
```

...or for a fully futureproofed version:

```rust
pem_rfc7468::Decoder { wrap_width: 70, ..Default::default() }.decode_vec(input)
```

kjvalencik:
@tarcieri I really like this API with the `Decoder`. Unfortunately, this only works when line-length is at a 4:3 ratio, otherwise base64 decode fails.

For example, this works with `64` and `72`. But, if I try to decode with `70` (what OpenSSH uses), it fails to decode base64 (since it's incomplete input).

```rust
    #[test]
    fn openssh_example() {
        let pem = include_bytes!("../tests/examples/openssh.pem");
        let result = Encapsulation::parse(pem).unwrap();
        assert_eq!(result.label, "OPENSSH PRIVATE KEY");

        let decoder = Decoder { wrap_width: 70 };
        let mut buf = [0; 2048];
        result.decode(&decoder, &mut buf).unwrap();
    }
```

```
---- decoder::tests::openssh_example stdout ----
thread 'decoder::tests::openssh_example' panicked at 'called `Result::unwrap()` on an `Err` value: Base64', pem-rfc7468/src/decoder.rs:308:43
```

tarcieri:
Aah wow, that's really annoying. Supporting anything else would require a buffered Base64 decoder, which isn't exactly trivial to implement, and something I'd like to avoid in proper PKCS/PKIX use cases.

I suppose I could add an error case for if `wrap_width % 4 != 0` for now, and if someone would like to contribute a buffered decoder, it could eventually just work.

PR #202, opened 2021-11-11
der: support for handling DEFAULT values in SEQUENCEs

carl-wallace:


der/derive/src/sequence.rs
```diff
@@ -127,35 +130,41 @@ impl SequenceField {
...
     /// Derive code for decoding a field of a sequence.
     fn to_decode_tokens(&self) -> TokenStream {
         let ident = &self.ident;
-        if self.attrs.asn1_type.is_some() {
-            let field_decoder = self.attrs.decoder();
-            quote! { let #ident = #field_decoder.try_into()?; }
+        let ty = self.field_type.clone();
+        if None != self.attrs.default {
+            let fname = syn::Ident::new(&self.attrs.default.clone().unwrap(), self.ident.span());
+            quote!(let mut #ident = Some(decoder.decode::<#ty>()?.unwrap_or_else(|| #fname()));)
```
tarcieri: Slightly more idiomatic way to write this would be something like:

```suggestion
        if let Some(default) = &self.attrs.default {
            let fname = syn::Ident::new(default, self.ident.span());
            quote!(let mut #ident = Some(decoder.decode::<#ty>()?.unwrap_or_else(|| #fname()));)
```

der/derive/src/attributes.rs
```diff
@@ -68,15 +68,22 @@ pub(crate) struct FieldAttrs {
     /// Inherits from the type-level tagging mode if specified, or otherwise
     /// defaults to `EXPLICIT`.
     pub tag_mode: TagMode,
+
+    /// Indicates name of function that supplies the default value, which will be used in cases
+    /// where encoding is omitted per DER and to omit the encoding per DER
+    pub default: Option<String>,
```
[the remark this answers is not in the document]
carl-wallace: The String value was inherited from AttrNameValue. Refactored to be an Ident using the span value from the syn::Attribute before the conversion. In serde, it's a Path.

der/derive/src/sequence.rs
carl-wallace: Refactored per the suggestion in this spot and one other similar spot.

der/derive/src/attributes.rs
```diff
@@ -101,6 +109,12 @@ impl FieldAttrs {
                 }
 
                 tag_mode = Some(mode);
+            } else if let Some(default_attr_str) = attr.parse_value::<String>("default") {
+                if default.is_some() {
+                    abort!(attr.value, "duplicate ASN.1 `tag_mode` attribute");
+                }
+
+                default = Some(Ident::new(&default_attr_str, attrs[counter].span()));
```
tarcieri, line 117: This works for the `Ident` case, although I think to do this "properly" it should be a `Path`.

I think you can use the [`Parse`](https://docs.rs/syn/latest/syn/parse/trait.Parse.html) trait to parse a `Path` from the `LitStr` (retaining span info automatically), and perhaps the `value` portion of `AttrNameValue` should retain the original `LitStr`.

der/derive/src/attributes.rs
```diff
@@ -101,6 +108,12 @@ impl FieldAttrs {
                 }
 
                 tag_mode = Some(mode);
+            } else if let Some(_default_attr_str) = attr.parse_value::<String>("default") {
+                if default.is_some() {
+                    abort!(attr.value, "duplicate ASN.1 `tag_mode` attribute");
+                }
```
tarcieri, line 114:
```suggestion
            } else if attr.parse_value::<String>("default").is_some() {
                if default.is_some() {
                    abort!(attr.value, "duplicate ASN.1 `default` attribute");
                }
```

der/derive/src/attributes.rs
```diff
@@ -68,15 +68,22 @@ pub(crate) struct FieldAttrs {
...
 
 impl FieldAttrs {
     /// Parse attributes from a struct field or enum variant.
     pub fn parse(attrs: &[Attribute], type_attrs: &TypeAttrs) -> Self {
         let mut asn1_type = None;
         let mut context_specific = None;
+
         let mut tag_mode = None;
 
+        let mut default = None;
```
tarcieri, line 85: I think the added whitespace here is unnecessary... these should be grouped together as the fields-under-construction

der/tests/derive.rs
```diff
@@ -218,8 +218,7 @@ mod sequence {
     pub struct SubjectPublicKeyInfo<'a> {
         pub algorithm: AlgorithmIdentifier<'a>,
 
-        #[asn1(type = "BIT STRING")]
-        pub subject_public_key: &'a [u8],
+        pub subject_public_key: BitString<'a>,
```
tarcieri, line 221: Why did you change this?

carl-wallace: Because it made the default stuff messy for not particular value to my eye. We could add it back with cost of some additional verbosity in the default bits. If we do restore, I'd prefer to wait until context specific proc macro support lands (or address in tandem with that).

tarcieri: I'd really prefer to retain the functionality, although I'd be happy with fixing it in a follow-up PR after merging this one

der/derive/src/sequence.rs
```diff
@@ -127,35 +130,39 @@ impl SequenceField {
...
 
     /// Derive code for decoding a field of a sequence.
     fn to_decode_tokens(&self) -> TokenStream {
         let ident = &self.ident;
-        if self.attrs.asn1_type.is_some() {
-            let field_decoder = self.attrs.decoder();
-            quote! { let #ident = #field_decoder.try_into()?; }
+        let ty = self.field_type.clone();
+        if let Some(default) = &self.attrs.default {
+            quote!(let mut #ident = Some(decoder.decode::<#ty>()?.unwrap_or_else(#default));)
```
tarcieri, line 151: Thought I'd merge this and try to clean up my remaining nits, but this seems a bit problematic. I'm not sure this is what you actually intended?

It seems like when `default` is present, it should attempt to decode the given type as `Option<#ty>`, and then call `unwrap_or_default` on the result, if the `Option` decodes successfully.

carl-wallace: OK. I think the Option<> on the default'ed fields will likely go away as Context Specific is worked. The goal here was to *always* return a value where there was a default, perhaps stretching Option a bit.

carl-wallace: I see now. The Option<> became unnecessary as default support was added. Originally, the absence of the field signified default and the caller had to supply the default value. Removing Option was natural next step.

PR #204, opened 2021-11-11
der_derive: fix `#[asn1(type = "...")]` attribute

tarcieri:
PR #202 removed support for the `#[asn1(type = "...")]` attribute when deriving `Sequence`.

This PR restores it, and also slightly changes the way the `default` attribute is handled, with added tests.


tarcieri:
@carl-wallace I attempted to address this as part of this PR: https://github.com/RustCrypto/formats/pull/202/files#r747837193

PR #231, opened 2021-11-16
spki: replace `from_spki` with `TryFrom`

tarcieri:


mzohreva:
A change to public API like this one requires bumping the minor version instead of releasing as a patch.

tarcieri:
This was a breaking change to the v0.5.0 release, however that release was yanked.

Did this actually cause a real-world breakage in a program?

mzohreva:
No breakage but a lot of confusion since I was seeing a different definition of the trait looking at docs.rs compared to what the compiler was claiming 🙂

tarcieri:
If you were looking at the docs.rs for a yanked release, it should have a fairly prominent:

> ⚠️ This release has been yanked, go to latest version

...at the top of the page

mzohreva:
True, but I was looking at this page: https://docs.rs/pkcs8/0.8.0/pkcs8/trait.DecodePublicKey.html (which is a re-export of the `spki` crate's public API)

tarcieri:
Aah, I see...

PR #238, opened 2021-11-17
spki v0.5.1

tarcieri:
### Changed
- Replace `from_spki` with `TryFrom` ([#231])

[#231]: https://github.com/RustCrypto/formats/pull/231


tarcieri:
Note: I had intended to do this prior to releasing `spki` v0.5.0, as it's a breaking change.

The original release will be yanked.

PR #246, opened 2021-11-18
Add support for context specific fields with default values

carl-wallace:


der/derive/src/attributes.rs
```diff
@@ -184,7 +185,7 @@ impl FieldAttrs {
 
             let context_specific = match self.tag_mode {
                 TagMode::Explicit => {
-                    if self.extensible || self.optional {
+                    if self.extensible || self.optional || self.default.is_some() {
```
tarcieri, line 188: Since `self.optional || self.default.is_some()` is getting checked in several places, it might make sense to make a binding or extract a method, e.g. `is_optional()`

der/tests/derive.rs
```diff
@@ -261,6 +339,60 @@ mod sequence {
...
+        //  0  15: SEQUENCE {
+        //   2   3:   OBJECT IDENTIFIER basicConstraints (2 5 29 19)
+        //   7   1:   BOOLEAN TRUE
+        //  10   5:   OCTET STRING, encapsulates {
+        //  12   3:     SEQUENCE {
+        //  14   1:       BOOLEAN TRUE
+        //        :       }
+        //        :     }
+        //        :   }
+        let ext1 = ExtensionExample::from_der(&hex!("300F0603551D130101FF040530030101FF")).unwrap();
```
tarcieri, line 381: One thing that might be fun with these is to use the comment support in `hex_literal`, e.g.:

```suggestion
        let ext1 = ExtensionExample::from_der(&hex!("
            300F // 15: SEQUENCE
            0603 // 3:   OBJECT IDENTIFIER basicConstraints (2 5 29 19)
            551D130101FF040530030101FF // etc
        ")).unwrap();
```

der/tests/derive.rs
```diff
@@ -206,6 +206,98 @@ mod sequence {
...
+        #[asn1(
+            context_specific = "2",
+            default = "default_false_example",
+            tag_mode = "IMPLICIT"
+        )]
+        pub only_contains_cacerts: bool,
+
+        // onlySomeReasons            [3] ReasonFlags OPTIONAL,
+        //#[asn1(context_specific="3", optional="true", tag_mode="IMPLICIT")]
+        //pub only_some_reasons: Option<ReasonFlags<'a>>,
```
tarcieri, line 253: Is there a reason why this is commented out?

der/tests/derive.rs
```diff
@@ -261,6 +353,60 @@ mod sequence {
...
+        .unwrap();
+        assert_eq!(ext1.critical, true);
+
+        let ext2 = ExtensionExample::from_der(&hex!("
+        301F                                            //  0  31: SEQUENCE {
+        0603551D23                                      //  2   3:   OBJECT IDENTIFIER authorityKeyIdentifier (2 5 29 35)
+        0418                                            //  7  24:   OCTET STRING, encapsulates {
+        3016                                            //  9  22:     SEQUENCE {
+        8014E47D5FD15C9586082C05AEBE75B665A7D95DA866    // 11  20:       [0] E4 7D 5F D1 5C 95 86 08 2C 05 AE BE 75 B6 65 A7 D9 5D A8 66
+        "))
```
tarcieri, line 405: Nice! But maybe indent these consistently?

der/tests/derive.rs
carl-wallace: The types for the commented out fields are not in the main x509 repo yet. Replicating those here seemed heavier than the value they would add to the example. The structure without the comments will land as x509 bits are pushed.

der/tests/derive.rs
carl-wallace: They both use minimum tab stop for the left hand code. I can bump the first example out to match the second if you like.

der/tests/derive.rs
tarcieri: Aah, ok. Maybe a TODO then?

der/tests/derive.rs
tarcieri: Perhaps indented like so?
```suggestion
            "
            300F        //  0  15: SEQUENCE {
            0603551D13  //  2   3:   OBJECT IDENTIFIER basicConstraints (2 5 29 19)
            0101FF      //  7   1:   BOOLEAN TRUE
            0405        //  10   5:   OCTET STRING, encapsulates {
            3003        //  12   3:     SEQUENCE {
            0101FF      //  14   1:       BOOLEAN TRUE
            "
        ))
        .unwrap();
        assert_eq!(ext1.critical, true);

        let ext2 = ExtensionExample::from_der(&hex!(
            "
            301F                                            //  0  31: SEQUENCE {
            0603551D23                                      //  2   3:   OBJECT IDENTIFIER authorityKeyIdentifier (2 5 29 35)
            0418                                            //  7  24:   OCTET STRING, encapsulates {
            3016                                            //  9  22:     SEQUENCE {
            8014E47D5FD15C9586082C05AEBE75B665A7D95DA866    // 11  20:       [0] E4 7D 5F D1 5C 95 86 08 2C 05 AE BE 75 B6 65 A7 D9 5D A8 66
            "
        ))
```

Not sure if rustfmt will be happy with that, but...

tarcieri (approved):

PR #250, opened 2021-11-21
Add `tls_codec(with = "module")` attribute to customize how struct fields are handled by derive macro

stephaneyfx:
This attribute allows to designate a module where to find functions to serialize, deserialize or get the serialized size of a field. This is similar to [serde's attribute](https://serde.rs/field-attrs.html#with).

The main motivation is to avoid having to resort to a manual implementation of the `tls_codec` traits for a struct as soon as the type of a field does not implement them, which is more error-prone and less maintainable.

This PR also makes structs and tuple structs share more code, and fixes a derive macro that relied on a trait being in scope.

Thank you!

cc @tomleavy

### DIFF
--- a/tls_codec/derive/src/lib.rs
+++ b/tls_codec/derive/src/lib.rs
@@ -1,4 +1,4 @@
-//! Derive macros for traits in `tls_codec`
+//! # Derive macros for traits in `tls_codec`
 //!
 //! The following attribute is available:
 //!
@@ -50,8 +50,12 @@ use syn::{
     Member, Meta, NestedMeta, Type,
 };
 
+/// Attribute name to identify attributes to be processed by derive-macros in this crate.
 const ATTR_IDENT: &str = "tls_codec";
 
+/// Prefix to add to `tls_codec` functions
+///
+/// This is either `<Type as Trait>` or a custom module containing the functions.
 #[derive(Clone)]
 enum Prefix {
     Type(Type),
@@ -59,6 +63,7 @@ enum Prefix {
 }
 
 impl Prefix {
+    /// Returns the path prefix to use for functions from the given trait.
     fn for_trait(&self, trait_name: &str) -> TokenStream2 {
         let trait_name = Ident::new(trait_name, Span::call_site());
         match self {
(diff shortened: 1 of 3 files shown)


tls_codec/derive/src/lib.rs
```diff
@@ -1,23 +1,80 @@
...
+const ATTR_IDENT: &str = "tls_codec";
+
+#[derive(Clone)]
+enum Prefix {
+    Type(Type),
+    Custom(ExprPath),
+}
+
+impl Prefix {
+    fn for_trait(&self, trait_name: &str) -> TokenStream2 {
```
franziskuskiefer, line 62: I know this file doesn't have a lot of comments yet. But let's start adding some.

tls_codec/derive/src/lib.rs
```diff
@@ -43,100 +100,116 @@ struct Enum {
...
 
-                let field_paths: Vec<TokenStream2> = paths.collect();
-                Ok(TlsStruct::TupleStruct(TupleStruct {
-                    call_site,
-                    ident: ident.clone(),
-                    generics: generics.clone(),
-                    field_indices,
-                    field_paths,
-                }))
+fn function_prefix(field: &Field) -> Result<Prefix> {
```
franziskuskiefer, line 154:
```suggestion
/// Get the [`Prefix`] for a field, i.e. the type itself
/// or a module that implements the tls_codec functions.
fn function_prefix(field: &Field) -> Result<Prefix> {
```

franziskuskiefer (changes requested):
Thanks!
That's a really nice addition. And thanks for the cleanup 🙈

Only a couple small nits to add a little more documentation. Otherwise this is good to go.

tls_codec/derive/src/lib.rs
stephaneyfx: I added more comments. Please let me know if that's ok. Thank you!

stephaneyfx:
You're welcome and thank you for the review. I addressed your comments.

franziskuskiefer (approved):
Thanks! lgtm

tomleavy:
Hey @franziskuskiefer question for you, how is the current encoding / decoding performance? If there are places you think we can help make improvements let us know.

PR #258, opened 2021-11-29
x509: Certificate

carl-wallace:
- added certificate decode/encode support
- added tlv_slice to Decoder to support accessing undecoded fields in a SEQUENCE, i.e., full tag/length/value
- bumped OID size to accomodate oversized OIDs, i.e., like those in certificate template extensions from Microsoft CAs


const-oid/src/lib.rs
```diff
@@ -55,7 +55,7 @@ pub struct ObjectIdentifier {
 #[allow(clippy::len_without_is_empty)]
 impl ObjectIdentifier {
     /// Maximum size of a BER/DER-encoded OID in bytes.
-    pub const MAX_SIZE: usize = 31; // 24-bytes total w\ 1-byte length
+    pub const MAX_SIZE: usize = 40; // 33-bytes total w\ 1-byte length
```
tarcieri, line 58: I was hoping to keep the total length a multiple of 4. If you really need 40 instead of 39, I'd suggest bumping this all the way to 43.

x509/src/certificate.rs
```diff
@@ -0,0 +1,243 @@
...
+
+/// returns false in support of boolean DEFAULT fields
+pub fn default_false() -> bool {
+    false
+}
+
+/// returns false in support of integer DEFAULT fields set to 0
+pub fn default_zero() -> u32 {
+    0
+}
```
tarcieri, line 21: These all make me think there should be some shorthand when using the proc macro attributes for using `Default` as the default in these cases, possibly `default = true`.

const-oid/src/lib.rs
carl-wallace: Changed to 39, which works for the largest observed values across the cert collections I ran against.

x509/src/certificate.rs
carl-wallace: There will be more complicated defaults coming soon (i.e AlgorithmIdentifiers) so this approach would likely need to remain but could be augmented with support for simpler cases.

der/src/decoder.rs
```diff
@@ -283,6 +283,29 @@ impl<'a> Decoder<'a> {
...
+        let pos = self.position();
+        let tag = Tag::decode(self)?;
+
+        let element_length = Length::decode(self).map_err(|e| {
+            if e.kind() == ErrorKind::Overlength {
+                ErrorKind::Length { tag }.into()
+            } else {
+                e
+            }
+        })?;
```
tarcieri, line 301: I think all of this can be replaced by...

```suggestion
       let header = self.peek_header()?;
       let header_len = header.encoded_len()?;
       let element_length = header.length;
```

This will also eliminate the need to use `self.position = pos;` to reset the internal position, as you can calculate the length of the TLV production using the above.

der/src/decoder.rs
```diff
@@ -283,6 +283,13 @@ impl<'a> Decoder<'a> {
         Ok(self.bytes.ok_or(ErrorKind::Failed)?.len())
     }
 
+    /// Obtain a slice of bytes contain a complete TLV production suitable for parsing later.
+    pub fn tlv_slice(&mut self) -> Result<&'a [u8]> {
+        let header = self.peek_header()?;
+        let header_len = self.peek_header()?.encoded_len()?;
```
tarcieri, line 289: This is calling `peek_header()` twice and thus parsing the header twice.

```suggestion
        let header_len = header.encoded_len()?;
```

der/src/decoder.rs
```diff
@@ -283,6 +283,13 @@ impl<'a> Decoder<'a> {
         Ok(self.bytes.ok_or(ErrorKind::Failed)?.len())
     }
 
+    /// Obtain a slice of bytes contain a complete TLV production suitable for parsing later.
+    pub fn tlv_slice(&mut self) -> Result<&'a [u8]> {
```
tarcieri, line 287: Generally other methods in these crates that operate over a slice of bytes are named `*_bytes`. How about:

```suggestion
    pub fn tlv_bytes(&mut self) -> Result<&'a [u8]> {
```

x509/src/certificate.rs
```diff
@@ -0,0 +1,243 @@
...
+            self.subject_unique_id
+        ))?;
+        if let Some(exts) = self.extensions.as_ref() {
+            for (i, e) in exts.iter().enumerate() {
+                f.write_fmt(format_args!("\tExtension #{}: {:?}\n", i, e))?;
+            }
+        } else {
+            f.write_fmt(format_args!("\tExtensions: None\n"))?;
+        }
+        Ok(())
```
tarcieri, line 179: Not something that needs to be addressed in this PR per se, but I think it would make sense to use the [`DebugStruct`](https://doc.rust-lang.org/stable/std/fmt/struct.DebugStruct.html) builder type to build this output, even if some of the fields have customized outputs.

Additionally, it would probably make sense to push the customized output formatting to the `Debug` impls of the individual field types wherever possible.

tarcieri (approved):

x509/src/certificate.rs
carl-wallace: OK. Will address in PR that adds extension decoders and encoders.

PR #263, opened 2021-11-30
Clean up bounds in tls_codec

stephaneyfx:
This PR cleans up unnecessary bounds in `tls_codec` (e.g. generic `Size` implementations requiring `Serialize`).
Thank you.

### DIFF
--- a/tls_codec/derive/src/lib.rs
+++ b/tls_codec/derive/src/lib.rs
@@ -46,8 +46,8 @@ use quote::quote;
 use syn::{
     self, parenthesized,
     parse::{ParseStream, Parser, Result},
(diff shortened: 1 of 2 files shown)


franziskuskiefer (approved):
Thanks for the cleanup!
I was first a little confused by some of the formatting changes, but looks good to me!

franziskuskiefer:
> You're welcome and thank you for the review! I have rustfmt set up to run on save, so any formatting change should be due to rustfmt. I am happy to reformat however you like if needed.

No, looks fine! I double checked with rustfmt at it's happy with the formatting. That's all I need 🙂

PR #264, opened 2021-11-30
Implement `tls_codec` traits for the unit type

stephaneyfx:
This allows to derive the `tls_codec` traits for types that contain fields of type unit (e.g. a private unit field to prevent direct construction in other modules).
Thank you.


franziskuskiefer (approved):
Good idea. Thanks!

stephaneyfx:
Thank you for the review!

PR #265, opened 2021-11-30
Derive `tls_codec` traits for enumerations with data

stephaneyfx:
This PR adds the ability to derive the `tls_codec` traits for enumerations with data. It also adds an attribute to specify what the discriminant should be. The `tls_codec_derive` crate documentation provides more details.
Thank you.


franziskuskiefer:
Thanks for the PR.
There are two different things in here.
The discriminant change looks good to me.

The change about enumerations with data is a little more intricate and I'm not sure about what to do about it yet. The TLS representation language doesn't support anything like this but only simple [enumerated](https://datatracker.ietf.org/doc/html/rfc8446#section-3.5) (well, there's nothing like this in C). So I wonder how you use this or why you think it is necessary.
We have a somewhat related PR open (#125) that tries to do tagged enums as used in MLS.
I'd like to keep the crate in line with what's specified without custom extensions. For #125 the idea is to put this into an IETF draft for implementation considerations for MLS (see the linked issues/PRs in #125 and in particular https://github.com/kkohbrok/mls-implementation-considerations/pull/2).
Maybe there's a similar reasoning for the change here, or maybe we could even find a common way of doing this/

stephaneyfx:
Thank you for reviewing.

The change adding a `#[tls_codec(discriminant = 123)]` attribute is rather useless on its own AFAICT given that C-style enumerations already support assigning values to variants in Rust.

This PR is based on [section 3.8 (Variants) of the TLS 1.3 RFC](https://datatracker.ietf.org/doc/html/rfc8446#section-3.8).
> Defined structures may have variants based on some knowledge that is
   available within the environment.  The selector must be an enumerated
   type that defines the possible variants the structure defines.

```text
      struct {
          VariantTag type;
          select (VariantRecord.type) {
              case apple:  V1;
              case orange: V2;
          };
      } VariantRecord;
```

This description and example from the TLS RFC match what enumerations (with variants with data) are in Rust IMO, and this PR aims to implement derivation of the TLS serialization traits to do exactly that, unless I am misinterpreting the RFC.

Does it make sense? Do you interpret [section 3.8 of the TLS RFC](https://datatracker.ietf.org/doc/html/rfc8446#section-3.8) differently?

Thank you.

franziskuskiefer:
Oh I see. So you want to represent the variants as enums, not structs? I'm not sure I fully understand how you want to use the changes here for variants. How would the `VariantRecord` example look like? Can you add it to the test?

I'd like to be able to write something like this for it:

```rust
#[derive(TlsDeserialize, TlsSerialize, TlsSize)]
struct AppleRecord {}
#[derive(TlsDeserialize, TlsSerialize, TlsSize)]
struct OrangeRecord {}

#[derive(TlsDeserialize, TlsSerialize, TlsSize)]
#[repr(u8)]
enum VariantTag {
    Apple,
    Orange,
}

#[derive(TlsDeserialize, TlsSerialize, TlsSize)]
enum VariantRecordValue {
    Apple(AppleRecord),
    Orange(OrangeRecord),
}

#[derive(TlsDeserialize, TlsSerialize, TlsSize)]
struct VariantRecord {
    // I added something here that we have to handle as well.
    // Usually structs have a bunch of values and then a variant somewhere.
    some_value: TlsVecU8<u8>,
    variant_tag: VariantTag,
    variant: VariantRecordValue,
}
```

stephaneyfx:
Sorry for my delayed response.

Your example would look like this:
```rust
#[derive(TlsDeserialize, TlsSerialize, TlsSize)]
struct AppleRecord {}
#[derive(TlsDeserialize, TlsSerialize, TlsSize)]
struct OrangeRecord {}

#[derive(TlsDeserialize, TlsSerialize, TlsSize)]
#[repr(u8)]
enum VariantRecordValue {
    Apple(AppleRecord),
    Orange(OrangeRecord),
}

#[derive(TlsDeserialize, TlsSerialize, TlsSize)]
struct VariantRecord {
    some_value: TlsVecU8<u8>,
    variant: VariantRecordValue,
}
```

There is no need for the extra field `variant_tag` in Rust. It is already embedded in the `variant` field as the discriminant of the Rust enum `VariantRecordValue`. The above representation prevents non-sensical values which the original code allows, e.g.:
```rust
let invalid = VariantRecord {
    some_value: Default::default(),
    variant_tag: VariantTag::Apple,
    variant: VariantRecordValue::Orange(OrangeRecord {}),
};
```

There's also nothing special to do to support a struct with some fields followed by a variant field, as demonstrated in your adapted example.

The `#[repr]` attribute on the Rust enum definition allows to specify the type to use for the discriminant. Custom discriminant values for serialization can be specified using the `#[tls_codec(discriminant = 123)]` attribute for enums with data, e.g.:
```rust
#[derive(TlsDeserialize, TlsSerialize, TlsSize)]
#[repr(u8)]
enum VariantRecordValue {
    #[tls_codec(discriminant = 123)]
    Apple(AppleRecord),
    #[tls_codec(discriminant = 124)]
    Orange(OrangeRecord),
}
```

Does that clarify the approach taken by this PR and how TLS variants map to Rust enums or regular Rust structs with a field of type a Rust enum? Please let me know if it is still unclear or you have concerns regarding this approach.

Thank you!

franziskuskiefer:
> Does that clarify the approach taken by this PR and how TLS variants map to Rust enums or regular Rust structs with a field of type a Rust enum? Please let me know if it is still unclear or you have concerns regarding this approach.

Thanks for the explanation @stephaneyfx 
I think I approached this from the other end and got confused by the approach. This certainly works for the example. But it only works when the variant type directly precedes the value. This isn't always the case.
Take the `Handshake` from TLS 1.3 as example

```
      enum {
          client_hello(1),
          ...
          (255)
      } HandshakeType;

      struct {
          HandshakeType msg_type;    /* handshake type */
          uint24 length;             /* remaining bytes in message */
          select (Handshake.msg_type) {
              case client_hello:          ClientHello;
              ...
          };
      } Handshake;
```

With the approach in this PR this would be encoded as the following struct.
```
      struct {
          uint24 length;             /* remaining bytes in message */
          HandshakeType msg_type;    /* handshake type */
          select (Handshake.msg_type) {
              case client_hello:          ClientHello;
              ...
          };
      } Handshake;
```

Sometimes the tag isn't even in the struct itself but defined in a different struct, e.g

```
      struct {
          select (Handshake.msg_type) {
              case client_hello:
                   ProtocolVersion versions<2..254>;

              case server_hello: /* and HelloRetryRequest */
                   ProtocolVersion selected_version;
          };
      } SupportedVersions;
```

That's why I think we need the tag/type to be explicit. Because we don't know where it is coming from.
The approach in this PR works for certain structs but I'm afraid people will start to use it for structs where it produces wrong (de)serialisations.

stephaneyfx:
Thank you for your feedback. You are correct that the support for Rust enums this PR adds to the derive macros does not cover all use cases of TLS variants.

I would argue that the in-memory representation does not have to match the wire format, and that is already the case for multiple scenarios:
- A structure with a field of type `TlsByteVecU8` contains pointers to a byte buffer located elsewhere in memory, whereas this memory is embedded in the struct representation in the wire format.
- Items in vectors of enumerations or vectors all have the same size (they occupy the same amount of bytes within the array, but of course they can have pointers to other buffers), whereas in the wire format, all items are serialized contiguously and only use the amount of bytes they need.

This is how the `Handshake` type should be handled in my opinion (not tested; may not be correct):
```rust
enum Handshake {
    ClientHello(ClientHello),
    ServerHello(ServerHello),
    // ...
}

impl Size for Handshake {
    fn tls_serialized_len(&self) -> usize {
        1 // Handshake type
        + 3 // Remaining length (24 bits)
        + match self {
            Handshake::ClientHello(hello) => hello.tls_serialized_len(),
            Handshake::ServerHello(hello) => hello.tls_serialized_len(),
            // ...
        }
    }
}

impl Serialize for Handshake {
    fn tls_serialize<W: Write>(&self, writer: &mut W) -> Result<usize, tls_codec::Error> {
        let ty = match self {
            Handshake::ClientHello(_) => 1u8,
            Handshake::ServerHello(_) => 2u8,
            // ...
        };
        ty.tls_serialize(writer)?;
        let remaining = self.tls_serialized_len() - 4;
        assert!(remaining < 1 << 24);
        writer.write_all(&remaining.to_be_bytes()[..3])?;
        let res = match self {
            Handshake::ClientHello(hello) => hello.tls_serialize(writer),
            Handshake::ServerHello(hello) => hello.tls_serialize(writer),
            // ...
        };
        Ok(4 + res?)
    }
}

impl Deserialize for Handshake {
    fn tls_deserialize<R: Read>(bytes: &mut R) -> Result<Self, tls_codec::Error> {
        let ty = u8::tls_deserialize(bytes)?;
        bytes.read_exact(&mut [0; 3])?;
        match ty {
            1 => Ok(Handshake::ClientHello(ClientHello::tls_deserialize(bytes)?)),
            2 => Ok(Handshake::ServerHello(ServerHello::tls_deserialize(bytes)?)),
            // ...
            _ => Err(tls_code::Error::DecodingError(format!("Unknown handshake type {}", ty))),
        }
    }
}

#[derive(TlsDeserialize, TlsSerialize, TlsSize)]
struct ClientHello {
    // ...
}

#[derive(TlsDeserialize, TlsSerialize, TlsSize)]
struct ServerHello {
    // ...
}
```

This PR adds support to derive the TLS serialization traits for the common case where the discriminant is serialized just before the variant data. Just like derive macros in `serde`, it does not support any arbitrary format. It is meant as a simple way to generate implementations for common cases. When that does not match the required format, a manual implementation is still possible as shown above. And in this handshake example, attempting to replicate the wire format in the Rust struct definitions has the disadvantage of allowing to represent invalid values in Rust. For example, with:
```rust
struct Handshake {
    msg_type: HandshakeType,
    length: u32,
    record: HandshakeRecord,
}

enum HandshakeRecord {
    ClientHello(ClientHello),
    ServerHello(ServerHello),
    // ...
}
```
, invalid values of type `Handshake` can be constructed:
- `msg_type` may not match the value of `record`
- `length` may not match the serialized length of `record`.

The TLS variant format is very flexible, and this derive macro (like most derive macros) only aims at supporting simple common cases with some level of customization (with the `repr` and `discriminant` attributes).

Does the above explanation make sense? Thank you for reading this wall of text!

franziskuskiefer:
> Does the above explanation make sense? Thank you for reading this wall of text!

Yes, totally makes sense. I'm just wondering what we can do to prevent people from misusing this.
- We could put this behind a feature flag and people who want to use it have to enable it, making it more likely they know what they are doing.
- We add a large warning to the docs making it clear how this can and can not be used (maybe with some of the example from above).

I think we should certainly do the latter. I'm open to opinions of requiring a feature for this.

stephaneyfx:
I added a warning to the documentation as suggested -- I didn't add the example because it felt too long and I was not sure how to truncate it. If you think an example is needed after reading the warning, please let me know.

I have not added any feature flag (I think the warning should be enough), but if you think a feature flag is warranted, please let me know and I can work on that.

Thank you!

franziskuskiefer (approved):
Thanks!
I think the warning now makes clear how this can and can not be used.

franziskuskiefer:
Thanks again!
I published 0.2.0-pre.3 of the crate with these changes.

stephaneyfx:
Thank you very much for the feedback and publishing a new version so quickly!

PR #271, opened 2021-12-02
Added support for GeneralName (minus x400Address and ediPartyName) and GeneralNames

carl-wallace:


x509/src/general_name.rs
```diff
@@ -0,0 +1,417 @@
+/// GeneralNames as defined in [RFC 5280 Section 4.2.1.6].
+use crate::alloc::string::ToString;
```
tarcieri, line 2: Toplevel doc comments should use `//!`

Also the `crate::` on the front of this import is redundant

```suggestion
//! GeneralNames as defined in [RFC 5280 Section 4.2.1.6].

use alloc::string::ToString;
```

x509/src/general_name.rs
```diff
@@ -0,0 +1,417 @@
...
+}
+
+impl<'a> Sequence<'a> for OtherName<'a> {
+    fn fields<F, T>(&self, f: F) -> ::der::Result<T>
+    where
+        F: FnOnce(&[&dyn der::Encodable]) -> ::der::Result<T>,
+    {
+        f(&[&self.type_id, &self.value])
+    }
+}
```
tarcieri, line 40: Was there an issue deriving `Sequence` on this type?

x509/src/general_name.rs
```diff
@@ -0,0 +1,417 @@
...
+
+/// GeneralNames as defined in [RFC 5280 Section 4.2.1.6] in support of the Subject Alternative Name extension.
+///
+/// ```text
+/// GeneralNames ::= SEQUENCE SIZE (1..MAX) OF GeneralName
+/// ```
+///
+/// [RFC 5280 Section 4.2.1.6]: https://datatracker.ietf.org/doc/html/rfc5280#section-4.2.1.6
+pub type GeneralNames<'a> = alloc::vec::Vec<GeneralName<'a>>;
```
tarcieri, line 49: I'd suggest adding a `use` directive for `alloc::vec::Vec` at the top of the file

x509/src/general_name.rs
```diff
@@ -0,0 +1,417 @@
...
+                let ia5 = decoder
+                    .context_specific::<Ia5String<'_>>(RFC822_NAME_TAG, TagMode::Implicit)?;
+                Ok(GeneralName::Rfc822Name(ia5.unwrap()))
+            }
+            0x82 => {
+                let ia5 =
+                    decoder.context_specific::<Ia5String<'_>>(DNS_NAME_TAG, TagMode::Implicit)?;
+                Ok(GeneralName::DnsName(ia5.unwrap()))
+            }
+            0xA3 => unimplemented!(),
```
tarcieri, line 125: Are these a TODO? I think it's bad to panic as opposed to returning an error.

x509/src/general_name.rs
carl-wallace: Fixed.

x509/src/general_name.rs
carl-wallace: Yes. If I comment out the above and add Sequence back to the derive, the following results:
```
the trait `DecodeValue<'_>` is not implemented for `OtherName<'_>`
```

x509/src/general_name.rs
carl-wallace: Fixed.

x509/src/general_name.rs
tarcieri: Weird. Guess I'll have to take a look at the `cargo expand` output and see why the derived implementation differs, but I can do that after merging this.

x509/src/general_name.rs
carl-wallace: Bah. Yes, those should be errors. I see no reason to mark as TODO but forgot to fix that. FIxed now.

PR #274, opened 2021-12-03
ssh-key v0.1.0

tarcieri:
Initial release

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -933,7 +933,7 @@ dependencies = [
 
 [[package]]
 name = "ssh-key"
-version = "0.0.0"
+version = "0.1.0"
 dependencies = [
  "base64ct",
  "hex-literal",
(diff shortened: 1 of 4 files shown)


kjvalencik:
@tarcieri We've been working on a parser for these key formats internally. The plan was to open source when complete, but since this was added, maybe you would be interested in upstreaming it here instead? Private key deserialization is complete.

It uses the RustCrypto crates for decryption; although that piece is still a work in progress. It's not clear if the ChaCha20Poly1305 cipher is compatible.

tarcieri:
@kjvalencik sounds interesting!

As a bit more background towards that, I would like to get `pem-rfc7468` modified to use `base64ct::Decoder` for incremental decoding, but that would also require adding support for handling linewrapped Base64 to the decoder.

With both of those implemented, it should be possible to use `pem-rfc7468` to support the OpenSSH `BEGIN OPENSSH PRIVATE KEY` encoding which linewraps at an awkward 70 Base64 characters, and also write a private key decoder with zero heap allocations, at least for ECDSA and Ed25519.

kjvalencik:
Sounds like a solid plan. The private key format can definitely be parsed without allocations. Additionally, `block_padding` might need to be updated to support the padding OpenSSH uses (padding is an incrementing count--`1, 2, 3, 4...`). Although, I don't think there's a way to know how much padding there is without parsing the key.

PR #275, opened 2021-12-03
add support for a variety of extensions

carl-wallace:


x509/src/extensions_utils.rs
```diff
@@ -0,0 +1,142 @@
...
+            KeyUsageValues::DigitalSignature => write!(f, "DigitalSignature"),
+            KeyUsageValues::NonRepudiation => write!(f, "NonRepudiation"),
+            KeyUsageValues::KeyEncipherment => write!(f, "KeyEncipherment"),
+            KeyUsageValues::DataEncipherment => write!(f, "DataEncipherment"),
+            KeyUsageValues::KeyAgreement => write!(f, "KeyAgreement"),
+            KeyUsageValues::KeyCertSign => write!(f, "KeyCertSign"),
+            KeyUsageValues::CRLSign => write!(f, "CRLSign"),
+            KeyUsageValues::EncipherOnly => write!(f, "EncipherOnly"),
+            KeyUsageValues::DecipherOnly => write!(f, "DecipherOnly"),
+        }
```
tarcieri, line 42: For something like this where there's a 1:1 mapping of variants to strings, I often find it helpful to have an `fn as_str(self) -> &'static str` method providing that mapping.

If you have that, the body of `Display::fmt` can just be:

```suggestion
        f.write_str(self.as_str())
```

x509/src/extensions_utils.rs
```diff
@@ -0,0 +1,142 @@
+//! Utility functions and enums related to X.509 extensions
+
+use alloc::vec::Vec;
+use core::fmt;
+use der::asn1::BitString;
+
+/// Enum representing values from the KeyUsage structure
+#[derive(Clone, Debug, Eq, PartialEq)]
```
tarcieri, line 8:
```suggestion
#[derive(Copy, Clone, Debug, Eq, PartialEq)]
```

x509/src/lib.rs
```diff
@@ -16,14 +16,17 @@ extern crate std;
 
 mod attribute;
 mod certificate;
+pub mod extensions_utils;
 mod general_name;
+pub mod pkix_extensions;
+pub mod pkix_oids;
```
tarcieri, line 22: Would it make sense to have `pub mod pkix` with separate `extensions` and `oids` submodules?

x509/src/pkix_extensions.rs
```diff
@@ -0,0 +1,1041 @@
...
+/// - [`PKIX_KP_CODESIGNING`](constant.PKIX_KP_CODESIGNING.html),
+/// - [`PKIX_KP_EMAILPROTECTION`](constant.PKIX_KP_EMAILPROTECTION.html),
+/// - [`PKIX_KP_TIMESTAMPING`](constant.PKIX_KP_TIMESTAMPING.html),
+///
+/// ```text
+/// ExtKeyUsageSyntax ::= SEQUENCE SIZE (1..MAX) OF KeyPurposeId
+/// ```
+///
+/// [RFC 5280 Section 4.2.1.12]: https://datatracker.ietf.org/doc/html/rfc5280#section-4.2.1.12
+pub type ExtendedKeyUsage<'a> = alloc::vec::Vec<ObjectIdentifier>;
```
tarcieri, line 62: Would suggest a `use alloc::vec::Vec` import at the top of the module, especially since there are multiple `Vec`s declared (e.g. SubjectDirectoryAttributes as well)

x509/src/pkix_extensions.rs
```diff
@@ -0,0 +1,1041 @@
...
+///    BasicConstraints ::= SEQUENCE {
+///         cA                      BOOLEAN DEFAULT FALSE,
+///         pathLenConstraint       INTEGER (0..MAX) OPTIONAL }
+/// ```
+///
+/// [RFC 5280 Section 4.2.1.9]: https://datatracker.ietf.org/doc/html/rfc5280#section-4.2.1.9
+#[derive(Clone, Debug, Eq, PartialEq, Sequence)]
+pub struct BasicConstraints {
+    /// cA                      BOOLEAN DEFAULT FALSE,
+    #[asn1(default = "default_false")]
```
tarcieri, line 112: Curious if this "just works" as an alternative:

```suggestion
    #[asn1(default = "Default::default")]
```

If so, you could use that pattern as an alternative to having to write a custom function that returns the `Default` for a given type, with no additional changes to the proc macro.

x509/src/extensions_utils.rs
carl-wallace: Done.

x509/src/extensions_utils.rs
carl-wallace: Done.

x509/src/pkix_extensions.rs
carl-wallace: Done (sorry about this one, meant to make that mod after the same comment on general_name and lost that thread today).

x509/src/pkix_extensions.rs
carl-wallace: That did just work for here. Still have not added anything with default alg ID yet (CMS support will generate those).

x509/src/lib.rs
carl-wallace: Offhand, I don't think so. pkix_extensions should probably be named extensions (or cert_extensions with CRL extensions split off) if we are going to regroup. The RDN OIDs in the pkix_oids file are used indirectly by the certificate module too so there's affinity with the sibling certificate module too.

PR #289, opened 2021-12-19
spki: impl `ValueOrd` for `AlgorithmIdentifier`

### LINKED ISSUE #288: AlgorithmIdentifier needs to implement DerOrd trait
Can the AlgorithmIdentifier type be modified to implement the DerOrd trait so that a SET OF AlgorithmIdentifier is allowed?

Are there any other types in this crate that should also implement DerOrd so that SET OF can be used with them?

tarcieri:
Allows for storing `AlgorithmIdentifier`s in SET OF types.

Closes #288

### DIFF
--- a/spki/src/algorithm.rs
+++ b/spki/src/algorithm.rs
@@ -1,9 +1,10 @@
 //! X.509 `AlgorithmIdentifier`
 
 use crate::{Error, Result};
+use core::cmp::Ordering;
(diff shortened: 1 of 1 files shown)


rod-chapman:
Looks good to me. Thanks for the very swift response.

tarcieri:
Released in v0.5.3

PR #291, opened 2021-12-19
tls_codec: preliminary `no_std` support

tarcieri:
Adds initial support for `no_std` targets (currently with a hard `alloc` dependency) by gating all `std`-related functionality on a newly added eponymous crate feature.

Unfortunately all serialization/deserialization is currently performed in terms of the `Read` and `Write` traits from `std::io`, so this doesn't result in something immediately useful on `no_std` targets. However, it does provide the initial boilerplate and first steps towards making that possible.

### DIFF
--- a/.github/workflows/tls_codec.yml
+++ b/.github/workflows/tls_codec.yml
@@ -40,6 +40,25 @@ jobs:
       - run: cargo build --target ${{ matrix.target }} --release --features serde_serialize
       - run: cargo build --target ${{ matrix.target }} --release --all-features
 
+  no_std:
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        rust:
+          - 1.56.0 # MSRV
+          - stable
+        target:
+          - thumbv7em-none-eabi
+    steps:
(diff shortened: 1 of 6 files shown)


tarcieri:
The next steps I'd suggest after this would be defining replacement traits for `std::io::{Read, Write}` which provide the subset of functionality currently used by this crate, so that they can always be available regardless of the target.

When the `std` feature is enabled, blanket impls of these traits can be defined for the corresponding `std::io` traits, so everything continues to work with them.

franziskuskiefer (approved):
Thanks! Definitely a good start for tackling this.

Right now there's no good way to do this without `Read` and `Write` I'm afraid.
There's some discussion around this happening. It's all over the place but https://github.com/rust-lang/rfcs/issues/2262 looks like a good issue to follow.

If we want to have something done before there's a solution that allows us using the traits, we could add a separate API that operates on byte slices/vectors.

tarcieri:
Cool. Will merge this and we can follow up on what to do about new `no_std` friendly APIs.

PR #324, opened 2022-01-12
base16ct: initial crate

tarcieri:
Constant-time implementation of Base16 (a.k.a. hexadecimal) based on this C++ implementation:

https://github.com/Sc00bz/ConstTimeEncoding/blob/master/hex.cpp

Original C++ is MIT licensed. Relicensed as Apache 2.0+MIT with permission from the original author.

### DIFF
--- a/base16ct/src/lib.rs
+++ b/base16ct/src/lib.rs
@@ -1,0 +1,228 @@
+#![no_std]
+#![cfg_attr(docsrs, feature(doc_cfg))]
+#![forbid(unsafe_code)]
+#![warn(missing_docs, rust_2018_idioms)]
+#![doc(
+    html_logo_url = "https://raw.githubusercontent.com/RustCrypto/meta/master/logo.svg",
+    html_favicon_url = "https://raw.githubusercontent.com/RustCrypto/meta/master/logo.svg",
+    html_root_url = "https://docs.rs/base16ct/0.0.0"
+)]
+#![doc = include_str!("../README.md")]
+
+//! # Implementation
+//!
+//! Implemented using integer arithmetic alone without any lookup tables or
+//! data-dependent branches, thereby providing portable "best effort"
+//! constant-time operation.
+//!
+//! Not constant-time with respect to message length (only data).
+//!
+//! Adapted from this C++ implementation:
+//!
+//! <https://github.com/Sc00bz/ConstTimeEncoding/blob/master/hex.cpp>
+//!
+//! Copyright (c) 2014 Steve "Sc00bz" Thomas (steve at tobtu dot com)
+//! Derived code is dual licensed MIT + Apache 2.0 (with permission from @Sc00bz)
+
+#[cfg(feature = "alloc")]
+#[macro_use]
+extern crate alloc;
+#[cfg(feature = "std")]
+extern crate std;
+
+use core::fmt;
+
+#[cfg(feature = "alloc")]
+use alloc::{string::String, vec::Vec};
+
+/// Error type
+#[derive(Clone, Eq, PartialEq, Debug)]
+pub enum Error {
+    /// Invalid encoding of provided Base16 string.
+    InvalidEncoding,
+
+    /// Insufficient output buffer length.
+    InvalidLength,
+}
+
+impl fmt::Display for Error {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        match self {
+            Error::InvalidEncoding => f.write_str("invalid Base16 encoding"),
+            Error::InvalidLength => f.write_str("invalid Base16 length"),
+        }
+    }
+}
+
+#[cfg(feature = "std")]
+impl std::error::Error for Error {}
+
+/// Base16 encoding (a.k.a. hexadecimal)
+#[derive(Copy, Clone, Debug, Default, Eq, Hash, PartialEq, PartialOrd, Ord)]
+pub struct Base16 {
+    /// Upper or lower case
+    case: Case,
+}
+
+impl Base16 {
+    /// Lower case hex: 0-9 a-f
+    pub fn lower_case() -> Base16 {
+        Base16 { case: Case::Lower }
+    }
+
+    /// Upper case hex: 0-9 A-F
+    pub fn upper_case() -> Base16 {
+        Base16 { case: Case::Upper }
+    }
+}
+
+impl Base16 {
+    /// Decode a Base16 (hex) string into the provided destination buffer.
+    pub fn decode<'a>(&self, src: impl AsRef<[u8]>, dst: &'a mut [u8]) -> Result<&'a [u8], Error> {
+        let src = src.as_ref();
+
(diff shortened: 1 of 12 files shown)


base16ct/src/lib.rs
```diff
@@ -0,0 +1,228 @@
+#![no_std]
+#![cfg_attr(docsrs, feature(doc_cfg))]
+#![forbid(unsafe_code)]
+#![warn(missing_docs, rust_2018_idioms)]
+#![doc(
+    html_logo_url = "https://raw.githubusercontent.com/RustCrypto/meta/master/logo.svg",
+    html_favicon_url = "https://raw.githubusercontent.com/RustCrypto/meta/master/logo.svg",
+    html_root_url = "https://docs.rs/base16ct/0.0.0"
+)]
+#![doc = include_str!("../README.md")]
```
newpavlov, line 10: I am not sure it's worth to include the whole readme while it does not contain any code examples. If anything, everything in it except the first 3 paragraphs looks like an unnecessary clutter to me in the docs context.

base16ct/src/lib.rs
```diff
@@ -0,0 +1,228 @@
+#![no_std]
+#![cfg_attr(docsrs, feature(doc_cfg))]
+#![forbid(unsafe_code)]
+#![warn(missing_docs, rust_2018_idioms)]
+#![doc(
+    html_logo_url = "https://raw.githubusercontent.com/RustCrypto/meta/master/logo.svg",
+    html_favicon_url = "https://raw.githubusercontent.com/RustCrypto/meta/master/logo.svg",
```
newpavlov, line 7: I think it's worth to replace the URLs with `https://raw.githubusercontent.com/RustCrypto/media/6ee8e381/logo.svg` in the case if will move the logo for some reason.

base16ct/src/lib.rs
```diff
@@ -0,0 +1,228 @@
...
+//! constant-time operation.
+//!
+//! Not constant-time with respect to message length (only data).
+//!
+//! Adapted from this C++ implementation:
+//!
+//! <https://github.com/Sc00bz/ConstTimeEncoding/blob/master/hex.cpp>
+//!
+//! Copyright (c) 2014 Steve "Sc00bz" Thomas (steve at tobtu dot com)
+//! Derived code is dual licensed MIT + Apache 2.0 (with permission from @Sc00bz)
```
newpavlov, line 25: Maybe use `//` for the last two lines? I don't think we have similar copyright information in our other crate docs.

base16ct/src/lib.rs
```diff
@@ -0,0 +1,228 @@
...
+
+#[cfg(feature = "std")]
+impl std::error::Error for Error {}
+
+/// Base16 encoding (a.k.a. hexadecimal)
+#[derive(Copy, Clone, Debug, Default, Eq, Hash, PartialEq, PartialOrd, Ord)]
+pub struct Base16 {
+    /// Upper or lower case
+    case: Case,
+}
```
newpavlov, line 65: Are you sure about using runtime branching instead of introducing two types? Also I wonder whether in practice we may need to support decoding of mixed values (i.e. `ABcd12`).

base16ct/src/lib.rs
```diff
@@ -0,0 +1,228 @@
...
+        let src = src.as_ref();
+
+        let dst_length = Self::decoded_len(src)?;
+        if dst_length > dst.len() {
+            return Err(Error::InvalidLength);
+        }
+
+        let mut err: usize = 0;
+
+        for (i, dst_byte) in dst.iter_mut().enumerate().take(dst_length) {
```
newpavlov, line 91: Personally I think `dst[..dst_len].iter_mut().enumerate()` is easier to read. Alternatively you could use something like this:
```rust
for (src, dst) in src.chunks_exact(2).zip(dst.iter_mut()) { .. }
```
Zipping should handle the `dst` truncation.

base16ct/src/lib.rs
```diff
@@ -0,0 +1,228 @@
...
+    pub fn decode_vec(&self, input: impl AsRef<[u8]>) -> Result<Vec<u8>, Error> {
+        let mut output = vec![0u8; Self::decoded_len(input.as_ref())?];
+        self.decode(input, &mut output)?;
+        Ok(output)
+    }
+
+    /// Compute decoded length of the given input.
+    pub fn decoded_len(bytes: &[u8]) -> Result<usize, Error> {
+        if bytes.len() & 1 == 0 {
+            Ok(bytes.len() >> 1)
```
newpavlov, line 118: It's fine to simply use `bytes.len() / 2`, compiler will replace it with the shift itself. :)

base16ct/src/lib.rs
```diff
@@ -0,0 +1,228 @@
...
+    /// If `input` length is greater than `usize::MAX/2`.
+    #[cfg(feature = "alloc")]
+    #[cfg_attr(docsrs, doc(cfg(feature = "alloc")))]
+    pub fn encode_string(&self, input: &[u8]) -> String {
+        let elen = Self::encoded_len(input);
+        let mut dst = vec![0u8; elen];
+        let res = self.encode(input, &mut dst).expect("encoding error");
+
+        debug_assert_eq!(elen, res.len());
+        String::from_utf8(dst).expect("character encoding error")
```
newpavlov, line 154: I think we can use `unsafe { String::from_utf8_unchecked(dst) }` like we do in `base64ct` to remove unreachable panic.

newpavlov (commented):
Later I will try to inspect the code in godbolt to see if there are some easily removable unreachable panics.

base16ct/src/lib.rs
newpavlov: Also the last suggestion helps compiler to remove unreachable panics.

newpavlov:
After replacing the encoding and decoding loops to the zip-based code and `String` conversion all functions are properly panic-free!

base16ct/tests/lib.rs
```diff
@@ -0,0 +1,84 @@
+//! Integration tests.
+
+#![cfg(feature = "alloc")]
```
newpavlov, line 3: I don't think you need to apply this gating to the whole module.

newpavlov:
Adding support of mixed decoding should be as simple as removing the match and using:
```rust
ret += (((0x2fisize - byte) & (byte - 0x3a)) >> 8) & (byte - 47);
ret += (((0x60isize - byte) & (byte - 0x67)) >> 8) & (byte - 86);
ret += (((0x40isize - byte) & (byte - 0x47)) >> 8) & (byte - 54);
```

Also it may be worth to use `i16`/`u16` in the nibble functions. It looks like it helps compiler with code auto-vectorization (the current version does not auto-vectorize at all).

tarcieri:
Merging so you can do a second pass, @newpavlov.

One note though:

> Adding support of mixed decoding should be as simple as...

That's interesting, and @Sc00bz had similar functionality which isn't included here.

I wonder if it would make sense to try to constrain it to matched case though. I suppose any support for it would be nice.

PR #325, opened 2022-01-12
base16ct: re-organize crate

newpavlov:
Code re-organized into free-standing functions inside 3 modules: mixed, lower, and upper. The former has only decoding functions, while the latter two have both decoding and encoding functions. Also added `encode_str` functions which return `&str` instead of `&[u8]`. I wonder if it could be useful to have `decode_in_place` functions like in `base64ct`.

Unfortunately, even though decoding functions [get auto-vectorized](https://rust.godbolt.org/z/f9Ws54bzY) properly with default build parameters, we are not so lucky with encoding functions: https://rust.godbolt.org/z/TxM6v1W45 It could be worth to help compiler with it somehow.

### DIFF
--- a/base16ct/src/lib.rs
+++ b/base16ct/src/lib.rs
@@ -82,8 +82,6 @@ pub fn decoded_len(bytes: &[u8]) -> Result<usize, Error> {
 }
 
 /// Get the length of Base16 (hex) produced by encoding the given bytes.
-///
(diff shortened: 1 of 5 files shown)


tarcieri (approved):
I like it! Nice work

newpavlov:
@tarcieri 
BTW I forgot to include this crate into the README table, so it could be worth doing it in a future release PR.

PR #327, opened 2022-01-13
Update dependencies

newpavlov:
Including `Cargo.lock` update.

IIUC the updates do not affect public API.

### DIFF
--- a/pkcs5/src/pbes2/encryption.rs
+++ b/pkcs5/src/pbes2/encryption.rs
@@ -6,10 +6,7 @@ use block_modes::{block_padding::Pkcs7, BlockMode, Cbc};
 use hmac::{
     digest::{
         block_buffer::Eager,
-        core_api::{
-            BlockSizeUser, BufferKindUser, CoreProxy,
-            FixedOutputCore, UpdateCore,
-        },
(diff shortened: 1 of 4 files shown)


pkcs5/src/pbes2/encryption.rs
```diff
@@ -178,9 +186,17 @@ impl EncryptionKey {
...
+            D: CoreProxy,
+            D::Core: HashMarker
+                + UpdateCore
+                + FixedOutputCore
+                + BufferKindUser<BufferKind = Eager>
+                + Default
+                + Clone
+                + Sync,
+            <D::Core as BlockSizeUser>::BlockSize: IsLess<U256>,
+            Le<<D::Core as BlockSizeUser>::BlockSize, U256>: NonZero,
```
newpavlov, line 199: I know it's not pretty... But it's that we currently have. We could use [`SimpleHmac`](https://docs.rs/hmac/latest/hmac/struct.SimpleHmac.html), but it will be a bit less efficient.

newpavlov: Ideally we would have trait aliases, but AFAIK they are far from being implemented and stabilized. An alternative approach could be introduction of a helper trait in the `hmac` crate with blanket impl for types satisfying the currently used bound, but it would require somewhat annoying passing of wrapped methods and associated types.

PR #329, opened 2022-01-15
base16ct: add `HexDisplay` type

tarcieri:
Adds a `HexDisplay` wrapper type with impls of `Display`, `UpperHex`, and `LowerHex` from `core::fmt` which can be used to display a byte slice using a constant-time hex encoder.


tarcieri:
The `elliptic-curve` crate uses hex as the `Display` impl for several types. This is needed to replace that (and make it constant time-ish).

PR #346, opened 2022-01-19
 der: support for the REAL type

### LINKED ISSUE #304: der: support for the `REAL` type
Hi there,

I'm seriously considering using your library to encode lots of `f64` data (cf. https://github.com/anise-toolkit , and specifically this flatbuffer example: https://github.com/anise-toolkit/specs/blob/1-draft-0-0-1/ephemeris.fbs ).

I'm brand new to ASN.1, so please excuse novice questions. According to a copy of some specs I found here, https://www.oss.com/asn1/resources/books-whitepapers-pubs/larmouth-asn1-book.pdf, section 2.4 (page 83) talks about a `REAL` type. The ASN.1 playground also uses that type in their default example: https://asn1.io/asn1playground/ .

However, I don't see it in the docs as one of the supported type (https://docs.rs/der/latest/der/asn1/index.html). 

Is the REAL type not supported in DER encoding? Or has it not yet been implemented in this library because it isn't needed for the crypto algorithms?

Thanks

ChristopherRabotin:
Covers sections 8.5 and 11.3 of ITU-T X.690 (02/2021)

Caveats:
	1. Interpretation of section 11.3.1 to be further discussed
	2. Currently, all encoding is binary (no ISO 6093 NR3)

Testing of arbitrarily large and small f64s shows that DER typically
requires 12 bytes to encode these 8-byte IEEE-754 values. This can be
explained because of the extra two bytes for the tag and length, and the
fact that the exponent is encoded on 1 to 2 bytes (only 11 bits in
IEEE-754), and the mantissa may require up to 8 bytes itself, hence 12
bytes total. Of note that, depending on Caveat 1 above, this information
may be wrong.

Closes #304


der/src/asn1/real.rs
```diff
@@ -0,0 +1,629 @@
...
+        1 => 0b00000010,
+        2 => 0b00000100,
+        3 => 0b00001000,
+        4 => 0b00010000,
+        5 => 0b00100000,
+        6 => 0b01000000,
+        7 => 0b10000000,
+        _ => return false,
+    };
+    bytes.get(0).map(|byte| byte & mask != 0).unwrap_or(false)
```
tarcieri, line 180:
```suggestion
    debug_assert!(N < 8)
    bytes.get(0).map(|byte| byte & (1 << N) != 0).unwrap_or(false)
```

der/src/asn1/real.rs
```diff
@@ -0,0 +1,629 @@
...
+    let data = if bit_m && bit_n {
+        0b11
+    } else if !bit_m && !bit_n {
+        0b00
+    } else if bit_n {
+        0b01
+    } else {
+        0b10
+    };
+    u8::from_be_bytes([data])
```
tarcieri, line 196:
```suggestion
    (bit_m as u8) << 1 | bit_n as u8
```

tarcieri, line 196:
```suggestion
    (bit_m as u8) << 1 | bit_n as u8
```

tarcieri:
If you can find another implementation to do interop testing with, that'd be great

ChristopherRabotin:
Thanks for the review. I'll add a demo encoding from the ASN1 playground

der/src/error.rs
```diff
@@ -220,6 +220,13 @@ pub enum ErrorKind {
     #[cfg_attr(docsrs, doc(cfg(feature = "std")))]
     PermissionDenied,
 
+    /// Real related error: base is not DER compliant (base encoded in enum)
+    RealBaseInvalid(u8),
+    /// Real related error: encoded exponent cannot be represented on an IEEE-754 double
+    RealExponentTooLong,
+    /// Real related error: encoding not supported or malformed
+    RealISO6093Error,
+
```
tarcieri, line 229: Not sure if I mentioned this before, but this is *a lot* of `REAL`-related error variants to add to what is primarily intended to be a cryptography library.

Pretty much every other type propagates errors like this as `Self::TAG.value_error()`, which only includes information about the tag.

ChristopherRabotin:
As a quick update, I'm working on adding a bunch of validation test cases. It's gonna take some time as it seems like some of my encoding does not match the ASN1 playground so I'm going through the specs again to better understand what I didn't implement correctly. I'll keep the branch up-to-date with master.

ChristopherRabotin:
After a two months of working on other stuff, here's the updated PR with validation cases. Let me know if you have any questions.

I'm not sure why clippy is failing in the Github actions, it works locally with `clippy 0.1.59 (9d1b210 2022-02-23)`.

Cheers

ChristopherRabotin:
Not sure what [this CI build](https://github.com/RustCrypto/formats/runs/5669154521?check_suite_focus=true) failed as running `cargo test --no-default-features` works locally and both the `val` and `val2` variables are in fact defined.

tarcieri:
It's failing here:

https://github.com/RustCrypto/formats/runs/5669154521?check_suite_focus=true

```
error: there is no argument named `val`
   --> der/src/asn1/real.rs:827:30
    |
827 |                 "fail - want {val}\tgot {val2}"
    |                              ^^^^^

error: there is no argument named `val2`
   --> der/src/asn1/real.rs:827:[41](https://github.com/RustCrypto/formats/runs/5669154521?check_suite_focus=true#step:7:41)
    |
827 |                 "fail - want {val}\tgot {val2}"
    |
```

I think the issue is that this is using the new [captured identifiers in format strings feature which was introduced in Rust 1.58](https://blog.rust-lang.org/2022/01/13/Rust-1.58.0.html#captured-identifiers-in-format-strings), and the MSRV of the crate is 1.57, which is what the test is failing on.

ChristopherRabotin:
Thanks for the suggestion, that fixed all of the tests.

der/src/asn1/real.rs
```diff
@@ -0,0 +1,987 @@
...
+
+impl FixedTag for f64 {
+    const TAG: Tag = Tag::Real;
+}
+
+impl<'a> From<f64> for Any<'a> {
+    fn from(_: f64) -> Any<'a> {
+        Any::from_tag_and_value(Tag::Real, ByteSlice::default())
+    }
+}
```
tarcieri, line 187: This seems broken? Since `Any` borrows I'm not sure you can define this properly without using `unsafe`, in which case I'd suggest just deleting it.

tarcieri:
Thank you!

PR #347, opened 2022-01-20
Clean up certificate extension tests

npmccallum:
Clippy now complains about the use of a manual counter in the loops. These
tests can be more succinctly expressed using `.zip(EXTENSIONS)`.


tarcieri:
Thanks!

PR #349, opened 2022-01-21
Implement context-specific tags on #[derive(Sequence)]

### LINKED ISSUE #322: `der` crate fails to emit implicit context-specific tags on `SET`

bstrie:
This also adds a context-specific tag test to the code for #[derive(Choice)], and cleans up the tests in these files.

If you'd like I would be happy to add more tests to cover the interactions between the various options, but I admit that I'm not an ASN.1 expert so I would need some guidance as to how these options interact so that I could formulate the correct desugaring.

Closes #322.

### DIFF
--- a/der/derive/src/choice/variant.rs
+++ b/der/derive/src/choice/variant.rs
@@ -83,8 +83,9 @@ impl ChoiceVariant {
 #[cfg(test)]
 mod tests {
     use super::ChoiceVariant;
-    use crate::{Asn1Type, FieldAttrs, Tag};
+    use crate::{Asn1Type, FieldAttrs, Tag, TagNumber};
     use proc_macro2::Span;
+    use quote::quote;
     use syn::Ident;
 
     #[test]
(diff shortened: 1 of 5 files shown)


npmccallum:
@bstrie I think you mean context specific, not context sensitive.

tarcieri:
> I think you mean context specific, not context sensitive

Thanks to the Chomsky Hierarchy I have made this mistake several times myself 😅

der/derive/src/choice/variant.rs
```diff
@@ -97,25 +98,91 @@ mod tests {
...
         assert_eq!(
             variant.to_decode_tokens().to_string(),
-            ":: der :: Tag :: Utf8String => Ok (Self :: ExampleVariant (decoder . decode () ? . try_into () ?)) ,"
+            quote! {
+                ::der::Tag::Utf8String => Ok(Self::ExampleVariant(
+                    decoder.decode()?
+                    .try_into()?
+                )),
+            }
+            .to_string()
```
tarcieri, line 109: Oh wow this is the simple and obvious-in-hindsight solution. Nice

npmccallum:
@bstrie I think you need to fix the commit message wording too.

npmccallum:
@bstrie Also, one of the tests is failing.

tarcieri:
One of the errors is on `IssuingDistributionPointsExample`:

```rust
    impl ::der::Sequence<'_> for IssuingDistributionPointExample {
        fn fields<F, T>(&self, f: F) -> ::der::Result<T>
        where
            F: FnOnce(&[&dyn der::Encodable]) -> ::der::Result<T>,
        {
            ()
        }
    }
```

...not sure what's happening there but the empty tuple is definitely not the right result.

Looks like something similar on the other error:

```
    impl<'a> ::der::Sequence<'a> for OneAsymmetricKey<'a> {
        fn fields<F, T>(&self, f: F) -> ::der::Result<T>
        where
            F: FnOnce(&[&dyn der::Encodable]) -> ::der::Result<T>,
        {
            ()
        }
    }
```

Perhaps the unparseable tokens are being lost?

```
error: proc-macro derive produced unparseable tokens
```

der/derive/src/sequence/field.rs
```diff
@@ -188,13 +192,29 @@ impl LowerFieldEncoder {
...
+        let encoder = &self.encoder;
+        let number_tokens = tag_number.to_tokens();
+        let mode_tokens = tag_mode.to_tokens();
+
+        self.encoder = quote! {
+            &::der::asn1::ContextSpecific {
+                tag_number: #number_tokens,
+                tag_mode: #mode_tokens,
+                value: Clone::clone(#encoder),
+            },
```
tarcieri, line 207: I think the `,` here needs to be removed:

```suggestion
            }
```

The proc macro assembles these together using `quote`:

```rust
f(&[
    #(#encode_body),*
])
```

der/derive/src/sequence/field.rs
```diff
@@ -188,13 +192,29 @@ impl LowerFieldEncoder {
...
     }
+
+    /// Make this field context-specific.
+    fn apply_context_specific(&mut self, tag_number: &TagNumber, tag_mode: &TagMode) {
+        let encoder = &self.encoder;
+        let number_tokens = tag_number.to_tokens();
+        let mode_tokens = tag_mode.to_tokens();
+
+        self.encoder = quote! {
+            &::der::asn1::ContextSpecific {
```
tarcieri, line 203: The `ContextSpecificRef` type should make the `Clone` unnecessary:

https://docs.rs/der/latest/der/asn1/struct.ContextSpecificRef.html

bstrie:
Code updated to pass all the tests (sorry about not doing so the first time, I forgot to test with `--all-features`). The tests that were failing were already using context-specific tags (which I guess were just being ignored before now) in conjunction with various other field attributes. Taking into account all the supported combinations of the currently-implemented attributes was a bit tricky and required some subtle changes to make sure all the references lined up correctly, but in the end I think I managed to find a nice arrangement that doesn't involve adding state to the Lowerer or else compromising the unit-testability of the Lowerer methods. Once again I tested this with https://github.com/RustCrypto/formats/pull/323 to confirm that it addresses https://github.com/RustCrypto/formats/issues/322 .

der/tests/derive.rs
```diff
@@ -353,6 +353,28 @@ mod sequence {
     const ALGORITHM_IDENTIFIER_DER: &[u8] =
         &hex!("30 13 06 07 2a 86 48 ce 3d 02 01 06 08 2a 86 48 ce 3d 03 01 07");
 
+    #[derive(Sequence)]
+    #[asn1(tag_mode = "IMPLICIT")]
+    pub struct TypeCheckExpandedSequenceFieldAttributeCombinations<'a> {
```
tarcieri, line 358: Oh my 😅

tarcieri (approved):

der/tests/derive.rs
bstrie: We call that "self-documenting code". :P

PR #355, opened 2022-01-27
add some missing pub declarations

carl-wallace:


npmccallum (approved):
@carl-wallace Make sure to tag me in the new PRs related to the previous PR.

carl-wallace:
@npmccallum The previous PR has been split into several parts. I think I hit all of your points but I forgot an idiomatic change to a test case. Will do that later, done for now.

PR #357, opened 2022-01-28
Partial signed data impl

carl-wallace:
add just enough CMS support to allow incoming path builder implementation to parse certs-only SignedData messages. The inclusion of SetOf<SignerInfo> caused several structs to need Ord and/or PartialOrd added.


tarcieri:
A general note: the `DerOrd` and `ValueOrd` traits are needed for `SetOf` ordering now. The `Ord` trait does not necessarily reflect the ordering when serialized as DER.

`ValueOrd` can be derived.

spki/src/algorithm.rs
```diff
@@ -16,7 +16,7 @@ use der::{
 /// ```
 ///
 /// [RFC 5280 Section 4.1.1.2]: https://tools.ietf.org/html/rfc5280#section-4.1.1.2
-#[derive(Copy, Clone, Debug, Eq, PartialEq)]
+#[derive(Copy, Clone, Debug, Eq, PartialEq, PartialOrd, Ord)]
```
tarcieri, line 19: This type already impls `ValueOrd`

tarcieri (changes requested):

spki/src/algorithm.rs
carl-wallace: Dropped Ord and PartialOrd from AlgorithmIdentifier and SignerInfo. Stubbed out ValueOrd for SignerInfo (not needed now and CMS is well short of complete at present).

carl-wallace:
Close pending refactoring to account for recent changes.

PR #358, opened 2022-01-28
Documentation tweeks

carl-wallace:
Mostly things to make cargo doc happy. Added a couple more Ord/PartialOrd. Added a display impl for Time.


tarcieri:
Does this PR contain commits from another? It seems like a lot more than just documentation tweaks

carl-wallace:
Yes, there were two trailing bits of mods from the original larger PR that I noted in the comment after misnaming the branch. Those mods were a few additional Ord/PartialOrd (that likely need revisiting in light of the ValueOrd discussion) and a Display impl for Time.

PR #360, opened 2022-01-28
Add AES ECB PKCS7 support to PKCS5

fishface60:
The most ergonomic way to make use of the ciphers is with a `pbes2::Parameters` struct, and to interoperate with a program written to use Java's default "AES" cipher I need AES in ECB mode with PKCS7 padding.

The ECB cipher modes are feature-gated because they shouldn't be encouraged for new code.

### DIFF
--- a/pkcs5/Cargo.toml
+++ b/pkcs5/Cargo.toml
@@ -34,7 +34,8 @@ hex-literal = "0.3"
 [features]
 alloc = []
 3des = ["pbes2", "des"]
-des-insecure = ["pbes2", "des"]
+des-insecure = ["pbes2", "des", "ecb"]
+ecb = ["block-modes"]
 pbes2 = ["aes", "block-modes", "hmac", "pbkdf2", "scrypt", "sha2"]
 sha1 = ["pbes2", "sha-1"]
 
--- a/pkcs5/Cargo.toml
+++ b/pkcs5/Cargo.toml
@@ -34,8 +34,8 @@ hex-literal = "0.3"
 [features]
 alloc = []
 3des = ["pbes2", "des"]
-des-insecure = ["pbes2", "des", "ecb"]
-ecb = ["block-modes"]
+aes-ecb-insecure = ["block-modes"]
+des-insecure = ["pbes2", "des"]
 pbes2 = ["aes", "block-modes", "hmac", "pbkdf2", "scrypt", "sha2"]
 sha1 = ["pbes2", "sha-1"]
 
--- a/pkcs5/src/pbes2/encryption.rs
+++ b/pkcs5/src/pbes2/encryption.rs
@@ -56,13 +56,9 @@ pub fn encrypt_in_place<'b>(
                 .map_err(|_| Error::EncryptFailed)
         }
         #[cfg(feature = "aes-ecb-insecure")]
-        EncryptionScheme::Aes128Ecb => {
-            let cipher = Aes128Ecb::new_from_slices(encryption_key.as_slice(), Default::default())
-                .map_err(|_| es.to_alg_params_invalid())?;
-            cipher
-                .encrypt(buffer, pos)
-                .map_err(|_| Error::EncryptFailed)
-        }
+        EncryptionScheme::Aes128Ecb => Err(Error::UnsupportedAlgorithm {
+            oid: super::AES_128_ECB_OID,
+        }),
         EncryptionScheme::Aes192Cbc { iv } => {
             let cipher = Aes192Cbc::new_from_slices(encryption_key.as_slice(), iv)
                 .map_err(|_| es.to_alg_params_invalid())?;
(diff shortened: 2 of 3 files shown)


pkcs5/Cargo.toml
```diff
@@ -34,7 +34,8 @@ hex-literal = "0.3"
 [features]
 alloc = []
 3des = ["pbes2", "des"]
-des-insecure = ["pbes2", "des"]
+des-insecure = ["pbes2", "des", "ecb"]
+ecb = ["block-modes"]
```
tarcieri, line 38: I don't think it makes sense to conflate these. How about:

```suggestion
aes-ecb-insecure = ["block-modes"]
des-insecure = ["pbes2", "des"]
```

pkcs5/src/pbes2/encryption.rs
```diff
@@ -49,20 +55,44 @@ pub fn encrypt_in_place<'b>(
...
                 .map_err(|_| Error::EncryptFailed)
         }
+        #[cfg(feature = "ecb")]
+        EncryptionScheme::Aes256Ecb => {
+            let cipher = Aes256Ecb::new_from_slices(encryption_key.as_slice(), Default::default())
+                .map_err(|_| es.to_alg_params_invalid())?;
+            cipher
+                .encrypt(buffer, pos)
+                .map_err(|_| Error::EncryptFailed)
+        }
```
tarcieri, line 95: Like 56-bit DES, I don't think it makes sense to support encrypting new files with known insecure modes. Instead, it should be decrypt-only.

tarcieri:
Slightly reluctant to expand our repertoire of known insecure modes, but so long as it's decrypt-only I'm OK with it.

pkcs5/src/pbes2/encryption.rs
fishface60: I'd missed you were doing that, I thought the feature-gating was sufficient discouragement.

Unfortunately for me I need to re-encrypt a response.
For me that feels somewhere in-between creating a new insecurely encrypted file and decrypting legacy content, but I can't think of a technical way to prevent making new files while still allowing participating in encrypted connections.

The application I'm working with takes other measures to protect against ECB-related insecurity.
Would an additional feature gate for insecure re-encryption be acceptable? I'm not sure how effective pushing for an encryption change and only supporting newer version of the app would be.

tarcieri: I’d really prefer not to add encryption support for any known insecure modes

fishface60: Alright, I've taken encryption support out, up to you if you whether to merge or close since it won't have a known user.

tarcieri:
If you don't have a use for this, then my preference is to close it as being both an insecure mode and an obscure one.

Alexx-G:
I have a similar use case to OP's - I need to exchange some inter-system non-critical encrypted data using `AES/ECB/PKCS5Paddin` (it's part of a protocol that cannot be changed). At some point the Rust component used `aes` (v0.7) and `block_modes` (deprecated) crates to solve that.
I've done several attempts to upgrade to new libraries and so far I don't have a clear vision on how I can fix Rust implementation so it keeps working with other services.

@tarcieri While I totally understand all the implications of using an insecure mode, I fail to see what's the impact in case its complete (encrypt/decrypt) support is implemented behind a non-default feature flag. I mean, quite often Rust is integrated with existing services, there are plenty of existing APIs that either cannot be changed (for various reasons) and they may need to use ECB. And since such platforms as JVM/NodeJS/Python offer support for ECB mode, Rust would only benefit from supporting it as well.

tarcieri:
@Alexx-G okay, well it seems this mode is less obscure than I thought.

Can you speak more to your use case that only supports AES-ECB mode? There are no other options whatsoever?

I'm reluctant to accumulate junk crypto, especially in a mode that can help proliferate it instead of migrate away, but if there's actually legitimate use cases it's something we can consider.

Re: `block-modes`, you don't need any of the block mode crates to support ECB mode. You can use the raw `BlockEncrypt`/`BlockDecrypt` API.

Alexx-G:
@tarcieri thanks! Yes, I managed to solve my problem by just using `aes` crate directly with Pkcs7 padding (which should be used interchangeable with Pkcs5). So far it seems to work just fine. So, technically, in case there's an example/documentation, it's possible to have a workaround. On the other hand, I had way less friction when implementing the same SDK for JVM/Python/NodeJS/etc.

In my particular case, and I think it's accurate for most folks that have to use AES-ECB, it's part of an existing API that cannot change its implementation, there are no associated security risks (data isn't critical, it rather needs to be obfuscated than securely encrypted, and it's on a latency-critical path) and ROI is just way too small.

tarcieri:
> On the other hand, I had way less friction when implementing the same SDK for JVM/Python/NodeJS/etc.

The "friction" is we don't want to support junk cryptography unless absolutely less necessary. The other environments you reference are both old and have a more cavalier kitchen sink approach to what cryptography they choose to support.

> In my particular case, and I think it's accurate for most folks that have to use AES-ECB, it's part of an existing API that cannot change its implementation

What does this mean? An API provided by one or more language environments you referenced earlier? Or something you developed internally?

PKCS#5 is parameterized by `AlgorithmIdentifier`s which should allow for a simple migration for encrypting new messages while still supporting ECB for old ones. So unless this API in question is particularly badly designed, or AES-ECB is literally the best option available for one or more of these languages environments, I don't see why you can't upgrade to something better.

PR #356, opened 2022-01-27
add support for RFC5914 structs. 

carl-wallace:
make adjusts to some prior structure definitions to work around macro-related issues. Notes on non-use of macros

In trust_anchor_format.rs
- Choice was not used on TrustAnchorChoice owing to types not available in Asn1Type enum
- Sequence was not used on TrustAnchorInfo owing to lack of DecodeValue required by context_specific

error[E0277]: the trait bound `TrustAnchorInfo<'a>: DecodeValue<'_>` is not satisfied
   --> x509/src/trust_anchor_format.rs:295:22
    |
295 |                     .context_specific::<TrustAnchorInfo<'a>>(TAC_TA_INFO_TAG, TagMode::Explicit)?;
    |                      ^^^^^^^^^^^^^^^^ the trait `DecodeValue<'_>` is not implemented for `TrustAnchorInfo<'a>`
    |
note: required by a bound in `Decoder::<'a>::context_specific`
   --> der/src/decoder.rs:172:12
    |
172 |         T: DecodeValue<'a> + FixedTag,
    |            ^^^^^^^^^^^^^^^ required by this bound in `Decoder::<'a>::context_specific`

- Sequence was not used on CertPathControls to use decode_implicit for each field (trailing bits were left using the default decoder even with mods to Certificate and NameConstraints)

In certificate.rs
- Sequence was not used on Certificate due to same issue as described above for TrustAnchorInfo (i.e., tension between use of DecodeValue and Decodable)
- Sequence was not used on NameConstraints for the same DecodeValue vs Decodable issue.

PR #363, opened 2022-01-29
Derive `Sequence` for more types

npmccallum:
This reduces the amount of custom code that is written.

### DIFF
--- a/pkcs1/src/private_key/other_prime_info.rs
+++ b/pkcs1/src/private_key/other_prime_info.rs
@@ -1,6 +1,6 @@
 //! PKCS#1 OtherPrimeInfo support.
 
-use der::{asn1::UIntBytes, Decodable, Decoder, Encodable, Sequence};
+use der::{asn1::UIntBytes, Sequence};
 
 /// PKCS#1 OtherPrimeInfo as defined in [RFC 8017 Appendix 1.2].
 ///
@@ -15,7 +15,7 @@ use der::{asn1::UIntBytes, Decodable, Decoder, Encodable, Sequence};
 /// ```
 ///
 /// [RFC 8017 Appendix 1.2]: https://datatracker.ietf.org/doc/html/rfc8017#appendix-A.1.2
-#[derive(Clone)]
+#[derive(Clone, Sequence)]
 #[cfg_attr(docsrs, doc(cfg(feature = "alloc")))]
 pub struct OtherPrimeInfo<'a> {
     /// Prime factor `r_i` of `n`, where `i` >= 3.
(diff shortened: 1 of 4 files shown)


tarcieri:
The `spki` crate is a deep, hard dependency of a number of different crates which don’t currently have any proc macro dependencies.

In cases like that, pulling in custom derive has a somewhat cost in terms of compile times of all of the proc macro dependencies, e.g. `syn`.

I’d prefer to keep `spki` proc macro-free at least, and probably `pkcs1` as well (`pkcs8` is another I’d prefer not to change).

`pkcs5` is a little more debatable, but also effectively “done” already.

The others are fine.

carl-wallace:
Did I miss a change to derive? I am wondering what enables proper function with these manual mods gone.

npmccallum:
@carl-wallace How does a developer run the aforementioned tests? And why aren't they in the CI?

carl-wallace:
The bulk of the tests are in the certval library (and corresponding pittv3 tool) that let you build and validate certs live, enabling pretty thorough exercise of the encoders and decoders. The PR I submitted earlier this week (then chunked up a day or two ago) needs to clear for those first before certval and pittv3 are pushed. The quick test I did just now was to pull over these changes to the version used by certval/pittv3.

npmccallum:
@carl-wallace Could you add those tests to this repo so that we can find breakage early?

carl-wallace:
No, as the encoder/decoder that generates this is in the PR in front of it (and the bulk of the testing is driven by building and validating paths). Why is it vital to delete the manual impls now before those pieces land? That seems like the easiest path.

npmccallum:
@carl-wallace I don't know which PR you're referring to. Could you provide a link?

carl-wallace:
There was a large PR that you reviewed a few days ago that you suggested be chunked up. I chunked it up (but then pulled one piece back until these clear). One chunk has been processed. Two are currently outstanding: 356 and 357. Once those are done, a minor housekeeping PR will be pushed (to make cargo doc happier) then repos for certval/pittv3 can be set up (there's a yubikey mod that will follow those as well).

npmccallum:
@carl-wallace Please note that if you add a hash (`#`) before the PR numbers, you get links directly to them. This aids readers of your comments. For example, the issues you are referring to are #356 and #357.

carl-wallace:
FYI, there are some tests in #356 that should highlight breakage if you introduce your changes there. See the tests for trust_anchor_format. I did not try that specifically but what I did earlier is a close cousin to that action. Note, I think we will hit similar edges as CMS support emerges (I did not go enumerate where I think the issues will be but I think there will be similar cases).

npmccallum:
@carl-wallace What I'm proposing is that if we can discover and fix these issues in the `derive` code, it makes all this later development easier. That means that each ASN.1 type that is added should have a test which governs it (either explicitly or implicitly). Otherwise, we have no way to know when things are broken or not.

carl-wallace:
Is there macro change to enable the tests in #356 to pass? If not, please hold off on removing manual impls from x509 until the macros support tagging those structures.

Issue #64, opened 2021-09-30
tls_codec: `no_std` support


franziskuskiefer:
Allow building tls-codec with `no_std`.

> I now see this is a bit more `std`-dependent than I realized due to the extensive use of `std::io::{Read, Write}`.

_Originally posted by @tarcieri in https://github.com/RustCrypto/formats/pull/56#pullrequestreview-764378515_

tarcieri:
I think we can close this as having initial `no_std` support in place

PR #364, opened 2022-01-29
Run tests using `cargo hack`

npmccallum:
This dynamically ensures that the various combinations of features flags
successfully work.

### DIFF
--- a/.github/workflows/base16ct.yml
+++ b/.github/workflows/base16ct.yml
@@ -35,8 +35,8 @@ jobs:
           toolchain: ${{ matrix.rust }}
           target: ${{ matrix.target }}
           override: true
-      - run: cargo build --target ${{ matrix.target }} --release
-      - run: cargo build --target ${{ matrix.target }} --release --features alloc
+      - run: cargo install cargo-hack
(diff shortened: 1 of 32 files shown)


npmccallum:
Please note that this PR is intended to reveal existing breakage.

franziskuskiefer:
Thanks @npmccallum 

The `no_std` support for `tls_codec` isn't really there yet (see #64). We can either disable default features on `serde` for `no_std` or not check for the `serde_serialize` feature on `no_std` platforms.

```
info: running `cargo build --release --target thumbv7em-none-eabi --no-default-features --features serde_serialize` on tls_codec (2/2)
   Compiling serde_derive v1.0.133
   Compiling serde v1.0.133
error[E0463]: can't find crate for `std`
  |
  = note: the `thumbv7em-none-eabi` target may not support the standard library
  = note: `std` is required by `serde` because it does not declare `#![no_std]`
```

npmccallum:
@franziskuskiefer I took a slightly different approach. Let me know if this doesn't work.

tarcieri (approved):

franziskuskiefer (approved):
> @franziskuskiefer I took a slightly different approach. Let me know if this doesn't work.

That works as well. Thanks!

PR #377, opened 2022-02-01
Add an `x501` crate

npmccallum:
This crate consolidates the attribute and name types from the `x509` and
`pkcs10` crates. The `pkcs10` crate now depends on `x501` rather than `x509`.
It should now be simpler to release a version of `x501` and `pkcs10` without
requiring a full release of `x509`.

Note that unlike `spki`, x501 requires both `derive` and `alloc`.


tarcieri:
> Note that unlike `spki`, x501 requires both `derive` and `alloc`.

Yeah that's cool, it's how we're treating `x509` as well.

carl-wallace:
I'll mod x509 to reference that soonish and drop the corresponding types from there. OCSP can be born referencing this.

npmccallum:
@carl-wallace x509 is already included in this commit.

tarcieri:
Another fun application would be adding an `attributes = ["x501"]` feature to `pkcs8`, which presently ignores attributes.

If nothing else, it would make it easier to round trip RFC test vectors that include attributes.

npmccallum:
I forgot to fix up the GitHub Actions workflows. Adding those now.

carl-wallace:
OK. I'll update x509 and OCSP will reference this.

npmccallum:
Ok. All the actions are updated and all the tests pass. I think this is ready for review.

x501/src/name.rs
```diff
@@ -0,0 +1,58 @@
...
+pub type Name<'a> = RDNSequence<'a>;
+
+/// RDNSequence as defined in [RFC 5280 Section 4.1.2.4].
+///
+/// ```text
+/// RDNSequence ::= SEQUENCE OF RelativeDistinguishedName
+/// ```
+///
+/// [RFC 5280 Section 4.1.2.4]: https://datatracker.ietf.org/doc/html/rfc5280#section-4.1.2.4
+pub type RDNSequence<'a> = alloc::vec::Vec<RelativeDistinguishedName<'a>>;
```
tarcieri, line 23: Idiomatic Rust naming has generally moved to "capitalized acronyms", which is to say the first letter of an acronym is capitalized. See https://rust-lang.github.io/api-guidelines/naming.html

> In `UpperCamelCase`, acronyms and contractions of compound words count as one word: use `Uuid` rather than `UUID`, `Usize` rather than `USize` or `Stdin` rather than `StdIn`. 

```suggestion
pub type RdnSequence<'a> = alloc::vec::Vec<RelativeDistinguishedName<'a>>;
```

There are clippy lints for this on structs, but apparently not on type aliases.

tarcieri (changes requested):
One nit re: `RDNSequence` naming but otherwise looks good.

x501/Cargo.toml
```diff
@@ -0,0 +1,12 @@
...
+version = "0.1.0" # Also update html_root_url in lib.rs when bumping this
+edition = "2021"
+
+# See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
+
+[dev-dependencies]
+hex-literal = "0.3"
+
+[dependencies]
+der = { version = "=0.6.0-pre.0", features = ["derive", "alloc", "oid"], path = "../der" }
```
tarcieri, line 12: You might include a slightly more complete `Cargo.toml` [looking off one of the other crates as a reference](https://github.com/RustCrypto/formats/blob/master/x509/Cargo.toml).

We generally include license information, links to the repository, keywords, categories, specify the `README.md` name, etc.

x501/src/name.rs
npmccallum: It was all caps in the previous version. So I just copied that. Fixed now.

x501/Cargo.toml
npmccallum: Oops! I meant to do that before pushing but obviously forgot. Fixed now.

tarcieri (approved):

PR #388, opened 2022-02-05
Move Time/Validity to x501

npmccallum:
These types were originally defined in X.501. They are also used more
broadly than just in X.509.


x501/src/time.rs
```diff
@@ -1,22 +1,21 @@
-//! Validity [`Time`] as defined in RFC 5280
+//! X.501 time types as defined in RFC 5280
```
tarcieri, line 1: Within X.501 itself, `Time` and `Validity` appear to be defined in ISO/IEC 9594-2:2005 § 28.2: Establish Operational Binding operation

npmccallum: That is true. However, I think this is also implied by the crate name (`x501`). Further, we don't actually follow the definition in `ISO/IEC 9594-2:2005` but rather the more modern definitions in RFC 5280.

npmccallum: @tarcieri I added a conformity paragraph to the crate documentation that explains the situation.

PR #389, opened 2022-02-05
[WIP] Add x509-ext crate

npmccallum:
Don't merge this yet. It is a work in progress.


x509-ext/src/lib.rs
```diff
@@ -0,0 +1,62 @@
...
+
+extern crate alloc;
+
+pub mod pkix;
+
+/// A trait expressing the OID associated with an extension.
+pub trait Typed {
+    /// The OID which identifies an extension.
+    const OID: ObjectIdentifier;
+}
```
tarcieri, line 22: This might be interesting to add to `const-oid`. There are a few other places it could be used.

npmccallum: Done.

const-oid/src/lib.rs
```diff
@@ -241,3 +241,9 @@ impl fmt::Display for ObjectIdentifier {
         Ok(())
     }
 }
+
+/// A trait expressing the association of a type with an OID.
+pub trait Typed {
```
tarcieri, line 246: This could probably use a more descriptive name, but I'm having trouble thinking of a good one

tarcieri: How about `OidValue`?

npmccallum:
This has been (almost) superseded by #414. Closing.

PR #392, opened 2022-02-07
Pass `Header` to `DecodeValue`

npmccallum:
This makes it possible to implement `DecodeValue` for tagless types such
as `CHOICE` and `ANY`.


tarcieri (approved):
This certainly does help address a few use cases.

I'm a bit worried in that it has a sharp edge in regard to context-specific `IMPLICIT` productions: anything which attempts to introspect on the tag will encounter the context-specific tag in that case in lieu of e.g. a universal tag they might be expecting.

It might be good to document that in the rustdoc for `DecodeValue`.

PR #393, opened 2022-02-07
x501+x509: `CertificateDocument`, `X509Version`, and more additions

carl-wallace:
Modifications resulting from code review of changes to yubikey library and client tool to use x509 crate:
- moved additional OID definitions from certval into x509
- moved, refactored and expanded OID lookup function
- added CertificateDocument implementation
- added X509Version enum and improved decode and encode for TBSCertificate relative to version-specific fields


x509/src/certificate.rs
```diff
@@ -25,6 +27,57 @@ pub fn default_zero() -> u32 {
...
+
+impl Decodable<'_> for X509Version {
+    fn decode(decoder: &mut Decoder<'_>) -> der::Result<Self> {
+        X509Version::try_from(u8::decode(decoder)?).map_err(|_| Self::TAG.value_error())
+    }
+}
+
+impl Encodable for X509Version {
+    fn encoded_len(&self) -> der::Result<der::Length> {
+        der::Length::from(1u8).for_tlv()
```
tarcieri, line 51:
```suggestion
        Length::ONE.for_tlv()
```

x501/src/time.rs
```diff
@@ -40,6 +41,14 @@ impl Time {
         }
     }
 
+    /// Get Time as DateTime
```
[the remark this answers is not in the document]
carl-wallace: Changed

x509/src/certificate.rs
carl-wallace: Changed

x509/src/certificate.rs
```diff
@@ -25,6 +28,57 @@ pub fn default_zero() -> u32 {
 /// Version  ::=  INTEGER  {  v1(0), v2(1), v3(2)  }
 pub const X509_CERT_VERSION: u8 = 2;
 
+/// Version identifier for X.509 certificates. In practice, only v3 is used.
+#[derive(Clone, Debug, Copy, PartialEq, Eq, PartialOrd)]
+pub enum X509Version {
```
tarcieri, line 33: In the spirit of [RFC356](https://github.com/rust-lang/rfcs/blob/master/text/0356-no-module-prefixes.md) I'd suggest simply calling this `Version`.

That also follows the convention used by other version types in this repo.

x509/src/certificate_traits.rs
```diff
@@ -0,0 +1,97 @@
...
+use der::Document;
+
+#[cfg(feature = "std")]
+use std::path::Path;
+
+#[cfg(feature = "pem")]
+use {alloc::string::String, der::pem::LineEnding};
+
+/// Parse a public key object from an encoded Certificate document.
+pub trait DecodeCertificate: for<'a> TryFrom<Certificate<'a>, Error = Error> + Sized {
```
tarcieri, line 19: I realize this is following the convention from similar traits in the private key types crates, but I'm not sure this really confers a similar benefit.

Traits like `DecodePrivateKey` and `EncodePrivateKey` exist because there are many potential downstream types which can load and store key data. I'm not sure I see a similar utility here.

x509/src/certificate.rs
```diff
@@ -25,6 +28,57 @@ pub fn default_zero() -> u32 {
 /// Version  ::=  INTEGER  {  v1(0), v2(1), v3(2)  }
 pub const X509_CERT_VERSION: u8 = 2;
```
tarcieri, line 29: This seems like it could either just be `(X509)Version::default()` or an inherent constant

x509/src/certificate.rs
carl-wallace: ok. will change as i mop up cargo hack findings.

x509/src/certificate_traits.rs
carl-wallace: i agree and only added that for sake of consistency with other. Will drop.

x509/src/certificate.rs
carl-wallace: I meant to remove that when i added the enum. Do you think we need to retain it as a default? if not, i will just delete.

tarcieri: I think adding a `Default` impl would be a good idea

carl-wallace: Added

x509/src/certificate.rs
```diff
@@ -21,9 +24,65 @@ pub fn default_zero() -> u32 {
...
+
+impl Encodable for Version {
+    fn encoded_len(&self) -> der::Result<der::Length> {
+        Length::ONE.for_tlv()
+    }
+
+    fn encode(&self, encoder: &mut Encoder<'_>) -> der::Result<()> {
+        u8::from(*self).encode(encoder)
+    }
+}
```
tarcieri, line 63: Why not `DecodeValue` and `EncodeValue` here?

x509/src/certificate.rs
```diff
@@ -80,25 +139,37 @@ pub struct TBSCertificate<'a> {
 impl<'a> ::der::Decodable<'a> for TBSCertificate<'a> {
     fn decode(decoder: &mut ::der::Decoder<'a>) -> ::der::Result<Self> {
         decoder.sequence(|decoder| {
-            let version =
+            let version_int =
                 ::der::asn1::ContextSpecific::decode_explicit(decoder, ::der::TagNumber::N0)?
                     .map(|cs| cs.value)
                     .unwrap_or_else(default_zero_u8);
+            let version = Version::try_from(version_int)?;
```
tarcieri, line 146: If you change `Version` to impl `DecodeValue` you should be able to do:

```suggestion
            let version = decoder.context_specific::<Version>(der::TagNumber::N0, der::TagMode::Explicit)?;
```

x509/src/certificate.rs
```diff
@@ -21,9 +24,65 @@ pub fn default_zero() -> u32 {
     0
 }
 
-/// only support v3 certificates
+/// only supporting v3 certificates, but all versions are listed here
 /// Version  ::=  INTEGER  {  v1(0), v2(1), v3(2)  }
-pub const X509_CERT_VERSION: u8 = 2;
+
+/// Version identifier for X.509 certificates. In practice, only v3 is used.
+#[derive(Clone, Debug, Copy, PartialEq, Eq, PartialOrd)]
```
tarcieri, line 31:
```suggestion
#[derive(Copy, Clone, Debug, Copy, PartialEq, Eq, PartialOrd)]
```

x509/src/certificate.rs
```diff
@@ -121,27 +192,45 @@ impl<'a> ::der::Sequence<'a> for TBSCertificate<'a> {
...
+        };
+
         #[allow(unused_imports)]
         use core::convert::TryFrom;
         f(&[
             &ContextSpecific {
                 tag_number: VERSION_TAG,
                 tag_mode: TagMode::Explicit,
-                value: self.version,
+                value: version_value,
```
tarcieri, line 223: Seems like just `self.version` should work here, especially if you make it `Copy`

x509/src/certificate.rs
carl-wallace: That won't work straight up because it yields an Option<> (and the version guards on the v2/v3 bits were on the int).

x509/src/certificate.rs
carl-wallace: Copy was there. Reordered so it is now first.

x509/src/certificate.rs
carl-wallace: No particular reason, habit I suppose.

x509/src/certificate.rs
carl-wallace: I tried that unsuccessfully, will pick this back up tomorrow.

x509/src/certificate.rs
carl-wallace: Actually, Decodable/Encodable were taken from PKCS8 Version

x509/src/certificate.rs
carl-wallace: Addressed here (initially I though you meant replace that production with self.version but landed as shown): https://github.com/carl-wallace/formats/commit/c226d37a77913bad7a0a674f679b7b85a4ecb280

x509/src/certificate.rs
carl-wallace: Addressed this as well as default version handling in general in light of change from u8 to version: https://github.com/carl-wallace/formats/commit/b600b9982e63c0a39a13edd8bf0f29bd0969ba83

x509/src/certificate_traits.rs
carl-wallace: Actually, on second look, this seems as good a place as any to hang PEM support. Have left in place for now.

tarcieri: I think you can use inherent methods for that. The trait is only impl'd on one type, right?

x509/src/certificate.rs
tarcieri: Aah yeah, that's outdated, and also PKCS#8 doesn't need to deal with a context-specific version field so it works there

x509/src/certificate.rs
```diff
@@ -1,14 +1,17 @@
...
+    TagMode, TagNumber,
+};
 use spki::{AlgorithmIdentifier, SubjectPublicKeyInfo};
 use x501::name::Name;
 use x501::time::Validity;
 
 /// returns false in support of integer DEFAULT fields set to 0
-pub fn default_zero_u8() -> u8 {
-    0
+pub fn default_v1() -> Version {
```
tarcieri, line 13: This doesn't need to be `pub`

x509/src/certificate_traits.rs
carl-wallace: D'oh. Of course. Done.

x509/src/certificate_document.rs
```diff
@@ -0,0 +1,166 @@
...
+}
+
+impl<'a> Document<'a> for CertificateDocument {
+    type Message = Certificate<'a>;
+    const SENSITIVE: bool = false;
+}
+
+impl CertificateDocument {
+    /// Parse [`Certificate`] from buffer containing DER-encoded certificate
+    pub fn from_certificate_der(bytes: &[u8]) -> Result<Self> {
```
tarcieri, line 44: Since these aren't trait methods and they're defined on `CertificateDocument` directly, I think having `*_certificate_*` in the name is redundant

carl-wallace: Actually they can just go away. Tests pass when directly leaning on blanket impl.

tarcieri: Aah yeah, indeed!

x509/src/certificate.rs
npmccallum: #410 allows you to derive this type.

npmccallum:
@tarcieri I would like to review this after my derive fixes and other cleanup has been merged.

x509/src/certificate_document.rs
```diff
@@ -0,0 +1,97 @@
...
+use {core::str::FromStr, der::pem};
+
+/// Certificate document.
+///
+/// This type provides storage for [`Certificate`] encoded as ASN.1
+/// DER with the invariant that the contained-document is "well-formed", i.e.
+/// it will parse successfully according to this crate's parsing rules.
+#[derive(Clone)]
+#[cfg_attr(docsrs, doc(cfg(feature = "alloc")))]
+pub struct CertificateDocument(Vec<u8>);
```
npmccallum, line 20: @tarcieri I know I've implemented one of these document types for `pkcs10`. However:

1. What is the purpose of them? I don't understand how they are intended to be used. It seems that PEM decoding is implemented through the document types. However, in order to instantiate a document, you have to fully parse the DER. So why not just have the fully parsed type?

2. If there is a good reason for the document types, surely these can be genericized. They are all exactly the same code and differ only by the type of the value decoded therein. So `Document<T: Decodable + Encodable>` seems like an obvious win here.

x509/src/certificate.rs
```diff
@@ -21,9 +24,65 @@ pub fn default_zero() -> u32 {
...
+    type Error = Error;
+    fn try_from(byte: u8) -> Result<Version, Error> {
+        match byte {
+            0 => Ok(Version::V1),
+            1 => Ok(Version::V2),
+            2 => Ok(Version::V3),
+            _ => Err(Self::TAG.value_error()),
+        }
+    }
+}
```
npmccallum, line 85: @carl-wallace All the changes in this file are obsolete if we merge #410 instead.

x509/src/pkix_oids.rs
```diff
@@ -188,3 +189,278 @@ pub const PKIX_OCSP_NOCHECK: ObjectIdentifier = ObjectIdentifier::new("1.3.6.1.5
...
+
+/// id-sha384  OBJECT IDENTIFIER  ::=
+///        { joint-iso-itu-t(2) country(16) us(840) organization(1) gov(101)
+///        csor(3) algorithms(4) hashalgs(2) 2 }
+pub const PKIXALG_SHA384: ObjectIdentifier = ObjectIdentifier::new("2.16.840.1.101.3.4.2.2");
+
+/// id-sha512  OBJECT IDENTIFIER  ::=
+///        { joint-iso-itu-t(2) country(16) us(840) organization(1) gov(101)
+///        csor(3) algorithms(4) hashalgs(2) 3 }
+pub const PKIXALG_SHA512: ObjectIdentifier = ObjectIdentifier::new("2.16.840.1.101.3.4.2.3");
```
npmccallum, line 353: I really don't like the "big pile of OID constants" approach. There has to be a way to organize these better.

x509/src/pkix_oids.rs
```diff
@@ -188,3 +189,278 @@ pub const PKIX_OCSP_NOCHECK: ObjectIdentifier = ObjectIdentifier::new("1.3.6.1.5
...
+        PKIXALG_PSPECIFIED => "pSpecified",
+        PKIXALG_MGF1 => "MFG1",
+        PKIXALG_RSASSA_PSS => "sha1",
+        PKIXALG_SHA224 => "sha224",
+        PKIXALG_SHA256 => "sha256",
+        PKIXALG_SHA384 => "sha384",
+        PKIXALG_SHA512 => "sha512",
+        _ => "",
+    }
+}
```
npmccallum, line 466: I really don't like this function. It smells like OpenSSL. What is the purpose of it?

x509/tests/certificate.rs
```diff
@@ -67,7 +67,9 @@ pub struct DeferDecodeTBSCertificate<'a> {
     /// Defer decoded field
     pub extensions: &'a [u8],
 }
-
+pub fn default_zero_u8() -> u8 {
+    0
+}
```
npmccallum, line 72: Please don't. Use `Default::default`.

npmccallum (changes requested):

x509/src/certificate.rs
carl-wallace: Sure. Further review of this until after alignment with changes to dependencies is likely not worthwhile.

x509/src/pkix_oids.rs
carl-wallace: Logging and display.

x509/src/pkix_oids.rs
carl-wallace: The OIDs are grouped by module with names derived by capitalization, insertion of underscores and addition of a prefix. I'm OK with this approach, but if there is something desired please let me know.

x509/tests/certificate.rs
carl-wallace: As part of the realignment as noted above.

x509/src/certificate_document.rs
tarcieri: This seems like a complicated and important enough issue I did a long-form write up on it, so as not to lose the answers to these questions in a PR comment: #421

Some short answers though:

1. The purpose of these types is discussed at length in #421, but in regard to PEM specifically, since `Decode<'a>` is zero-copy and needs to borrow from a slice containing DER, parsing PEM needs to decode DER into a buffer that `Decode<'a>` can borrow from. If we had something like `DecodeOwned` which didn't need to borrow, documents could be decoded directly from PEM. As it were, this is how the `ssh-key` crate works, and I just added support for incremental/buffered PEM decoding to the `pem-rfc7468` crate (#406) and changed the `ssh-key` crate to use that (#407)

2. The problem with `Document<T: Decodable + Encodable>` is you omitted the lifetime on `Decodable<'a>`. Since it's a `'self` borrow and [Rust doesn't actually have a way to notate a `'self` lifetime](https://users.rust-lang.org/t/access-to-implicit-lifetime-of-containing-object-aka-self-lifetime/18917/17) you can't notate that on a struct definition, however as I noted in #421 it can probably be done by moving the bounds to individual methods so `Decodable<'a>` can hang off `pub fn decode<'a>(&'a self) -> T where T: Decode<'a>`

carl-wallace:
outdated

PR #394, opened 2022-02-07
Choose a better field name for `Attribute`

npmccallum:


richsalz:
thanks for this.

PR #395, opened 2022-02-07
Derive `EncodeValue` for `Choice`

npmccallum:
This enables `Choice` values to be used as a field in a `Sequence`.


tarcieri (approved):
Will go ahead and merge this.

It'd be nice to have test of the context-specific lowerings for `Choice`, but those can be added in a followup.

PR #397, opened 2022-02-08
Only emit `.try_into()?` when a type is specified

npmccallum:
The purpose of this transformation is for when the caller specifies `type =
"..."`. In that case, we decode into the specified type and then convert into
actual type. However, in cases where no type is specified, we have to decode
into an unspecified type `T` and then convert into the final type; but the
compiler (obviously) cannot determine the type `T`.

Therefore, when no `type` is specified, we drop the `.try_into()?` and just
attempt to decode into the final type directly.

### DIFF
--- a/der/derive/src/choice/variant.rs
+++ b/der/derive/src/choice/variant.rs
@@ -47,8 +47,10 @@ impl ChoiceVariant {
         let tag = self.tag.to_tokens();
         let ident = &self.ident;
         let decoder = self.attrs.decoder();
-        quote! {
-            #tag => Ok(Self::#ident(#decoder.try_into()?)),
+
+        match self.attrs.asn1_type {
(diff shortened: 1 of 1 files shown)


der/derive/src/choice/variant.rs
```diff
@@ -153,7 +154,7 @@ mod tests {
                 ::der::Tag::ContextSpecific {
                     constructed: false,
                     number: ::der::TagNumber::N0,
-                } => Ok(Self::ImplicitVariant(decoder.decode()?.try_into()?)),
+                } => Ok(Self::ImplicitVariant(decoder.decode()?)),
```
tarcieri, line 157: Can you add a test for a case where `.try_into()` should be emitted?

npmccallum: Done.

tarcieri (approved):

PR #398, opened 2022-02-08
Add `#[asn1(constructed = "true")]`

npmccallum:
This allows for deriving `Choice` over constructed types.


der/derive/src/attributes.rs
```diff
@@ -174,7 +186,7 @@ impl FieldAttrs {
                 .map(|tag_number| {
                     Some(Tag::ContextSpecific {
                         // TODO(tarcieri): handle constructed inner types
```
tarcieri, line 188: You can remove the TODO now

npmccallum: Done.

PR #400, opened 2022-02-08
Fix length calculation for explicit tags

npmccallum:
When deriving `Choice`, the existing macros emit incorrect length
calculation code when explicit tags are used.

### DIFF
--- a/der/derive/src/choice/variant.rs
+++ b/der/derive/src/choice/variant.rs
@@ -67,8 +67,22 @@ impl ChoiceVariant {
     /// Derive a match arm for the impl body for `der::EncodeValue::value_len`.
     pub(super) fn to_value_len_tokens(&self) -> TokenStream {
         let ident = &self.ident;
-        quote! {
-            Self::#ident(variant) => variant.value_len(),
+
+        match self.attrs.context_specific {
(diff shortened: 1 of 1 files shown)


tarcieri:
Nice! Can you add a test for the code generated for an `EXPLICIT` context-specific field?

npmccallum:
@tarcieri Is it sufficient to convert `GeneralName` to a derive with all tests passing? Because that's the real test case driving these fixes.

tarcieri:
@npmccallum that's probably fine for now. I can circle back and add some tests

carl-wallace:
I can run in builder/validator against PKITS, FPKI and Web PKI certs when there's a chance to update. That hits a fair number of extensions that ought exercise the moving parts.

tarcieri:
I'm really more worried about custom derive unit tests which test the output against some expected generated code.

The reason why they're important is debugging lowering failures outside of unit tests is a huge headache.

carl-wallace:
Sure, but catching errors from a broad set of data can inform adding unit tests. Many of the tests there are from cases that failed and highlighted an issue. But given much of what is moving is addressing compilation errors, it probably matters less. Will hold off.

PR #410, opened 2022-02-10
Derived Version Structs

npmccallum:
1. add a derive macro for deriving enumerations that are tagged as `INTEGER`.
2. use the new macro for `pkcs10::Version`.
3. add `x509::Version`.

### DIFF
--- a/der/derive/src/enumerated.rs
+++ b/der/derive/src/enumerated.rs
@@ -6,7 +6,7 @@ use crate::ATTR_NAME;
 use proc_macro2::TokenStream;
 use proc_macro_error::abort;
 use quote::quote;
-use syn::{DeriveInput, Expr, ExprLit, Ident, Lit, LitInt, Variant};
+use syn::{DeriveInput, Expr, ExprLit, Ident, Lit, LitInt, Meta, MetaList, NestedMeta, Variant};
 
 /// Valid options for the `#[repr]` attribute on `Enumerated` types.
 const REPR_TYPES: &[&str] = &["u8", "u16", "u32"];
@@ -18,6 +18,9 @@ pub(crate) struct DeriveEnumerated {
 
     /// Value of the `repr` attribute.
     repr: Ident,
+
+    /// Whether or not to tag the enum as an integer
+    integer: bool,
 
     /// Variants of this enum.
     variants: Vec<EnumeratedVariant>,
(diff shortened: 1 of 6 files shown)


der/derive/src/lib.rs
```diff
@@ -214,7 +215,44 @@ pub fn derive_choice(input: TokenStream) -> TokenStream {
...
+/// ```
+///
+/// Note that the derive macro will write a `TryFrom<...>` impl for the
+/// provided `#[repr]`, which is used by the decoder.
+#[proc_macro_derive(Integer)]
+#[proc_macro_error]
+pub fn derive_integer(input: TokenStream) -> TokenStream {
+    let input = parse_macro_input!(input as DeriveInput);
+    DeriveEnumerated::new(input, true).to_tokens().into()
 }
```
tarcieri, line 256: This seems like you're somewhat needlessly defining another proc macro which is ultimately just a 1-argument wrapper for `Enumerated`.

`Integer` is really not very descriptive and kind of ambiguous here: what it's doing is mapping integers to enums, just like `Enumerated`.

I'd suggest just making the `Enumerated` macro more flexible instead.

der/derive/src/enumerated.rs
```diff
@@ -19,13 +19,16 @@ pub(crate) struct DeriveEnumerated {
     /// Value of the `repr` attribute.
     repr: Ident,
 
+    /// Whether or not to tag the enum as an integer
+    integer: bool,
+
```
tarcieri, line 24: This seems like an inextensible way to override the tag of an `Enumerated`, so why not just add a `#[asn1(tag = "...")]` argument instead that lets you override the `ENUMERATED` tag however you want? Then it could potentially work for more than just `INTEGER`.

der/derive/src/lib.rs
npmccallum: Add I said before, I already tried that. But Rust doesn't remove the type-level attribute, so it is passed through to the final AST and generates a compile error.

tarcieri: That must've been a bug in your code, as I've written many proc macros that do that without issue, including this one (see `TypeAttrs`)

npmccallum:
@tarcieri I have implemented it with the attributes. However, the error I mentioned before has returned:

```
error: cannot find attribute `asn1` in this scope
 --> pkcs10/src/version.rs:9:3
  |
9 | #[asn1(type = "INTEGER")]
  |   ^^^^
```

This is because the derive macro outputs the `asn1` attribute in the final AST. But Rust doesn't know about this attribute.

tarcieri:
@npmccallum the problem is in the definition of the `Enumerated` proc macro:

https://github.com/RustCrypto/formats/blob/master/der/derive/src/lib.rs#L213-L218

```rust
#[proc_macro_derive(Enumerated)]
#[proc_macro_error]
pub fn derive_enumerated(input: TokenStream) -> TokenStream {
    let input = parse_macro_input!(input as DeriveInput);
    DeriveEnumerated::new(input).to_tokens().into()
}
```

You'll need to notate that it accepts an `asn1` attribute. Change:

```
#[proc_macro_derive(Enumerated)]
```

to

```rust
#[proc_macro_derive(Enumerated, attributes(asn1))]
```

npmccallum:
@tarcieri THANK YOU! You have no idea how much I was pulling my hair out on that one.

npmccallum:
@tarcieri This is ready for final review now.

der/derive/src/enumerated.rs
```diff
@@ -36,13 +39,25 @@ impl DeriveEnumerated {
...
+                                    match lit.value().as_str() {
+                                        "ENUMERATED" => integer = false,
+                                        "INTEGER" => integer = true,
+                                        s => abort!(lit, "`type = \"{}\"` is unsupported", s),
+                                    }
+                                }
+                            }
+                        }
+                    }
+                }
```
tarcieri, line 60: This is... less than ideal, but I have some ideas about how to refactor this along with attribute parsing in general which I can do in a followup PR.

tarcieri (approved):

PR #412, opened 2022-02-10
Use FlagSet to provide an ergonomic experience with flag types

npmccallum:


der/Cargo.toml
```diff
@@ -16,6 +16,7 @@ edition = "2021"
 rust-version = "1.57"
 
 [dependencies]
+flagset = { version = "^0.4.3", optional = true }
```
tarcieri, line 19: As I mentioned on #414, the `^` is the default range matching operator for Cargo:

```suggestion
flagset = { version = "0.4.3", optional = true }
```

tarcieri (changes requested):
A couple nits about how the dependency is listed in Cargo.toml but otherwise this looks fine

der/Cargo.toml
npmccallum: Done.

tarcieri (approved):

PR #414, opened 2022-02-11
Large Cleanup of the x509 crate

npmccallum:
This PR depends on:
- [x] #410
- [x] #412
- [x] #415
- [x] #416 
- [x] #422 

Yes, there's a lot of commits here. However, this PR actually represents a significant reduction in code in the x509 crate without loss of functionality. All types are derived and there are no custom implementations.


der/Cargo.toml
```diff
@@ -16,6 +16,7 @@ edition = "2021"
 rust-version = "1.57"
 
 [dependencies]
+flagset = { version = "^0.4.3", optional = true }
```
tarcieri, line 19: Seems this PR incorporates https://github.com/RustCrypto/formats/pull/412 (and I'm guessing some of the others as well)?

BTW the `^` is redundant as Cargo defaults to that type of range comparison:

```suggestion
flagset = { version = "0.4.3", optional = true }
```

npmccallum: @tarcieri Yes. The description field at the top of the PR has check boxes for each of the dependency PRs.

npmccallum:
I have moved the early part of this PR into #422 for independent review.

npmccallum:
@tarcieri @carl-wallace All the dependencies for this PR have now been merged. What remains in this PR is just moving types around to create a better hierarchical module structure and improving documentation. I think the best way to review this PR is to check it out and run `cargo doc` to inspect the layout.

npmccallum:
@tarcieri Any progress on this? I have more types to add, but I want to add them to the post-reorg structure.

x509/src/lib.rs
```diff
@@ -14,12 +14,110 @@ extern crate alloc;
...
+/// ```
+///
+/// [RFC 5280 Section 4.1]: https://datatracker.ietf.org/doc/html/rfc5280#section-4.1
+#[derive(Clone, Debug, Eq, PartialEq, Sequence)]
+#[allow(missing_docs)]
+pub struct Certificate<'a> {
+    pub tbs_certificate: TbsCertificate<'a>,
+    pub signature_algorithm: AlgorithmIdentifier<'a>,
+    pub signature: BitString<'a>,
+}
```
tarcieri, line 123: Hmm, could go either way on moving this stuff into the toplevel module. I think it might be a bit surprising, considering everything else is factored into respectively-named modules.

npmccallum: The reason I did that was because these are the primary data types of the crate. Extensions are all under the `ext` module. The same is true with trust anchor types which are under `anchor`. But with x509, the certificate types are the main types of the crate. So I thought it was most ergonomic to put them in the root. I would do the same thing with types that are "groups of certificates" such as `PkiPath`.

npmccallum: Also, please note that certificate types are already re-exported in the root of the crate.

tarcieri: I think it's somewhat nonidiomatic to have a mixture of types scoped to their own submodules and types defined in `lib.rs`. It's also inconsistent with the other crates in this module, where types are scoped to submodules.

> Extensions are all under the ext module. The same is true with trust anchor types which are under anchor

Yes, but there you are scoping the types to submodules underneath `ext` and `anchor`.

In general I think there are some nice code quality benefits to placing types in respective modules, for example unit tests can be placed in a `mod tests` within each submodule as opposed to a toplevel module which is a big bag of types.

x509/src/ext/mod.rs
```diff
@@ -0,0 +1,45 @@
+//! Standardized X.509 Certificate Extensions
```
tarcieri, line 1: Regarding a `mod.rs`, these files were effectively soft deprecated by [RFC2126: path clarity](https://rust-lang.github.io/rfcs/2126-path-clarity.html) and the 2018 edition.

Stylistically across the RustCrypto project we don't use `mod.rs` files.

x509/src/lib.rs
npmccallum: I have no interesting in dying on this hill. :)

Are you suggesting to put cert types in a `cert` module? Do you think they should be reexported in the root directory?

tarcieri: I'd just keep `Certificate` in the `certificate.rs` it was previously defined in, and put `Version` in a `version.rs`

x509/src/ext/mod.rs
npmccallum: @tarcieri I have no problem complying with this project's convention.

However, I don't think the claim that this is deprecated by Rust is actually correct. The RFC explicitly states `The use of mod.rs continues to be allowed without any deprecation...`.

There's an "old" way and a "new" way. The new way was created to improve the editor experience. But now the editors have better support for the old way than the new way. So, IMHO, the old way is actually the better choice.

But again, I really don't care all that much and I'm happy to follow this project's conventions.

x509/src/lib.rs
npmccallum: What about `TbsCertificate` and new types like `PkiPath`?

Generally, my strategy has been to keep trivial but related types together. `Version`, `Certificate` and `PkiPath` are all trivial types related to the `TbsCertificate` type. Having a separate file for each seems a bit like overkill. But perhaps not.

`certificate` is rather long. And shortened module names is definitely idiomatic Rust. That's why I proposed `cert`.

Also please note that extensions have a specific re-export strategy. Namely, there are Extension types, which we consider as first-class types and all get defined or re-exported under `ext::pkix`. Then there are related types which are used by the extensions. These are generally not re-exported since they are typically used by simply dereferencing fields on the extensions.

x509/src/ext/mod.rs
tarcieri: Everything else aside, I think it’s a bad idea to use a mixture of styles, and across all of the RustCrypto we use the “new” way by convention

npmccallum: This is now fixed.

x509/src/lib.rs
npmccallum: I tried reorganizing this and couldn't come up with anything I liked. For example, `use x509::Certificate` seems very ergonomic while `use x509::certificate::Certificate` or even `use x509::cert::Certificate` seems very redundant. Further, trying to split up these related types results in an awkward import structure.

I propose the structure that I just pushed.

1. `lib.rs` is kept clean.
2. redundancy is avoided.
3. closely-related types are grouped by file (this follows the style in the rest of the crate).

tarcieri: Many of our other crates re-export types from the toplevel, but define them in modules, keeping those modules private. In fact, that's exactly how it's structured prior to this PR.

I guess I'm just asking for you to put it back the way it was.

tarcieri: Aah, looks better now. Maybe keep the module name as `certificate`? It's not part of the public API, so I don't see a reason to abbreviate it.

npmccallum: Yeah. Fixing now...

npmccallum: Done.

x509/src/lib.rs
```diff
@@ -1,4 +1,4 @@
-#![no_std]
+#![cfg_attr(not(test), no_std)]
```
tarcieri, line 1: Was there a specific reason for this? I couldn't see why it was needed.

The main problem with doing this is potential build failures for `std`-dependent (or `std` prelude-dependent) features that you only see when doing `cargo build` that are masked when doing `cargo test`

npmccallum: I did it because tests didn't run without it. But I forget why. I can investigate this afternoon.

tarcieri (approved):
Will go ahead and approve this. Can take a look at the tests separately.

PR #415, opened 2022-02-11
der: fix missing format argument

npmccallum:

### DIFF
--- a/der/derive/src/attributes.rs
+++ b/der/derive/src/attributes.rs
@@ -331,7 +331,7 @@ impl AttrNameValue {
                 self.value
                     .value()
                     .parse()
-                    .unwrap_or_else(|_| abort!(self.name, "error parsing `{}` attribute")),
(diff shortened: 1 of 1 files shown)


der/derive/src/attributes.rs
```diff
@@ -331,7 +331,7 @@ impl AttrNameValue {
                 self.value
                     .value()
                     .parse()
-                    .unwrap_or_else(|_| abort!(self.name, "error parsing `{}` attribute")),
+                    .unwrap_or_else(|_| abort!(self.name, "error parsing `{}` attribute", name)),
```
tarcieri, line 334: `self.name` and `name` are the same thing here. Maybe just change it to this?
```suggestion
                    .unwrap_or_else(|_| abort!(self.name, "error parsing attribute")),
```

npmccallum: Done.

tarcieri (approved):

PR #416, opened 2022-02-11
der: use type's tag by default on derive(Choice)

npmccallum:
By default, presume that a choice enum's type implements FixedTag. For
many cases, this means that no `#[asn1(...)]` is required at all. In
particular, this means that deriving `Choice` for `SEQUENCE` types
without a context-specific tag is now possible.   

### DIFF
--- a/der/derive/src/choice/variant.rs
+++ b/der/derive/src/choice/variant.rs
@@ -4,7 +4,60 @@ use crate::{FieldAttrs, Tag, TypeAttrs};
 use proc_macro2::TokenStream;
 use proc_macro_error::abort;
 use quote::quote;
-use syn::{Fields, Ident, Variant};
+use syn::{Fields, Ident, Path, Type, Variant};
+
+#[derive(Clone, Debug, PartialEq, Eq)]
+pub(super) enum TagOrPath {
+    Tag(Tag),
+    Path(Path),
+}
+
+impl PartialEq<Tag> for TagOrPath {
+    fn eq(&self, rhs: &Tag) -> bool {
+        match self {
+            Self::Tag(lhs) => lhs == rhs,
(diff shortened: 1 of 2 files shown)


tarcieri:
I think the `Choice` macro could benefit from a change like this, but I'm not sure making `SEQUENCE` the implicit default is actually the best fix here.

Instead, I think it would make sense to avoid making any presumptions about the ASN.1 type, but try to treat it as if it impls `Decode` + `FixedTag` at decoding time, and `Encode` + `Tagged` at encoding time, and use that to either infer or query the appropriate tag for decoding/encoding.

npmccallum:
@tarcieri I believe the latest version of this fixes your concerns.

der/derive/src/choice/variant.rs
```diff
@@ -4,7 +4,63 @@ use crate::{FieldAttrs, Tag, TypeAttrs};
...
+        match &input.fields {
+            Fields::Unnamed(fields) => match fields.unnamed.len() {
+                1 => match &fields.unnamed[0].ty {
+                    Type::Path(path) => return path.path.clone().into(),
+                    _ => (),
+                },
+                _ => (),
+            },
+            _ => (),
+        }
```
tarcieri, line 47:
```suggestion
        match &input.fields {
            Fields::Unnamed(fields) if fields.unnamed.len() == 1 =>  {
                match &fields.unnamed[0].ty {
                    Type::Path(path) => return path.path.clone().into(),
                    _ => (),
                },
            },
            _ => (),
        }
```

tarcieri: Nice! Looks good now

tarcieri (approved):

PR #422, opened 2022-02-11
x509: derive all implementations

npmccallum:
This PR derives implementations for x509 types. I have taken great care to ensure that every commit passes all existing tests. This PR is the early portion of the larger #414 PR.

### DIFF
--- a/x509/src/certificate.rs
+++ b/x509/src/certificate.rs
@@ -1,7 +1,7 @@
 //! Certificate [`Certificate`] and TBSCertificate [`TBSCertificate`] as defined in RFC 5280
 
-use der::asn1::{BitString, ContextSpecific, ObjectIdentifier, UIntBytes};
-use der::{Enumerated, Sequence, TagMode, TagNumber};
+use der::asn1::{BitString, ObjectIdentifier, UIntBytes};
+use der::{Enumerated, Sequence};
 use spki::{AlgorithmIdentifier, SubjectPublicKeyInfo};
 use x501::name::Name;
 use x501::time::Validity;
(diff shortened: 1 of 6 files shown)


x509/src/certificate.rs
```diff
@@ -33,160 +33,52 @@ impl Default for Version {
...
 ///
-/// [RFC 5280 Section 4.1.2.5]: https://datatracker.ietf.org/doc/html/rfc5280#section-4.1.2.5
-#[derive(Clone, Eq, PartialEq)]
-pub struct TBSCertificate<'a> {
-    /// version         [0]  Version DEFAULT v1,
-    //#[asn1(context_specific = "0", default = "Default::default")]
+/// [RFC 5280 Section 4.1.2.5]: https://datatracker.ietf.org/doc/html/rfc5280#section-4.1
+#[derive(Clone, Debug, Eq, PartialEq, Sequence)]
+#[allow(missing_docs)]
+pub struct TbsCertificate<'a> {
```
tarcieri, line 63: 👍 on the name change. Was going to submit a PR for this myself

tarcieri (approved):

tarcieri:
@carl-wallace mind taking a look at this one?

x509/src/certificate.rs
carl-wallace: It's odd clippy did not complain. The CRL and OCSP bits won't have this problem.

x509/src/certificate.rs
```diff
@@ -33,160 +33,52 @@ impl Default for Version {
...
-///       subjectPublicKeyInfo SubjectPublicKeyInfo,
-///       ... ,
-///       [[2:               -- If present, version MUST be v2
-///       issuerUniqueID  [1]  IMPLICIT UniqueIdentifier OPTIONAL,
-///       subjectUniqueID [2]  IMPLICIT UniqueIdentifier OPTIONAL
-///       ]],
-///       [[3:               -- If present, version MUST be v3 --
-///       extensions      [3]  Extensions{{CertExtensions}} OPTIONAL
-///       ]], ... }
+/// TBSCertificate  ::=  SEQUENCE  {
```
carl-wallace, line 43: The brackets in comments should be escaped to make cargo doc happy.

x509/src/certificate.rs
tarcieri: Yeah, I'm a bit unclear on when `clippy` does or doesn't warn for these, but regardless this is the idiomatic capitalization in Rust.

Agree it would be good to retain the version comment. I'm actually kind of surprised it didn't trigger the `missing_docs` lint.

tarcieri (changes requested):
Please add back the schema comment on `TbsCertificate::version`

x509/src/certificate.rs
npmccallum: They are in a text block like they should be in order to avoid this problem. No escaping is needed.

tarcieri: Yeah, we should really add a test that `cargo doc` completes without warnings

carl-wallace: Got it. Thanks.

npmccallum:
@tarcieri Rather than just copy/pasting schema into the version field comment, instead I wrote a proper comment explaining the full meaning of the situation.

tarcieri:
@npmccallum fantastic!

tarcieri (approved):

npmccallum:
@tarcieri Ready to merge?

PR #424, opened 2022-02-11
pkcs1: derive(Sequence) on types

npmccallum:


tarcieri:
I’d prefer not to add proc macro dependencies to `pkcs1`. It’s effectively “done” and the proc macros add a lot of dependencies without saving that much code.

The next release of the `rsa` crate has a hard dependency on `pkcs1` but will otherwise not pull in the proc macro stack.

The same goes for `pkcs8` and `spki` but all of the other crates are fair game.

Issue #421, opened 2022-02-11
der: `*Document` type improvements


tarcieri:
Several crates that implement ASN.1 DER-based formats in this repo define "document" types such as:

- `sec1::EcPrivateKeyDocument`
- `spki::PublicKeyDocument`
- `pkcs1::{RsaPrivateKeyDocument, RsaPublicKeyDocument}`
- `pkcs10::CertReqDocument`
- `pkcs8::{PrivateKeyDocument, EncryptedPrivateKeyDocument}`

These types are all newtype wrappers around a `Vec<u8>` containing a DER-encoded document which is guaranteed to infallibly parse to a particular message type which impls the `der` crate's `Decodable` and `Encodable` traits.

Where the `der` crate is designed around zero-copy borrowed types in order to accommodate heapless `no_std` targets,  the "document" types provide an owned analogue of each message type without the need to duplicate each struct definition so as to have an owned and borrowed equivalent of each type.

This is codified in the [`der::Document`](https://docs.rs/der/latest/der/trait.Document.html) trait, which has an associated [`Message`](https://docs.rs/der/latest/der/trait.Document.html#associatedtype.Message) type which incorporates the aforementioned bounds. The [`der::Document::decode`](https://docs.rs/der/latest/der/trait.Document.html#method.decode) method infallibly parses the contents of the `Vec<u8>` into the associated `Message` type, having already checked in advance that it parses correctly at the time it's instantiated.

The `der::Document` trait also wraps up common functionality around decoding PEM to DER as well as encoding DER to PEM, and also provides utility functions for reading/writing documents from/to the filesystem, also incorporating awareness as to whether or not the document is a secret and doing things like setting file permissions appropriately.

## Generic document type

It would be nice to not have to define separate document types for every message type, instead changing `der::Document` from a trait into a struct which is generic around the inner message type. 

This has been blocked on the problem of specifying the lifetime to associate with `Decodable<'a>`. However, that might be resolvable by avoiding any bounds on the generic type in the struct definition, and hanging the lifetime off of the `der::Document::decode` method instead. Something like:

```rust
use core::marker::PhantomData;

pub struct Document<M> {
    der: Vec<u8>,
    message: PhantomData<M>
}

impl<M> Document<M> {
    pub fn decode<'a>(&'a self) -> M
    where
        M: Decode<'a> + Encode + Sized
    {
        [...]
    }
}
```

The bounds will need to be repeated for every method on the struct.

## Avoid reparsing messages

[...]

Issue #11, opened 2021-09-14
pem-rfc7468: allow parsing multiple encapsulated objects from single PEM file


akhilles:
I think it's pretty common to have a single `.pem` file that contains a bunch of public keys. Not sure what the best API for this is, maybe return the remaining string/byte slice on successful parse? I wrote a PEM encoder/decoder that follows this pattern, might be useful: https://gist.github.com/akhilles/1ebaa0a6f15e46d6b4414bdf7d8c8ced.

JonathanWilbur:
> It is? It's definitely common for X.509 certificate chains to be encoded as concatenated PEM files, but I don't think I've seen "a bunch of public keys" encoded that way before. What's the use case?

I think @akhilles might have misspoken. It is a common use case to have multiple PEM-encoded _public key certificates_ concatenated (not _public keys_). Many applications, such as Postfix, support PEM-encoded certificate chains in this fashion. Here is an example of where this practice is documented by a large PKI company: https://www.digicert.com/kb/ssl-support/pem-ssl-creation.htm.

On a side note, the PEM file format is supposed to tolerate "explanatory text" before, after, and between such objects: https://www.rfc-editor.org/rfc/rfc7468#section-5.2.

I think it would be constructive to implement this as `Iterator<(type_label, data)>`.

This functionality is definitely needed. I need it now and I will implement it and submit a PR, if you're open to it.

tarcieri:
> I think @akhilles might have misspoken. It is a common use case to have multiple PEM-encoded public key certificates concatenated

That's exactly what I said and what you just quoted?

> On a side note, the PEM file format is supposed to tolerate "explanatory text" before...

Support for the "before" case for explanatory text was implemented in #40.

Again, I'm fine with adding support for iterating over sequences of encapsulated objects. Just please remember this crate has a guarantee of constant time operation which shouldn't be violated in service of implementing this feature.

PR #427, opened 2022-02-12
[WIP] der: generic `Document` type

tarcieri:
Changes `Document` from a trait into a generic type.

The previous approach necessitated a lot of repetitive boilerplate for each "document" type. Using a generic type eliminates the boilerplate.

The definition was somewhat tricky due to the lifetime on `Decodable<'a>`. For more context, see #421.

### DIFF
--- a/der/src/document.rs
+++ b/der/src/document.rs
@@ -1,67 +1,103 @@
 //! ASN.1 DER-encoded documents stored on the heap.
 
 use crate::{Decodable, Encodable, Error, Result};
-use alloc::{boxed::Box, vec::Vec};
+use alloc::vec::Vec;
+use core::marker::PhantomData;
 
 #[cfg(feature = "pem")]
-use {crate::pem, alloc::string::String};
+use {
+    crate::pem::{self, PemLabel},
+    alloc::string::String,
+};
 
 #[cfg(feature = "std")]
 use std::{fs, path::Path};
 
+/// Marker trait which identifies whether or not documents contain sensitive
+/// information, such as private keys.
+#[cfg_attr(docsrs, doc(cfg(feature = "alloc")))]
+pub trait Sensitivity {
+    /// Does this type contain potentially sensitive data?
+    ///
+    /// This enables hardened file permissions when persisting data to disk.
+    const SENSITIVE: bool;
+}
+
 /// ASN.1 DER-encoded document.
 ///
-/// This trait is intended to impl on types which contain an ASN.1 DER-encoded
-/// document which is guaranteed to encode as the associated `Message` type.
+/// This type wraps an encoded ASN.1 DER message which is guaranteed to
+/// infallibly decode as type `T`.
 ///
 /// It implements common functionality related to encoding/decoding such
 /// documents, such as PEM encapsulation as well as reading/writing documents
 /// from/to the filesystem.
 #[cfg_attr(docsrs, doc(cfg(feature = "alloc")))]
-pub trait Document<'a>: AsRef<[u8]> + Sized + TryFrom<Vec<u8>, Error = Error> {
-    /// ASN.1 message type this document decodes to.
-    type Message: Decodable<'a> + Encodable + Sized;
+pub struct Document<T: Sensitivity> {
+    /// ASN.1 DER-encoded document guaranteed to decode to `T` infallibly.
+    der_bytes: Vec<u8>,
 
-    /// Does this type contain potentially sensitive data?
-    ///
-    /// This enables hardened file permissions when persisting data to disk.
-    const SENSITIVE: bool;
+    /// Rust type corresponding to the ASN.1 DER message the bytes can be
+    /// infallibly deserialized as.
+    msg_type: PhantomData<T>,
+}
 
+impl<T> Document<T> where T: Sensitivity {}
+
+impl<T> Document<T>
+where
+    T: Sensitivity,
+{
     /// Borrow the inner serialized bytes of this document.
-    fn as_der(&self) -> &[u8] {
-        self.as_ref()
-    }
-
-    /// Return an allocated ASN.1 DER serialization as a boxed slice.
-    fn to_der(&self) -> Box<[u8]> {
-        self.as_ref().to_vec().into_boxed_slice()
+    pub fn as_der(&self) -> &[u8] {
+        self.der_bytes.as_slice()
     }
 
     /// Decode this document as ASN.1 DER.
-    fn decode(&'a self) -> Self::Message {
-        Self::Message::from_der(self.as_ref()).expect("ASN.1 DER document malformed")
+    pub fn decode<'a>(&'a self) -> T
+    where
+        T: Decodable<'a> + Sized,
+    {
+        self.try_decode().expect("ASN.1 DER document malformed")
     }
 
     /// Create a new document from the provided ASN.1 DER bytes.
-    fn from_der(bytes: &[u8]) -> Result<Self> {
-        bytes.to_vec().try_into()
+    pub fn from_der(bytes: impl Into<Vec<u8>>) -> Result<Self>
+    where
+        T: for<'a> Decodable<'a> + Sized,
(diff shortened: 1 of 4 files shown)


tarcieri:
NOTE: early WIP. Not expected to pass tests yet, but it compiles.

der/src/document.rs
```diff
@@ -1,67 +1,102 @@
...
 use std::{fs, path::Path};
 
+/// Marker trait which identifies documents which contain sensitive
+/// information, such as private keys.
+pub trait Sensitivity {
+    /// Does this type contain potentially sensitive data?
+    ///
+    /// This enables hardened file permissions when persisting data to disk.
+    const SENSITIVE: bool;
+}
```
tarcieri, line 23: Marker trait for types which represent private keys. The effects should probably be better documented in rustdoc, but basically come down to:

1. Zeroizing `der_bytes` when the `zeroize` crate feature is enabled
2. Setting appropriate file permissions when writing DER or PEM data to disk

Alternatively this could be a const generic parameter. It's something to consider, but exposing the information under a trait might be useful in other settings.

npmccallum: If we gave this trait another name, we could also include the PEM "type" as a `&'static str` here too.

der/src/document.rs
```diff
@@ -1,67 +1,102 @@
...
-    fn to_der(&self) -> Box<[u8]> {
-        self.as_ref().to_vec().into_boxed_slice()
+impl<T> Document<T>
+where
+    T: Sensitivity,
+{
+    /// Borrow the inner serialized bytes of this document.
+    pub fn as_der(&self) -> &[u8] {
+        self.der_bytes.as_slice()
     }
```
npmccallum, line 52: We don't actually know if the bytes are `der` because we haven't decoded them yet. Perhaps we should just let `AsRef` or even `Deref` do the heavy lifting here?

der/src/document.rs
```diff
@@ -1,67 +1,102 @@
...
+    {
+        let doc = Self {
+            der_bytes: bytes.into(),
+            msg_type: PhantomData,
+        };
+
+        // Ensure document parses successfully
+        doc.try_decode()?;
+        Ok(doc)
+    }
```
npmccallum, line 75: I don't think we should try decoding the der. This has performance and security implications. I think the `Document` type should only have the invariant that it contains bytes successfully decoded from a PEM file. The important bit is that `T` is linked via the type system to the PEM document "name" and that we know the sensitivity of the bytes.

der/src/document.rs
```diff
@@ -71,49 +106,99 @@ pub trait Document<'a>: AsRef<[u8]> + Sized + TryFrom<Vec<u8>, Error = Error> {
...
+    ///
+    /// This method doesn't uphold the invariant that `T` always decodes
+    /// successfully, but is needed to make the lifetimes for the constructor
+    /// work.
+    fn try_decode<'a>(&'a self) -> Result<T>
+    where
+        T: Decodable<'a> + Sized,
+    {
+        T::from_der(self.as_der())
+    }
```
npmccallum, line 167: I think we should ditch that invariant altogether.

der/src/document.rs
```diff
@@ -71,49 +106,99 @@ pub trait Document<'a>: AsRef<[u8]> + Sized + TryFrom<Vec<u8>, Error = Error> {
...
+    T: Sensitivity,
+{
+    fn drop(&mut self) {
+        #[cfg(feature = "zeroize")]
+        if T::SENSITIVE {
+            use zeroize::Zeroize;
+            self.der_bytes.zeroize();
+        }
+    }
+}
```
npmccallum, line 192: It would probably be easier to just unconditionally `Zeroize<Vec<u8>>`. I presume you're worried about performance of zeroing large certificates. If you're willing to take the performance tradeoffs, zeroing even the non-sensitive bits has security benefits. For example, attackers often search for the public key to correlate the private key's memory position (or the position of some of its non-zeroed fragments). Zeroing both adds one additional hurdle.

npmccallum (commented):
@tarcieri Over all, I like this direction! Here's some thoughts and various nitpicks.

der/src/document.rs
tarcieri: If you take a look at the definition of `from_der`, you'll see it checks the document decodes using `doc.try_decode()` before returning it to the caller. The type definitely does maintain the invariant that `self.der_bytes` will always decode successfully, and as I noted in #421, it may be possible to store the parsed message to prevent repeat decoding (definitely not going to attempt that in this PR).

Also for some reason GitHub isn't letting me reply to your comment about the PEM label, but there's already a `PemLabel` trait for that.

And FWIW locally I'm trying to replace the trait with a const generic parameter, and liking that better for now.

der/src/document.rs
tarcieri: It's an option, but the sensitivity information is already there for other purposes, so it seems ok to me to make use of it for this application as well.

tarcieri:
Pushed up a version that uses a const generic `SENSITIVE` parameter.

I like it: it allows specializing things like trait impls depending on whether or not `SENSITIVE` is false or true. I used that for the `Debug` impls, which can also have slightly different bounds (a `SENSITIVE` document doesn't need a `Debug` impl for the inner type, for example)

der/src/document.rs
npmccallum: `Document` combines three functionalities:
1. Zeroing the `Vec<u8>`. This has clear value.
2. Ensuring file permissions on write. This has clear value. I have some minor concern, however, that this is only useful in a narrow set of cases (i.e. where you want to do full, synchronous io). If you want to do partial io (such as a certificate chain incremental io) or asynchronous io, this type drops this value.
3. Ensuring the bytes will successfully decode to `T`. The value of this is unclear to me. And it carries a performance and security cost.

Dropping the third guarantee still leaves `Document` with two features. So it is still better than a `Vec<u8>`.

npmccallum:
> Pushed up a version that uses a const generic `SENSITIVE` parameter.
> 
> I like it: it allows specializing things like trait impls depending on whether or not `SENSITIVE` is false or true. I used that for the `Debug` impls, which can also have slightly different bounds (a `SENSITIVE` document doesn't need a `Debug` impl for the inner type, for example)

However, the tradeoff here is that `SENSITIVE` is now detached from the underlying data type, which makes it error prone (with potentially dangerous failure modes). A `PrivateKey` is always `SENSITIVE`. The user of these types should not have the option of changing that.

npmccallum:
One other feature to consider is how to read/write files that contain multiple `Document` objects. Certificate chain files are probably the most obvious case of this. And we probably want a `PkiPath` type in `x509` that should be able to be serialized/deserialized to/from PEM.

tarcieri:
> However, the tradeoff here is that SENSITIVE is now detached from the underlying data type, which makes it error prone

That's potentially true, however if we define type aliases for the relevant documents, someone would need to go out of their way in order to turn it off.

I think there's pros and cons either way.

> One other feature to consider is how to read/write files that contain multiple Document objects. 

See #11

tarcieri:
I'm not sure this will actually work out in practice. Several of the methods are using HRTBs, but the problem comes when you actually need to notate that lifetime for the borrowed `T`:

```
error[E0277]: the trait bound `for<'a> SubjectPublicKeyInfo<'_>: Decodable<'a>` is not satisfied
  --> spki/src/document.rs:21:12
   |
21 |         Ok(Self::from_der(bytes)?)
   |            ^^^^^^^^^^^^^^ the trait `for<'a> Decodable<'a>` is not implemented for `SubjectPublicKeyInfo<'_>`
```

So yes, alternatively this could just be a bag-of-bytes newtype which perhaps might (or might not) perform some basic sanity checks on the inner DER message, with no generic parameter.

That unfortunately loses any type safety around the message it contains, though, and will introduce more fallibility where there currently is none.

tarcieri:
Closing in favor of #571

PR #428, opened 2022-02-14
docs: fixed internal links

Scarjit:
This pr fixes some dead internal links in the base64ct, pem_rfc7468 & pkcs1 crate.

x509 has deadlinks too but fixing them seems to be irrelevant, until #414 is merged.

### DIFF
--- a/base64ct/src/decoder.rs
+++ b/base64ct/src/decoder.rs
@@ -26,7 +26,7 @@ const CHAR_LF: u8 = 0x0a;
 /// Stateful Base64 decoder with support for buffered, incremental decoding.
 ///
 /// The `E` type parameter can be any type which impls [`Encoding`] such as
-/// [`Base64`] or [`Base64Unpadded`].
+/// [Base64][`crate::variant::standard::Base64`] or [Base64Unpadded][`crate::variant::standard::Base64Unpadded`].
 ///
 /// Internally it uses a sealed `Variant` trait which is an implementation
 /// detail of this crate, and leverages a [blanket impl] of [`Encoding`].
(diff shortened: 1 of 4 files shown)


base64ct/src/decoder.rs
```diff
@@ -26,7 +26,7 @@ const CHAR_LF: u8 = 0x0a;
 /// Stateful Base64 decoder with support for buffered, incremental decoding.
 ///
 /// The `E` type parameter can be any type which impls [`Encoding`] such as
-/// [`Base64`] or [`Base64Unpadded`].
+/// [Base64][`crate::variant::standard::Base64`] or [Base64Unpadded][`crate::variant::standard::Base64Unpadded`].
```
tarcieri, line 29: Check line 17 of this file. There are imports of these, although they should probably be gated on `#[cfg(doc)]` instead of `#[cfg(docsrs)]`.

base64ct/src/encoder.rs
```diff
@@ -13,7 +13,7 @@ use crate::{Base64, Base64Unpadded};
 /// Stateful Base64 encoder with support for buffered, incremental encoding.
 ///
 /// The `E` type parameter can be any type which impls [`Encoding`] such as
-/// [`Base64`] or [`Base64Unpadded`].
+/// [Base64][`crate::variant::standard::Base64`] or [Base64Unpadded][`crate::variant::standard::Base64Unpadded`].
```
tarcieri, line 16: Likewise here. Line 10.

tarcieri:
Just in general I would prefer importing the types for generating the docs instead of using both the type names and the excessively long fully qualified paths. They make the docs much harder to read if you're just looking at the source code.

tarcieri:
Also it'd be good to add CI for this. Some of these were "working" but not with the same rustdoc configuration.

tarcieri:
Note that we now CI for rustdoc building warning free (#530), so this PR seems obsolete

PR #429, opened 2022-02-16
Unconditionally specify no_std

npmccallum:


npmccallum:
@tarcieri Is this what you had in mind?

tarcieri:
Yeah, that works.

We might also consider putting these tests under `tests/` which unconditionally links `std`.

tarcieri (approved):

PR #431, opened 2022-02-17
Fix typos in key purpose OID definitions

carl-wallace:


npmccallum (changes requested):
There are multiple unrelated commits in this PR. Can you please separate them into separate PRs? There are also stray merge commits that shouldn't belong in any PR. Please remove them.

PR #438, opened 2022-02-17
Fix and test min versions build

### LINKED ISSUE #439: Use pre-compiled cargo hack
Right now CI time is wasted on re-compiling `cargo hack` from scratch each time. It can be done similarly to how we work with `cross`.

newpavlov:
All fixed crates except `tls_codec(_derive)` are on the pre versions, so I haven't mentioned the fix in changelogs.

@tarcieri
I will leave for you to decide whether a patch release for `tls_codec` is needed or not.

Closes #439

### DIFF
--- a/.github/workflows/der.yml
+++ b/.github/workflows/der.yml
@@ -36,8 +36,13 @@ jobs:
           toolchain: ${{ matrix.rust }}
           target: ${{ matrix.target }}
           override: true
-      - run: cargo install cargo-hack
-      - run: cargo hack build --release --target ${{ matrix.target }} --feature-powerset --exclude-features std
+      - uses: RustCrypto/actions/cargo-hack-install@master
+      - run: cargo hack build --target ${{ matrix.target }} --feature-powerset --exclude-features std
+
+  minimal-versions:
+    uses: RustCrypto/actions/.github/workflows/minimal-versions.yml@master
+    with:
+        working-directory: ${{ github.workflow }}
 
   test:
     strategy:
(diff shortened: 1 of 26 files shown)


tls_codec/Cargo.toml
```diff
@@ -11,7 +11,7 @@ edition = "2021"
 rust-version = "1.56"
 
 [dependencies]
-zeroize = { version = "1", default-features = false, features = ["alloc", "zeroize_derive"] }
+zeroize = { version = "1", default-features = false, features = ["alloc"] }
```
newpavlov, line 14: IIUC the `zeroize_derive` feature is not used in any way by this crate, so I removed it.

.github/workflows/der.yml
```diff
@@ -61,13 +66,14 @@ jobs:
...
           # 64-bit Windows
-          - target: x86_64-pc-windows-msvc
-            platform: windows-latest
-            rust: 1.57.0 # MSRV
-          - target: x86_64-pc-windows-msvc
-            platform: windows-latest
-            rust: stable
+          #- target: x86_64-pc-windows-msvc
+          #  platform: windows-latest
+          #  rust: 1.57.0 # MSRV
```
newpavlov, line 73: Windows tests are disabled for now since the `cargo-hack-install` action does not currently work on Windows. It fails with `wget: command not found`, IIUC `wget` should be installed by default, but, unfortunately, we have to use full path to it.

AFAIK `der` is platform independent, so disabling those tests should not be a big issue.

PR #442, opened 2022-02-21
tls codec: variable-length vectors

franziskuskiefer:
This implements the variable-length integer encoding from [RFC 9000] (quic) used to serialize variable-length vectors in quic and in [MLS].
It's opinionated because it doesn't allow anything longer than `u16::MAX` when deserialising (same as the other vectors), and it requires minimum lengths like MLS (unlike quic).

cc @kkohbrok @raphaelrobert if one of you wants to have a look at this.

[rfc 9000]: https://www.rfc-editor.org/rfc/rfc9000.html#name-variable-length-integer-enc
[mls]: https://messaginglayersecurity.rocks/mls-protocol/draft-ietf-mls-protocol.html#name-variable-size-vector-header

### DIFF
--- a/tls_codec/src/quic_vec.rs
+++ b/tls_codec/src/quic_vec.rs
@@ -37,7 +37,7 @@ fn read_variable_length<R: std::io::Read>(bytes: &mut R) -> Result<(usize, usize
     }
     for _ in 0..len_len {
         let mut next = [0u8; 1];
-        bytes.read(&mut next)?;
+        bytes.read_exact(&mut next)?;
         length = (length << 8) + usize::from(next[0]);
     }
 
@@ -61,7 +61,6 @@ fn length_encoding_bytes(length: u64) -> usize {
 // === (De)Serialize for `Vec<T>` and &[T].
 
 impl<T: Serialize + std::fmt::Debug> Serialize for Vec<T> {
-    #[cfg(feature = "std")]
     #[inline(always)]
     fn tls_serialize<W: std::io::Write>(&self, writer: &mut W) -> Result<usize, Error> {
         self.as_slice().tls_serialize(writer)
@@ -69,7 +68,6 @@ impl<T: Serialize + std::fmt::Debug> Serialize for Vec<T> {
 }
 
 impl<T: Size> Size for Vec<T> {
-    #[cfg(feature = "std")]
     #[inline(always)]
     fn tls_serialized_len(&self) -> usize {
         self.as_slice().tls_serialized_len()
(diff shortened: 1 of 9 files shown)


raphaelrobert:
Why specifically `u16::MAX`? Is that to ensure it will work on 16bit platforms? If not for the latter, I would recommend using `u32::MAX` instead.

tls_codec/src/quic_vec.rs
```diff
@@ -0,0 +1,339 @@
...
+    #[cfg(feature = "std")]
+    #[inline(always)]
+    fn tls_deserialize<R: std::io::Read>(bytes: &mut R) -> Result<Self, Error> {
+        let (length, len_len) = read_variable_length(bytes)?;
+        if length == 0 {
+            // An empty vector.
+            return Ok(Vec::new());
+        }
+
+        let mut result = Vec::new();
```
kkohbrok, line 86: Wouldn't it make sense to use something like `with_capacity((length-len_len)/element.tls_serialized_len())` here?

franziskuskiefer:
> Why specifically `u16::MAX`? Is that to ensure it will work on 16bit platforms? If not for the latter, I would recommend using `u32::MAX` instead.

It's just what's used everywhere else already. As long as no one is actually running into the limit I don't see a reason to bump it.

raphaelrobert:
> > Why specifically `u16::MAX`? Is that to ensure it will work on 16bit platforms? If not for the latter, I would recommend using `u32::MAX` instead.
> 
> It's just what's used everywhere else already. As long as no one is actually running into the limit I don't see a reason to bump it.

In the context of MLS this is limiting. It means the maximum size of encrypted payload is 64kb. When you use the ratchet tree extension for example (where the public tree gets encrypted in its entirety), this would severely limit the group size (to maybe less than 1000 members, depending on the size of KeyPackages). 
I wasn't aware that this limit also existed for `Vec32`, and I think it's equally problematic.

kkohbrok (commented):
Looks good to me in general, although I have to admit, I didn't follow the byte operations in detail.

tls_codec/src/quic_vec.rs
franziskuskiefer: We could. It's a trade-off between allocating unnecessary memory and speeding up things. As long as no one measures performance here and identifies this as a bottleneck I'm conservative and don't allocate unnecessary memory.

kkohbrok: I see. Do I understand correctly, though, that memory would only be unnecessarily allocated if the encoded length is inaccurate?

franziskuskiefer: No, any broken element in the vector will cause the function to abort. The scenario I have in mind here is for example a vector that claims to be extremely long where the first element doesn't deserialize correctly. Then we'd allocate a ton of memory here and bail trying to read the first one.

PR #443, opened 2022-02-21
Add AlgorithmIdentifier::oids() helper function

npmccallum:

### DIFF
--- a/spki/src/algorithm.rs
+++ b/spki/src/algorithm.rs
@@ -74,6 +74,26 @@ impl<'a> AlgorithmIdentifier<'a> {
     pub fn parameters_oid(&self) -> Result<ObjectIdentifier> {
         Ok(ObjectIdentifier::try_from(self.parameters_any()?)?)
     }
+
+    /// Convert to a pair of [`ObjectIdentifier`]s.
+    ///
+    /// This method is helpful for decomposing in match statements. Note in
+    /// particular that `NULL` parameters are treated the same as missing
+    /// parameters.
+    ///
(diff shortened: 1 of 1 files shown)


spki/src/algorithm.rs
```diff
@@ -74,6 +74,26 @@ impl<'a> AlgorithmIdentifier<'a> {
...
+    pub fn oids(&self) -> der::Result<(ObjectIdentifier, Option<ObjectIdentifier>)> {
+        Ok((
+            self.oid,
+            match self.parameters {
+                None => None,
+                Some(p) => match p {
+                    Any::NULL => None,
+                    _ => Some(p.oid()?),
+                },
+            },
```
tarcieri, line 94:
```suggestion
            self.parameters.map(|p| match p {
                Any::NULL => None,
                _ => Some(p.oid()?),
            })
```

tarcieri (changes requested):
One nit.

I'd be curious to see a use case for this. There are a lot of helper methods like this already.

npmccallum:
@tarcieri It makes it easy to do things like compare key algorithms with signature algorithms. See:

https://github.com/profianinc/steward/blob/main/src/signed.rs#L169-L170

spki/src/algorithm.rs
npmccallum: @tarcieri This actually doesn't compile because it embeds the error handling in the function. I think my version is actually cleaner due to this.

tarcieri (approved):

PR #450, opened 2022-02-26
feat: add new NewType derive

npmccallum:
There are a number of der types that could benefit from wrapping in a
newtype. For example, `RelativeDistinguishedName` could benefit from
an `impl Display` to convert it to a string representation. But because
it is just `Vec<AttributeTypeAndValue>` we can't really do this.

Having a systematic way to derive newtypes in these cases is thus
beneficial.

### DIFF
--- a/der/derive/src/lib.rs
+++ b/der/derive/src/lib.rs
@@ -120,6 +120,7 @@ mod asn1_type;
 mod attributes;
 mod choice;
 mod enumerated;
+mod newtype;
 mod sequence;
 mod tag;
 mod value_ord;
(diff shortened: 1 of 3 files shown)


tarcieri:
I guess this is fine. My only remaining concern is that "newtype" is pretty much exclusively used as a closed compound word, in which case the proper Rust capitalization is "Newtype"

npmccallum:
@tarcieri I have renamed `NewType => Newtype` and `new_type => newtype`.

tarcieri (approved):

PR #453, opened 2022-02-28
Additional OID DB improvments

npmccallum:


const-oid/oiddbgen/src/root.rs
```diff
@@ -1,80 +1,38 @@
...
-                    .entry(spec)
-                    .or_insert_with(Spec::default)
-                    .push(Node::new(obji, name));
-            }
-        }
+    pub fn add(
+        &mut self,
+        spec: impl Deref<Target = str>,
+        name: impl Deref<Target = str>,
+        obid: impl Deref<Target = str>,
```
tarcieri, line 16: Implicit deref coercion should make the `impl Deref`s unnecessary:

```suggestion
        spec: &str,
        name: &str,
        obid: &str,
```

npmccallum: It forces `main.rs` to take a reference. But I don't care either way.

npmccallum: Fixed.

tarcieri: IMO it's pretty much the main reason the `Deref` trait exists: so "smart pointer" types that impl `Deref` can be treated as regular references

const-oid/oiddbgen/src/asn1.rs
```diff
@@ -0,0 +1,118 @@
+use std::collections::{BTreeMap, HashSet};
+
+use regex::Regex;
+
+#[derive(Clone, Debug)]
+pub struct Asn1Parser(String, BTreeMap<String, (Option<String>, Option<String>)>);
```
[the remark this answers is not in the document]
npmccallum: Fixed.

tarcieri (approved):

PR #458, opened 2022-03-02
const-oid: fallible const parser + `::new_unwrap`

tarcieri:
With the introduction of const panic it's now possible to implement a proper fallible parser with `const fn`, eliminating some of the duplication between the `const fn` and fallible code in this crate.

Unfortunately it's not yet possible to call `Result::unwrap` in a const context, so instead this commit adds an `ObjectIdentifier::new_unwrap` method which leverages const panic to print the `Result` as best it can.

In future versions of Rust, it will be possible to replace this with `ObjectIdentifier::new(...).unwrap()`.


tarcieri:
FYI @npmccallum

npmccallum (approved):
Approved, pending tests passing.

PR #460, opened 2022-03-03
fix typo in OID definition

carl-wallace:


tarcieri:
This just changed in #458 

Hopefully `ObjectIdentifier::new_unwrap(...)` can go away soon and we can use `ObjectIdentifier::new(...).unwrap()`, but that needs `Result::unwrap` to be a `const fn`.

PR #461, opened 2022-03-03
Remove the NamedOid type

npmccallum:
Since it isn't possible to match an ObjectIdentifier against a NamedOid,
this allows the OIDs to be used directly as patterns while still
preserving the ability to look up names for the OIDs.


tarcieri (approved):

npmccallum:
@tarcieri I fixed the tests.

Issue #417, opened 2022-02-11
x501: support string encoding/decoding for Distinguished Names


npmccallum:
https://datatracker.ietf.org/doc/html/rfc4514

The above RFC defines the syntax.

PR #464, opened 2022-03-04
feat(x509): implement RFC 4514 string serialization for x509 names

npmccallum:
The main commit is the last one. The previous commits are just enablement.

This PR depends on:
- [x] #467 
- [x] #468 
- [x] #469
- [x] #470

Resolves: #417

### DIFF
--- a/x509/src/attr.rs
+++ b/x509/src/attr.rs
@@ -1,7 +1,11 @@
 //! Attribute-related definitions as defined in X.501 (and updated by RFC 5280).
 
+use alloc::vec::Vec;
+use core::fmt::{self, Write};
+
+use const_oid::db::DB;
 use der::asn1::{Any, ObjectIdentifier, SetOfVec};
-use der::{Decodable, Sequence, ValueOrd};
+use der::{Decodable, Encodable, Error, ErrorKind, Sequence, Tag, Tagged, ValueOrd};
 
 /// X.501 `AttributeType` as defined in [RFC 5280 Appendix A.1].
 ///
@@ -51,7 +55,7 @@ pub struct Attribute<'a> {
 }
 
 impl<'a> TryFrom<&'a [u8]> for Attribute<'a> {
-    type Error = der::Error;
+    type Error = Error;
 
     fn try_from(bytes: &'a [u8]) -> Result<Self, Self::Error> {
         Self::from_der(bytes)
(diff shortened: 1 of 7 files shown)


x509/src/attr.rs
```diff
@@ -87,3 +91,164 @@ pub struct AttributeTypeAndValue<'a> {
...
+        let atv = AttributeTypeAndValue { oid, value };
+        atv.to_vec()
+    }
+
+    /// Converts an AttributeTypeAndValue string into an encoded AttributeTypeAndValue
+    ///
+    /// This function follows the rules in [RFC 4514].
+    ///
+    /// [RFC 4514]: https://datatracker.ietf.org/doc/html/rfc4514
+    pub fn parse(s: &str) -> Result<Vec<u8>, der::Error> {
```
tarcieri, line 199: I'm a little confused by the `fn parse` name along with `parse_str` and `parse_der`. These seem to be combination parser/serializer methods?

x509/src/attr.rs
```diff
@@ -1,7 +1,11 @@
 //! Attribute-related definitions as defined in X.501 (and updated by RFC 5280).
 
+use alloc::vec::Vec;
+use core::fmt::Write;
```
tarcieri, line 4: If you use:

```suggestion
use core::fmt::{self, Write};
```

...you can replace the instances of `core::fmt::*` with `fmt::*` below.

x509/src/attr.rs
```diff
@@ -87,3 +91,164 @@ pub struct AttributeTypeAndValue<'a> {
...
+
+        Ok(())
+    }
+
+    pub fn as_bytes(&self) -> &[u8] {
+        &self.bytes
+    }
+}
+
+impl<'a> AttributeTypeAndValue<'a> {
```
tarcieri, line 147: It doesn't look like you're using the `'a` lifetime anywhere, in which case it can be omitted:

```suggestion
impl AttributeTypeAndValue<'_> {
```

...and so on below

x509/src/name.rs
```diff
@@ -20,7 +22,73 @@ pub type Name<'a> = RdnSequence<'a>;
 /// ```
 ///
 /// [RFC 5280 Section 4.1.2.4]: https://datatracker.ietf.org/doc/html/rfc5280#section-4.1.2.4
-pub type RdnSequence<'a> = alloc::vec::Vec<RelativeDistinguishedName<'a>>;
+#[derive(Clone, Debug, Default, PartialEq, Eq, Newtype)]
+pub struct RdnSequence<'a>(alloc::vec::Vec<RelativeDistinguishedName<'a>>);
```
tarcieri, line 26: `Vec` is in scope:

```suggestion
pub struct RdnSequence<'a>(Vec<RelativeDistinguishedName<'a>>);
```

We should probably turn on the `unused_qualifications` lint. I'm kind of surprised it's off.

x509/src/name.rs
```diff
@@ -20,7 +22,73 @@ pub type Name<'a> = RdnSequence<'a>;
...
+fn split(s: &str, b: u8) -> impl '_ + Iterator<Item = &'_ str> {
+    let mut prev = 0;
+    find(s, b).chain([s.len()].into_iter()).map(move |i| {
+        let x = &s[prev..i];
+        prev = i + 1;
+        x
+    })
+}
+
+impl<'a> RdnSequence<'a> {
```
tarcieri, line 57: Unused lifetime:

```suggestion
impl RdnSequence<'_> {
```

...and we should probably also turn on the `unused_lifetimes` lint.

tarcieri (changes requested):
A few small nits, otherwise looks good

x509/src/attr.rs
npmccallum: Yes, I agree it is a bit awkward. Basically, RFC 4514 defines a human readable encoding for certificate name types. That is, we want to convert from the name type into a string (this part is easy) and from the string back into the relevant name type. The latter operation is difficult because the result needs to be owned, but the name types borrow references.

The compromise is that we parse the string into an owned buffer. Then we create the named type referencing that owned buffer. Then, because of the lifetime issues, we serialize the named type to der.

In practice it looks like this to avoid the lifetime issues:

```rust
            let der = RdnSequence::parse("CN=foo").unwrap();
            let rdns = RdnSequence::from_der(&der).unwrap();
```

If you have alternative suggestions, I'm all ears.

x509/src/attr.rs
npmccallum: I actually find that less clear.

tarcieri: I think that's far and away the idiomatic thing to do.

Ballparking from GitHub search results:

- `impl fmt::`: [227,914 code results](https://github.com/search?l=Rust&q=%22impl+fmt%3A%3A%22&type=Code)
- `impl core::fmt::`: [7,588 code results](https://github.com/search?l=Rust&q=%22impl+core%3A%3Afmt%3A%3A%22&type=Code)

x509/src/name.rs
npmccallum: I'll turn it on in another patch.

x509/src/name.rs
npmccallum: I'll turn the lint on in another patch.

x509/src/attr.rs
npmccallum: Fixed.

x509/src/attr.rs
tarcieri: Since it's producing encoded DER, I'd suggest something like `RdnSequence::encode_*`, like `RdnSequence::encode_from_string`

npmccallum: I renamed all the public functions to `encode_from_string()`. The internal functions were renamed to `encode_str()` and `encode_hex()`.

tarcieri (approved):

PR #465, opened 2022-03-04
chore(x509): remove oids now defined in the oiddb

npmccallum:


x509/Cargo.toml
```diff
@@ -16,6 +16,7 @@ rust-version = "1.56"
 
 [dependencies]
 der = { version = "=0.6.0-pre.1", features = ["derive", "alloc", "flagset"], path = "../der" }
+const-oid = { version = "0.8", features = ["db"], path = "../const-oid" }
```
npmccallum, line 19: "This belongs in dev-dependencies!" you might say. And you'd be right. I left this as a regular dependency because #464 adds a hard dependency on the oiddb and this will make the rebase easier.

PR #466, opened 2022-03-05
feat(x509): add PkiPath type

npmccallum:


x509/src/certificate.rs
```diff
@@ -103,3 +105,17 @@ pub struct Certificate<'a> {
...
+/// top-most certificate to the bottom-most certificate. This means that
+/// the first certificate signs the second certificate and so on.
+///
+/// ```text
+/// PkiPath ::= SEQUENCE OF Certificate
+/// ```
+///
+/// [RFC 6066]: https://datatracker.ietf.org/doc/html/rfc6066#section-10.1
+#[derive(Clone, Debug, PartialEq, Eq, Default, Newtype)]
+pub struct PkiPath<'a>(Vec<Certificate<'a>>);
```
tarcieri, line 121: Seems like this could use some sort of accessor for the inner `Certificate`s, such as `AsRef<[Certificate<'a>]>` and/or something that exposes an `Iterator<Item=&Certificate<'a>>`

npmccallum: @tarcieri Like `#[derive(Newtype)]`? :)

npmccallum: https://github.com/RustCrypto/formats/blob/master/der/derive/src/newtype.rs#L78-L92

tarcieri: Oh right, I forgot it was adding `Deref`/`DerefMut` impls.

Thinking about that again, doing that unilaterally somewhat defeats the point of having a newtype, since it effectively leaks the inner type and precludes adding your own inherent methods (at least if you want it to be an idiomatic use of `Deref`).

But that's something we can circle back on another day, I suppose.

npmccallum: The intent of `newtype` is to expose the inner type, but allow the outer type to pick up implementations of traits relevant to specific contexts. Of note here is this new type: https://github.com/RustCrypto/formats/pull/479

This is particularly useful where an identical inner type is used in multiple contexts. A primary example of this is `Vec<Extension>` which has different OIDs in different contexts. The outer type determines which `Identifiable` instance to give to the inner type.

tarcieri: > The intent of newtype is to expose the inner type

This description is antithetical to the typical notion of "newtype", which is to encapsulate the inner type

npmccallum: @tarcieri I don't think the invariants of "newtype" are that well defined. They are commonly used to allow the implementation of a trait from one crate on the type from a second crate within a third crate, which is roughly how we're using them here.

tarcieri: Implicit deref coercion has somewhat magical and often surprising powers which allow you to transparently treat one type as another, which go well beyond what's typically meant by "newtype". It means you *can't* encapsulate the inner type, and it's a mandatory part of the public API.

A very common usage of the newtype pattern is to encapsulate another type to keep it out of the public API, so you can upgrade the dependency that provides the inner type without it resulting in a breaking change to the public API of the consumer providing the outer newtype.

PR #468, opened 2022-03-05
fix(der): manually impl Default for SetOfVec

npmccallum:


npmccallum:
FYI, this is needed because the derive conditionalizes `Default` on `T: Default` which is irrelevant in this case.

PR #472, opened 2022-03-07
feat: add ExtensionReq type

npmccallum:

### DIFF
--- a/x509/src/request.rs
+++ b/x509/src/request.rs
@@ -1,9 +1,14 @@
 //! PKCS#10 Certification Request types
 
+use crate::ext::Extension;
 use crate::{attr::Attributes, name::Name};
 
+use alloc::vec::Vec;
+
+use const_oid::db::rfc5912::ID_EXTENSION_REQ;
+use const_oid::{AssociatedOid, ObjectIdentifier};
 use der::asn1::BitString;
-use der::{Decodable, Enumerated, Sequence};
+use der::{Decodable, Enumerated, Newtype, Sequence};
 use spki::{AlgorithmIdentifier, SubjectPublicKeyInfo};
 
 /// Version identifier for certification request information.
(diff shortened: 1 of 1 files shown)


npmccallum:
@tarcieri This is an example of where my definition of `newtype` makes sense. In a `CertReqInfo` we have extension requests. These are just `Vec<Extension>`. However, `Vec<Extension>` should ONLY have an associated OID in this context since it is used in different contexts with different OIDs.

Even if we wanted to `impl AssociatedOid for Vec<Extension>`, we can't because `AssociatedOid` is defined in `const-oid` and `Vec` is defined in `alloc`; but we are making the association between the two in the `x509` crate. The `newtype` allows us to do this in a completely sane way.

tarcieri:
Using a newtype to permit trait impls for a type you can't control is idiomatic.

The issue is more with inherent methods than traits: if `Deref`/`DerefMut` are mandatory, you lose the ability to provide any encapsulation around the inner type. The inner type becomes a part of the public API and all of its methods are always available.

Types which impl `Deref`/`DerefMut` are more typically referred to "smart pointer" types than "newtypes", IMO.

npmccallum:
@tarcieri The Rust std library is inconsistent in this regard.

Types like `Rc` use `Deref` and then define associated methods on the wrapper type (for example: `Rc::into_raw()`). OTOH, types like `String` use `Deref` and also define inherent methods (for example: `String::push()`).

It seems to me that there are two types of `Deref` implementations: "smart pointers" and "superset of another type." The "smart pointers" examples don't add inherent methods since they are just supposed to be transparently the other type. The "superset" ones do.

tarcieri:
> It seems to me that there are two types of Deref implementations: "smart pointers" and "superset of another type."

I feel like the latter is [generally considered to be an anti-pattern](https://github.com/rust-unofficial/patterns/blob/main/anti_patterns/deref.md).

npmccallum:
@tarcieri I think there is a difference between that anti-pattern (i.e. trying to use `Deref` for a general inheritance semantic) and using `Deref` to link two types whose relationship is well-defined (like `String`/`str`).

In any case, I don't have any intention of adding inherent methods to any of the newtypes. My intent is solely to use them for implementations of traits, which we agree is legitimate.

tarcieri:
The problem with mandating the use of `Deref`/`DerefMut` as the only way to use the `Newtype` custom derive means you can't ever encapsulate the inner type.

That seems rather heavy handed and like something which should be handled on a type-by-type basis.

tarcieri:
From a "least surprise" perspective, `Deref`/`DerefMut` are already rather surprising, and having a custom derive that writes them makes it invisible to someone reading the code (although they are finally fairly visible in the rustdoc).

If it tripped me up, it's going to trip up others.

npmccallum:
@tarcieri Can we decide how to move forward on this specific question outside the context of this particular PR which doesn't depend on the `Deref` behavior?

tarcieri:
Yes, that's fine.

I'd still recommend adding at least an `AsRef<[T]>` impl to these types as a common protocol for accessing their members which doesn't rely on using deref coercion to access the inner type's `AsRef` impl, but that can be handled in a followup.

PR #473, opened 2022-03-07
feat(x509): re-export der crate

npmccallum:

### DIFF
--- a/x509/src/lib.rs
+++ b/x509/src/lib.rs
@@ -14,6 +14,8 @@ extern crate alloc;
 #[cfg(feature = "std")]
 extern crate std;
 
+pub use der;
+
 pub mod anchor;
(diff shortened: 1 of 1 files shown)


x509/src/lib.rs
```diff
@@ -14,6 +14,8 @@ extern crate alloc;
 #[cfg(feature = "std")]
 extern crate std;
 
+pub use ::der;
```
tarcieri, line 17: Unless it's actually ambiguous, I'd suggest just using:

```suggestion
pub use der;
```

...otherwise it renders like this in the rustdoc:

Screen Shot 2022-03-07 at 11 29 48 AM

npmccallum: Fixed.

PR #462, opened 2022-03-03
add support for CRLs, OCSP and certs-only SignedData messages.

carl-wallace:
- The ocsp crate is new and features encoders/decoders for the OCSP protocol.
- the pkcs7 crate now features a new module that provides a partial implementation of SignedData support. This is intended to provide parsers for certs-only PKCS7 messages in support of dynamic path building in the new certval library. This adds a new dependency on the x509 crate. ASN.1 definitions of structures that were not implemented are included as comments.
- The spki crate was modified because the signer_infos field in SignedData generated the need to add PartialOrd to AlgorithmIdentifier.
- The x509 crate received a new crl module that features encoders/decoders for CRLs. The Rust version in Cargo.toml was modified to match README and sibling modules. A couple of escape characters were added to keyusage.rs to quiet warnings from cargo doc. Typos were fixed in oids.rs and new changed to new_unwrap. Certificate tests were modified to feature new_unwrap. Only a simple CRL parsing test is included here. The certval library more thoroughly exercises CRLs via executing PKITS for validation and revocation status determination.

PR #475, opened 2022-03-07
feat(x509): add CRL type support

npmccallum:
This contains just the CRL portion of #462. I made sure to give @carl-wallace the author's credit.

One remaining question I have is why these are in the `crl` module. Are these types used in extensions? Or are they disconnected data types? If the latter, `crl` is probably the best place for them.

@tarcieri FYI


carl-wallace:
I don't know what you mean by "these". The definitions in crl.rs are associated with crl processing. I will close #462 and keep eye on this repo and align higher level bits as this catches up.

npmccallum:
@carl-wallace I mean the types in the `crl` module in this PR. Are the `crl` types used in extensions?

carl-wallace:
No. There are only three structures: the primary CRL structure, its TBS structure and a structure fabricated from the revoked certs field of the TBS structure (as noted in the comment, for some reason the spec authors defined this structure inline). Feel free to split into smaller pieces if you prefer. Credit should probably go to Alan Smithee:-)

npmccallum:
@carl-wallace Who would use these types and how are they used?

carl-wallace:
See the comments in the code. https://datatracker.ietf.org/doc/html/rfc5280#section-5.1

npmccallum:
@carl-wallace The standard defines the data types and what is hopes they accomplish from the perspective of 2008. It doesn't describe how these types are actually used in 2022. I'm hoping you can fill me in on the common usage patterns.

carl-wallace:
these are fundamental structures. where crls are used, these are in play.

tarcieri:
Thanks!

@carl-wallace hopefully this works for your use cases, or if not, feel free to open a PR to make adjustments

PR #476, opened 2022-03-07
feat(spki): impl PartialOrd for AlgorithmIdentifier

npmccallum:
@carl-wallace FYI

### DIFF
--- a/spki/src/algorithm.rs
+++ b/spki/src/algorithm.rs
@@ -14,7 +14,7 @@ use der::{Decodable, DecodeValue, Decoder, DerOrd, Encodable, Header, Sequence, 
 /// ```
 ///
 /// [RFC 5280 Section 4.1.1.2]: https://tools.ietf.org/html/rfc5280#section-4.1.1.2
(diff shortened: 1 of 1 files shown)


spki/src/algorithm.rs
```diff
@@ -14,7 +14,7 @@ use der::{Decodable, DecodeValue, Decoder, DerOrd, Encodable, Header, Sequence,
 /// ```
 ///
 /// [RFC 5280 Section 4.1.1.2]: https://tools.ietf.org/html/rfc5280#section-4.1.1.2
-#[derive(Copy, Clone, Debug, Eq, PartialEq)]
+#[derive(Copy, Clone, Debug, Eq, PartialEq, PartialOrd)]
```
tarcieri, line 17: Might as well add `Ord` at this point

carl-wallace: Most of the testing was with Ord. It was dropped while preparing the PR as not necessary to satisfy the requirements generated by the partial SignedData impl

npmccallum: I just added `Ord`.

PR #478, opened 2022-03-07
chore(x509): remove unhelpful doc strings

npmccallum:

### DIFF
--- a/x509/src/ext/pkix/name/dirstr.rs
+++ b/x509/src/ext/pkix/name/dirstr.rs
@@ -39,12 +39,11 @@ use der::Choice;
 ///
 /// [RFC 5280 Section 4.2.1.4]: https://datatracker.ietf.org/doc/html/rfc5280#section-4.2.1.4
 #[derive(Clone, Debug, Eq, PartialEq, Choice)]
+#[allow(missing_docs)]
 pub enum DirectoryString<'a> {
-    /// ia5String        IA5String      (SIZE (1..200))
     #[asn1(type = "PrintableString")]
     PrintableString(PrintableString<'a>),
 
-    /// utf8String       UTF8String     (SIZE (1..200))
     #[asn1(type = "UTF8String")]
     Utf8String(Utf8String<'a>),
 }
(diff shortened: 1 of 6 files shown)


tarcieri:
I can't say I'm terribly enthusiastic about the trend towards this crate becoming largely undocumented.

Some of the information in the comments seems helpful since it expresses invariants the custom derive doesn't yet support.

In other @RustCrypto crates we try to hoist up some interesting relevant information from the RFC into comments. Especially in a standard like X.509 which is laden with jargon, I think at least some of that would be helpful.

x509/src/ext/pkix/name/dirstr.rs
```diff
@@ -39,12 +39,11 @@ use der::Choice;
 ///
 /// [RFC 5280 Section 4.2.1.4]: https://datatracker.ietf.org/doc/html/rfc5280#section-4.2.1.4
 #[derive(Clone, Debug, Eq, PartialEq, Choice)]
+#[allow(missing_docs)]
 pub enum DirectoryString<'a> {
-    /// ia5String        IA5String      (SIZE (1..200))
```
tarcieri, line 43: Well that's just wrong but...

x509/src/ext/pkix/name/dirstr.rs
```diff
@@ -39,12 +39,11 @@ use der::Choice;
...
     #[asn1(type = "PrintableString")]
     PrintableString(PrintableString<'a>),
 
-    /// utf8String       UTF8String     (SIZE (1..200))
```
tarcieri, line 47: ...the size limitations seem like useful invariants which aren't currently captured by the proc macro.

npmccallum:
@tarcieri I agree completely about the invariants. I shouldn't have removed those.

However, I don't think the trend is towards reducing documentation at all. The trend is toward reducing **redundant** documentation. I'm 100% in favor of adding documentation that doesn't just dump an ASN.1 fragment and adds genuinely useful information.

Redundant documentation actually has negative value because it tends to be updated in one place but not the other.

npmccallum:
@tarcieri Actually, upon further review the size limitations appear to be incorrect. RFC 5280 doesn't have any limitations at all. This actually highlights my point about removing redundant documentation. In this case, we had documentation repeated twice: once on the struct and once on each field. But the latter documentation was actually incorrect and has now been removed.

carl-wallace:
Those should have been a la RFC5912.
```
  DirectoryString{INTEGER:maxSize} ::= CHOICE {
      teletexString    TeletexString(SIZE (1..maxSize)),
      printableString  PrintableString(SIZE (1..maxSize)),
      bmpString        BMPString(SIZE (1..maxSize)),
      universalString  UniversalString(SIZE (1..maxSize)),
      uTF8String       UTF8String(SIZE (1..maxSize))
  }
```

PR #479, opened 2022-03-07
feat(const-oid): add the AssociatedOid trait

npmccallum:

### DIFF
--- a/const-oid/src/lib.rs
+++ b/const-oid/src/lib.rs
@@ -29,6 +29,12 @@ pub use crate::{
 use crate::encoder::Encoder;
 use core::{fmt, str::FromStr};
 
+/// A trait which associates an OID with a type.
+pub trait AssociatedOid {
+    /// The OID associated with this type.
+    const OID: ObjectIdentifier;
+}
+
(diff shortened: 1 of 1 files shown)


const-oid/src/lib.rs
```diff
@@ -29,6 +29,12 @@ pub use crate::{
 use crate::encoder::Encoder;
 use core::{fmt, str::FromStr};
 
+/// Identifies a type with a given OID.
+pub trait Identified {
```
tarcieri, line 33: I can't think of a better name myself, but I think this trait would be clearer if it had `Oid` somewhere in the name.

npmccallum: Object Identifier Identified? :)

tarcieri: `Identified` is an extremely vague name for a trait

npmccallum: @tarcieri How about the name `AssociatedOid` which I just pushed?

tarcieri: Sounds good, thanks

PR #480, opened 2022-03-07
feat(der): clean up Newtype implementations

npmccallum:


tarcieri:
Thanks!

PR #481, opened 2022-03-07
feat(x509): add extension query methods to `TbsCertificate`

npmccallum:

### DIFF
--- a/x509/src/certificate.rs
+++ b/x509/src/certificate.rs
@@ -2,8 +2,9 @@ use crate::{name::Name, time::Validity};
 
 use alloc::vec::Vec;
 
+use const_oid::AssociatedOid;
 use der::asn1::{BitString, UIntBytes};
-use der::{Enumerated, Newtype, Sequence};
+use der::{Decodable, Enumerated, Error, ErrorKind, Newtype, Sequence};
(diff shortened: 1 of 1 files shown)


x509/src/certificate.rs
```diff
@@ -87,6 +88,41 @@ pub struct TbsCertificate<'a> {
...
+    ) -> Result<Option<(bool, T)>, Error> {
+        let mut iter = self.filter::<T>().peekable();
+        if let Some(item) = iter.next() {
+            return match iter.peek() {
+                Some(..) => Err(ErrorKind::Failed.into()),
+                None => Ok(Some(item?)),
+            };
+        }
+
+        Ok(None)
```
tarcieri, line 108: I think this would be marginally clearer without an early `return`:

```suggestion
        if let Some(item) = iter.next() {
            match iter.peek() {
                Some(..) => Err(ErrorKind::Failed.into()),
                None => Ok(Some(item?)),
            }
        } else {
            Ok(None)
        }
```

npmccallum: Fixed. Though I swapped out the branches for a match statement.

tarcieri (approved):

PR #486, opened 2022-03-09
fix(oiddb): don't filter out branch nodes

npmccallum:
Some of these branch nodes are actually used for types.


tarcieri:
Thanks!

PR #490, opened 2022-03-10
chore(x509): enable unused_lifetimes and unused_qualifications lints

npmccallum:
Also, clean up the types they warn about.


npmccallum:
@tarcieri As requested.

PR #492, opened 2022-03-10
feat(x509): impl TryFrom<Duration> for Validity

npmccallum:


tarcieri:
I'm not sure a `TryFrom` is the best way to model this, particularly one that has a side effect like checking the system clock. `TryFrom` seems better suited for deterministic type conversions.

How about an inherent method like `Validity::from_now(duration)`?

npmccallum:
@tarcieri Fixed.

PR #511, opened 2022-03-14
chore: rename x509 crate to x509-cert for publishing

bstrie:


tarcieri:
@bstrie looks like the fuzzer needs fixed. also rustfmt

bstrie:
@tarcieri All fixed.

PR #512, opened 2022-03-14
x509-cert v0.0.2

tarcieri:


bstrie:
That pkcs5 failure is spurious, right? But that raises the question of why pkcs5 is being built at all: it looks like it's told to run when `Cargo.*` is touched, and it's running because this PR touches Cargo.lock. Is that the desired behavior? For that matter, I'm not sure why Cargo.lock is checked into the repo at all here, is there a reason?

tarcieri:
I've opened an issue for the spurious build failure:

https://github.com/RustCrypto/actions/issues/11

> For that matter, I'm not sure why Cargo.lock is checked into the repo at all here when these are all just library crates, not binaries. Is there a reason I'm missing?

I've answered this questions so many times I really need to start maintaining a FAQ item for it. Unfortunately I'm unable to find the previous dozen answers I've given, so I guess I'll write it up again.

Generally across the project we check in Cargo.lock. This keeps the build reproducible.

Debugging cryptography is confusing, especially in cases where the only failure mode is a miscomputed result, and a lot of that confusion can be created by people not having the same versions of things they're trying to debug.

Here's an example of such an issue:

https://github.com/informalsystems/tendermint-rs/issues/967

In this particular case, a user managed to get a broken release of `sha2` which we caught and yanked after a few hours. It wound up in their Cargo.lock.

That meant that this user was experiencing a problem no one else could reproduce:

- Users who had old copies of the repo with an old Cargo.lock didn't experience the failure. If they ran `cargo update`, they got the version with the fix and still couldn't reproduce the issue
- Users who had cloned the repo fresh got the latest version

Debugging these sorts of issues can be very time consuming, particularly for maintainers who are trying to get downstream consumers of the crates to reproduce the issue.

Keeping `Cargo.lock` checked in keeps everyone on the same page.

We use Dependabot to keep dependencies in Cargo.lock up-to-date.

I know there's an orthodoxy that says "Check in Cargo.lock for binaries, don't check it in for libraries" but in our experience it is more helpful for debugging to keep the build reproducibly clean, and seeing failures manifest when particular dependencies are changed.

PR #516, opened 2022-03-17
feat(sec1): add default cargo feature for disabling point module

bstrie:
This allows users who aren't using EncodedPoint to omit dependencies on generic-array, typenum, and version_check.


tarcieri:
You'll need to enable the `point` feature inside of the `ssh-key` crate

bstrie:
@tarcieri Fixed. I find it quite disorienting that a `cargo test` in the top level directory would succeed while a `cargo check` in the ssh-key directory would fail.

PR #518, opened 2022-03-21
add certificate document support

carl-wallace:


tarcieri:
Re: the filename, we generally prefer `certificate/document.rs` over `certificate_document.rs` when there's already a `certificate.rs`.

If you fix that, adding something like this is OK for now to unblock your other work, but we'll probably circle back on removing the `der::Document` trait in favor of a generic type before the next release. See #427 (which I intend to finish up fairly soon).

carl-wallace:
Changed to certificate/document. Thanks.

PR #521, opened 2022-03-21
add OCSP encoders and decoders

carl-wallace:


ocsp/Cargo.toml
```diff
@@ -0,0 +1,30 @@
+[package]
+name = "ocsp"
```
tarcieri, line 2: The `ocsp` crate name is taken, however you can use this:
```suggestion
name = "x509-ocsp"
```

carl-wallace: Changed.

carl-wallace:
Sure, but possibly not until latter part of next week.

PR #531, opened 2022-03-25
crypto-serde

tarcieri:
Initial crate boilerplate and implementation.

Provides a `HexOrBin` serializer/deserializer type which encodes binary data with binary-oriented formats, or lower/upper hex when used with text-based formats.

Uses the `base16ct` crate for hex encoding.

Currently has a hard dependency on the `alloc` crate in order to keep the implementation simple.

Tested against the following serializer crates:

- `bincode`
- `ciborium`
- `json`
- `toml`


crypto-serde/src/lib.rs
```diff
@@ -0,0 +1,97 @@
...
+    html_favicon_url = "https://raw.githubusercontent.com/RustCrypto/meta/master/logo.svg"
+)]
+#![warn(
+    missing_docs,
+    rust_2018_idioms,
+    unused_lifetimes,
+    unused_qualifications
+)]
+
+extern crate alloc;
```
tarcieri, line 14: Hard dependency for now. It might be possible to make it optional.

crypto-serde/tests/toml.rs
```diff
@@ -0,0 +1,25 @@
...
+fn hex_lower() {
+    let serialized = toml::to_string(&HexLowerOrBin::from(EXAMPLE_BYTES)).unwrap();
+    assert_eq!(serialized, HEX_LOWER);
+}
+
+#[test]
+fn hex_upper() {
+    let serialized = toml::to_string(&HexUpperOrBin::from(EXAMPLE_BYTES)).unwrap();
+    assert_eq!(serialized, HEX_UPPER);
+}
```
tarcieri, line 25: TODO: deserializer tests. I'll add them before merging this PR.

tarcieri:
I'd like to add some `proptest`s but plan on doing that as a follow-up to this PR

PR #538, opened 2022-03-26
crypto-serde: Implement separate handling for arrays

daxpedda:
This introduces allocation-free de/serialization for fixed sized arrays. Along the way I also expanded integrations tests.

- [x] Implement separate `HexOrBin` for arrays.
- [x] Separate array and slices into modules to improve naming.
- [x] Abstract some repeating patterns.
- [x] Test array de/serialization for the remaining present formats.
- [ ] ~~Test `minicbor`.~~ (`minicbor` doesn't support `serde`)
- [ ] ~~Test `serde_cbor`?~~ (is archived)
- [x] Test `serde-json-core`.
- [ ] Figure out if we want to remove the const generic and pass a buffer instead for the array functions.


crypto-serde/src/lib.rs
```diff
@@ -17,47 +17,182 @@ extern crate alloc;
...
+    T: AsRef<[u8]>,
+{
+    #[cfg(feature = "alloc")]
+    if serializer.is_human_readable() {
+        return base16ct::upper::encode_string(value.as_ref()).serialize(serializer);
+    }
+    #[cfg(not(feature = "alloc"))]
+    if serializer.is_human_readable() {
+        return Err(S::Error::custom(
+            "serializer is human readable, which requires the `alloc` crate feature",
```
tarcieri, line 73: There's not much point providing an error message in this case. serde will just throw it away:

https://github.com/serde-rs/serde/blob/38edb47/serde/src/de/value.rs#L73-L81

Also it'd probably be good to extract the `base16ct::upper::encode_string(value.as_ref()).serialize(serializer)` bits into a function (e.g. `serialize_hex`) which is conditionally defined when alloc is available and returns an error when it isn't, rather than duplicating this logic.

tarcieri:
IMO this critically hinges on the CBOR serializations. If it degrades them there's not much point.

crypto-serde/src/lib.rs
daxpedda: That's just the default implementation, bincode actually does something:
https://docs.rs/bincode/1.3.3/src/bincode/error.rs.html#112-114

tarcieri: Looks like `bincode` is a std-only crate?

I guess it works if the format implementation has alloc enabled but `crypto-serde` (which is `alloc` on-by-default) has it disabled.

Maybe extract the error message into a constant then.

daxpedda: I already abstracted it into a single function, there should be no code duplication around this now. Tell me if what you think of it:
https://github.com/RustCrypto/formats/pull/538/commits/21de019aacd21af282e58b14e75284fcec26e40f

crypto-serde/src/slice.rs
```diff
@@ -0,0 +1,140 @@
...
+/// Deserialize the given slice from hex when using human-readable formats or
+/// binary if the format is binary.
+#[cfg(feature = "alloc")]
+#[cfg_attr(docsrs, doc(cfg(feature = "alloc")))]
+pub fn deserialize_hex_or_bin<'de, D>(deserializer: D) -> Result<Vec<u8>, D::Error>
+where
+    D: Deserializer<'de>,
+{
+    if deserializer.is_human_readable() {
+        base16ct::mixed::decode_vec(<&str>::deserialize(deserializer)?).map_err(D::Error::custom)
```
daxpedda, line 53: `<&str>::deserialize(deserializer)` this will blow up if a serialization format does something funny with strings and can't just borrow the bytes.

Explained in reverse: I believe this only works for serializations formats that serialize strings into bytes as is without modification.

To fix this, we would have to implement a `Visitor`.

tarcieri: I think this is a “cross that bridge when we get there” situation. While it’s a hypothetical, in absence of a major format implementation that actually does that I don’t think it’s worth concerning ourselves with.

daxpedda: Sounds good to me.

crypto-serde/src/array.rs
```diff
@@ -0,0 +1,172 @@
...
+/// [`HexOrBin`] serializer which uses lower case.
+pub type HexLowerOrBin<const N: usize> = HexOrBin<N, false>;
+
+/// [`HexOrBin`] serializer which uses upper case.
+pub type HexUpperOrBin<const N: usize> = HexOrBin<N, true>;
+
+/// Serializer/deserializer newtype which encodes bytes as either binary or hex.
+///
+/// Use hexadecimal with human-readable formats, or raw binary with binary formats.
+#[derive(Clone, Copy, Debug, Eq, Hash, Ord, PartialEq, PartialOrd)]
```
daxpedda, line 117: Is this appropriate or should I revert it?

daxpedda:
> - [ ] Figure out if we want to remove the const generic and pass a buffer instead for the array functions.

Basically I want to support `GenericArray`, to do that a simple change is required: `array::deserialize_hex_or_bin` needs to take an input buffer instead of returning an array. Alternatively we could introduce a separate function behind a crate feature that takes an `ArrayLength<u8>` generic parameter to support this (I believe that's a bit overkill).

crypto-serde/src/array.rs
tarcieri: It's fine. Potentially useful for declarative (de)serializers using `serde_derive`'s `try_from` and `into` attributes

crypto-serde/src/lib.rs
tarcieri: Looks good 👍

tarcieri:
> array::deserialize_hex_or_bin needs to take an input buffer instead of returning an array.

Yeah, that's what I was suggesting originally for the deserializers: have them accept a `&mut [u8]` as a parameter, rather than allocating and returning a value.

The array deserializers could ensure it's filled to completion. The slice deserializers could potentially fill as much as possible and reslice the input buffer as a `Result<&[u8]>` which is returned.

crypto-serde/src/array.rs
tarcieri: If you changed the array decoder functions to write into a provided `&mut [u8]`, this type could be used to handle allocating the array

daxpedda: Done.

daxpedda:
> The slice deserializers could potentially fill as much as possible and reslice the input buffer as a `Result<&[u8]>` which is returned.

I had that originally on the TODO list, but re-thinking this I realized that it doesn't really make sense because you can't pass any buffers into the `Deserialize::deserialize` method. I'm not sure what the use-case here would be. Am I missing something?

crypto-serde/src/slice.rs
daxpedda: Some more thoughts I had here:
- This can also fail with `DeserializeOwned`.
- To a degree this might be useful to make sure that the serialization format doesn't actually do something with the string serialization, as that would probably break constant-time assurances and circumvent the whole thing we are doing here.

tarcieri:
> you can't pass any buffers into the Deserialize::deserialize method.

The idea would be for downstream types to impl `Deserialize::deserialize` themselves, inside of which they can allocate an empty buffer, then pass the empty buffer along with the deserializer to a helper function defined in `crypto-serde`.

crypto-serde/src/slice.rs
tarcieri: > This can also fail with DeserializeOwned.

There's nothing magic about `DeserializeOwned`. It's just a marker trait w\ HRTB + a blanket impl which ensures that the `Deserialize` impl is valid for all lifetimes.

This means it's perfectly fine to *temporarily* borrow data within the deserializer, so long as it's only borrowed for the lifetime of the deserializer function and not referenced in the resulting value.

> To a degree this might be useful to make sure that the serialization format doesn't actually do something with the string serialization, as that would probably break constant-time assurances and circumvent the whole thing we are doing here.

I'm not sure what we can really do there. Once the string has been parsed by the input format during deserialization, or handed off to the serializer, it's really out of our hands. The most `crypto-serde` can do is minimize harm.

daxpedda:
Yes, but what would be the advantage of that?

On a side note, this would cause some code-duplication in the current implementation because `slice::HexOrBin` can't pre-allocate a `Vec` as it doesn't know the size.

tarcieri:
> Yes, but what would be the advantage of that?

It allows the use of any buffer type, not just a fixed-sized array.

You could pass in a `&mut Vec<u8>` preallocated to some maximum-known length, and a slice deserializer could return the actual length, at which point the `Vec` can be truncated to that length. That could even be built-in to the slice-based serializers via the `HexOrBin` newtypes.

Or it could allocate a `GenericArray` and pass it by reference instead of value. That avoids having to convert a returned array into a `GenericArray` by ensuring it's a `GenericArray` all along.

crypto-serde/src/slice.rs
daxpedda: > This means it's perfectly fine to _temporarily_ borrow data within the deserializer, so long as it's only borrowed for the lifetime of the deserializer function and not referenced in the resulting value.

This isn't the experience I had, see https://github.com/RustCrypto/crypto-bigint/pull/73#issuecomment-1079139655. But maybe this is a "bug" in bincode?

tarcieri: Yes, that has nothing to do with `DeserializeOwned`. As I said on that issue, `bincode` doesn't support borrowing a slice from the input.

Empirically the human readable formats we're currently testing against (`serde_json` and `toml`) do support this, so until we encounter a format that doesn't it's not an issue.

daxpedda: > Yes, that has nothing to do with `DeserializeOwned`. As I said on that issue, `bincode` doesn't support borrowing a slice from the input.

When you don't use `deserialize_from` but `deserialize` it works, I assumed it was because of `DeserializeOwned`. I didn't dig into it, so it being `DeserializeOwned` was just a guess.

> Empirically the human readable formats we're currently testing against (`serde_json` and `toml`) do support this, so until we encounter a format that doesn't it's not an issue.

Yeah I wasn't arguing for that, I just wanted to clear up left-over threads of conversation :smiley:.

tarcieri (approved):
Seems like a good start

PR #540, opened 2022-03-27
crypto-serde: implement slice deserializing with buffer

daxpedda:
Follow-up on #538. I hope this is what you meant.

### DIFF
--- a/crypto-serde/src/slice.rs
+++ b/crypto-serde/src/slice.rs
@@ -1,14 +1,13 @@
 //! TODO
 //! crypto-serde implementation for slices.
 
-use serde::{Serialize, Serializer};
+use core::fmt;
+
+use serde::de::{Error, Visitor};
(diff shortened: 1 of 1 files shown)


crypto-serde/src/slice.rs
```diff
@@ -41,6 +40,68 @@ where
     }
 }
 
+/// Deserialize the given slice from hex when using human-readable formats or
+/// binary if the format is binary.
+pub fn deserialize_hex_or_bin_with_memory<'de, D>(
```
tarcieri, line 45: In several of the other crates in this repo we use a `deserialize_*` vs `deserialize_*_vec` to distinguish the mutating vs allocating APIs. I'd recommend that here.

daxpedda: Done.

tarcieri (approved):

PR #545, opened 2022-03-27
crypto-serde: Improve documentation

daxpedda:
My English isn't the yellow of the egg, any feedback is welcome :laughing:.

### DIFF
--- a/crypto-serde/Cargo.toml
+++ b/crypto-serde/Cargo.toml
@@ -1,7 +1,7 @@
 [package]
 name = "crypto-serde"
 description = "Serde serializer/deserializer helpers used by the RustCrypto project"
-version = "0.0.0"
+version = "0.1.0"
 authors = ["RustCrypto Developers"]
 license = "Apache-2.0 OR MIT"
 repository = "https://github.com/RustCrypto/formats/tree/master/crypto-serde"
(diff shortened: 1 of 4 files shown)


crypto-serde/Cargo.toml
```diff
@@ -31,3 +31,8 @@ toml = "0.5"
 [features]
 default = ["alloc"]
 alloc = ["base16ct/alloc", "serde/alloc"]
+
+[package.metadata.docs.rs]
+all-features = true
+rustdoc-args = ["--cfg", "docsrs"]
+targets = []
```
daxpedda, line 38: I noticed that is something that RustCrypto isn't doing in other projects (I think). Should I remove it?
See:
- https://blog.rust-lang.org/2020/03/15/docs-rs-opt-into-fewer-targets.html
- https://docs.rs/about/metadata

crypto-serde/Cargo.toml
```diff
@@ -1,7 +1,7 @@
 [package]
 name = "crypto-serde"
 description = "Serde serializer/deserializer helpers used by the RustCrypto project"
-version = "0.0.0"
+version = "0.1.0"
```
tarcieri, line 4: Not ready to cut a release yet
```suggestion
version = "0.0.0"
```

crypto-serde/Cargo.toml
tarcieri: Looks good. We should probably do it everywhere we don't have target-dependent features.

PR #552, opened 2022-03-29
crypto-serde v0.1.0

tarcieri:
Initial release


tarcieri:
Going to rename this crate to `serdect` prior to a release

PR #553, opened 2022-03-29
Rename `crypto-serde` => `serdect`

tarcieri:
The crate provides constant-time(ish) `serde` serialization helpers, so this change better reflects the crate's purpose.


tarcieri:
cc @daxpedda

PR #571, opened 2022-04-02
der: `Document` and `SecretDocument` types

tarcieri:
Replaces the previous `Document` trait with types for wrapping serialized ASN.1 DER SEQUENCEs stored on the heap.

The `SecretDocument` is composed in terms of `Document` and provides some additional hardening for sensitive data, namely zeroize-on-drop and file permissions hardening when writing to disk.

This commit also removes all of the format-specific `*Document` types found in the `pkcs1`, `pkcs8`, `spki`, and `sec1` crates with the new `Document` and `SecretDocument` types.


tarcieri:
Removing WIP/draft.

The core implementation is effectively done, although could probably use some additional work to reduce repetitiveness. Still, this removes quite a bit of duplicated code from the original implementation.

I'm going to continue adding additional/replacement tests for the format-specific `Decode*`/`Encode*` traits.

cc @bstrie @npmccallum

x509/src/certificate/document.rs
```diff
@@ -1,97 +0,0 @@
-//! CertificateDocument implementation
```
tarcieri, line 1: @carl-wallace FYI, ended up removing this along with all of the format-specific `*Document` types. Hope the `Document` type works for your use cases

PR #616, opened 2022-04-25
spki: streaming fingerprint builder

tarcieri:
Uses the `Writer` trait to compute SPKI fingerprints by hashing output DER on-the-fly as opposed to encoding it first to an intermediate buffer then computing the SHA-256 digest.

Besides simply being more efficient, the buffered method had a hardcoded limit of 4096-bytes. The new implementation has no such limit because it hashes data as soon as it's encoded.

This additionally renames `SubjectPublicKeyInfo::fingerprint` to `::fingerprint_bytes`. "True" SPKI fingerprints are Base64-encoded, and this commit also promotes `fingerprint_base64` to being earlier in the impl block so it's the first fingerprinting method people see.

### DIFF
--- a/spki/src/fingerprint.rs
+++ b/spki/src/fingerprint.rs
@@ -1,0 +1,43 @@
+//! SPKI fingerprint support.
+
+use der::Writer;
+use sha2::{Digest, Sha256};
+
+/// Size of a SHA-256 SPKI fingerprint in bytes.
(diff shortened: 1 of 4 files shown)


tarcieri:
It might be interesting to make `fingerprint::Builder` generic and integrate it directly into the `der` crate.

This works as a start, though.

PR #624, opened 2022-04-27
Add support for parsing keys out of the ssh known_hosts file format

stormbrew:
This is largely based on the authorized_keys parse functionality already present, and there are probably some opportunities to remove redundancy between them, but this is a functional start.

Note that also in this PR is making it so that it treats the comments field of an ssh key as optional. This is correct both according to the ssh(8) man page and the comments all over the codebase here, but it was using a lack of comment as an indication that the key was truncated. This does mean that a validity check has been removed, but this was probably always the wrong place to do it (the key parser itself should know the expected length of the key).


tarcieri:
Looks good, thank you!

And good catch on absent comments. It seems like there might still be a few rough edges there elsewhere in the codebase. I'll have to look through it for any other potential problems.

PR #625, opened 2022-04-27
der: remove decode-time SET OF ordering checks

tarcieri:
Some DER serializer implementations fail to properly sort elements of a `SET OF`. This is technically non-canonical, but occurs frequently enough that most DER decoders tolerate it. Unfortunately because of that, we must also follow suit.

This commit builds an `ArrayVec` or `Vec` of the incoming decoded set elements, and then unconditionally sorts it retroactively, ensuring all `SET OF` types internally reflect the proper order, but out-of-order `SET OF` messages are tolerated when coming off-the-wire.

The sort implementation is merge sort, which should be efficient for already-in-order or mostly-in-order sets.

### DIFF
--- a/der/src/arrayvec.rs
+++ b/der/src/arrayvec.rs
@@ -85,6 +85,18 @@ impl<T, const N: usize> ArrayVec<T, N> {
     }
 }
 
+impl<T, const N: usize> AsRef<[Option<T>]> for ArrayVec<T, N> {
+    fn as_ref(&self) -> &[Option<T>] {
+        &self.elements[..self.length]
+    }
+}
+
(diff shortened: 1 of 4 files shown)


tarcieri:
@carl-wallace it's now failing due to this test which checks for the old behavior:

https://github.com/RustCrypto/formats/blame/master/x509/tests/name.rs#L129-L133

I think it can be safely removed if we do wish to tolerate out-of-order sets.

carl-wallace:
Removed (or inverted). Thanks for making that change.

tarcieri:
Ok, I inverted the check it was performing.

Perhaps I should vendor it into the `der` crate as it doesn't currently have a test for that.

Issue #525, opened 2022-03-25
der: `Decoder` and `Encoder` traits


tarcieri:
In the `ssh-key` crate I've used a similar (but simplified) trait structure inspired by the `der` crate. It also has `Decode`/`Encode` traits, but additionally provides traits for `Decoder`/`Encoder`:

- https://github.com/RustCrypto/formats/blob/d42432a3/ssh-key/src/decoder.rs
- https://github.com/RustCrypto/formats/blob/d42432a3/ssh-key/src/encoder.rs

These traits are impl'd for types like `pem_rfc7468::{Decoder, Encoder}`, allowing it to decode from/encode to PEM directly without an intermediate step where data is first decoded from/encoded to a `Vec<u8>`. This makes it possible to use PEM encoding in heapless environments, which currently isn't possible with the `der` crate.

PEM decoding is a bit tricky: since the original buffer can't be referenced (since it's encoded as Base64) decoding only works for fully owned types, which happens to be what the `ssh-key` provides. We don't currently have a way of bounding on such types (though that's trivial to add).

If we were to add this, type signatures would change as follows:

- `fn decode(decoder: &mut Decoder<'a>) -> Result<T>` => `fn decode(decoder: &mut impl Decoder<'a>) -> Result<T>`
- `fn encode(&self, encoder: &mut Encoder<'_>) -> Result<()>` => `fn encode(&self, encoder: &mut impl Encoder<'_>) -> Result<()>`

We'd also need to come up with new names for the current `struct Decoder` / `struct Encoder`, possibly something like `SliceDecoder` and `SliceEncoder`.

tarcieri:
The description suggests using `&mut impl Reader<'a>` and `&mut impl Writer` but there's a limitation of using impl trait in the argument position: when using a method of a generic type, you presently can't specify generic type arguments as well.

So they should probably be generic types instead.

tarcieri:
With #635 and #636 this is fully implemented

PR #626, opened 2022-04-29
der+x509: replace `Newtype` custom derive macro with macro_rules!

tarcieri:
In attempting to implement the `Decode`/`DecodeValue` changes for #525, I ran into trouble trying to modify the `Newtype` derive macro. In its present form it both does quite a bit which makes it difficult to test the lowerings. As the lowerings aren't well-tested, it becomes extremely difficult to debug when the proc macro is emitting bad tokens.

In looking at those problems, I came to the conclusion that a `macro_rules!` macro would be better for several reasons:

- Easier to modify and debug
- There is no corresponding `Newtype` trait, which breaks with the spirit of custom derive macros
- The `Newtype` macro was stealthily defining several macro as the same time, which makes it somewhat inscrutible (especially with the absence of a same-named trait which bounds on the traits it impls).
- At least within the context of this repo, this is a pattern specific to the `x509` crate.

If we want to go the custom derive route, it would be better to break it down into separate trait-specific macros for e.g `DecodeValue`/`EncodeValue` which are easier to test/debug and better map to the custom derive pattern.

cc @npmccallum


tarcieri:
Going to go ahead and merge this to unblock #525.

I think we can potentially revisit doing some of this with proc macros, but if we do, I would like to break them down into separate custom derive macros for each trait, such as `DecodeValue`, `EncodeValue`, and `FixedTag`.

PR #631, opened 2022-05-02
Use crates.io info badges from buildstats.info in README.mds

tarcieri:
These badges include download figures, which eliminates the need to have a separate badge that measures downloads.

### DIFF
--- a/base16ct/README.md
+++ b/base16ct/README.md
@@ -42,7 +42,7 @@ dual licensed as above, without any additional terms or conditions.
 
 [//]: # (badges)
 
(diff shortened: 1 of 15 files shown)


tarcieri:
@franziskuskiefer I left `tls_codec` out of this PR but perhaps at some point we should try to synchronize on a common style for README.md badges

franziskuskiefer:
👍🏻  yes we should synchronise the style. I haven't had time for the crate lately but I'll do some work on it soonish and make a proposal.

PR #664, opened 2022-05-21
export newtype macro

carl-wallace:


x509-cert/src/macros.rs
```diff
@@ -1,5 +1,7 @@
 //! Macros used by this crate
 
+/// macro_rules implementation of former newtype proc macro. See `<https://github.com/RustCrypto/formats/pull/626>`.
```
tarcieri, line 3: I think it might be helpful to describe what the macro actually does.

```suggestion
/// Implements the following traits for a newtype of a `der` decodable/encodable type:
///
/// - `From` conversions to/from the inner type
/// - `AsRef` and `AsMut`
/// - `DecodeValue` and `EncodeValue`
/// - `FixedTag` mapping to the inner value's `FixedTag::TAG`
///
/// The main case is simplifying newtypes which need an `AssociatedOid`
```

x509-cert/src/lib.rs
```diff
@@ -20,7 +20,7 @@ extern crate alloc;
 extern crate std;
 
 #[macro_use]
-mod macros;
+pub mod macros;
```
tarcieri, line 23: It's not necessary (or desirable) to make this `pub`

```suggestion
mod macros;
```

x509-cert/src/macros.rs
carl-wallace: Changed.

x509-cert/src/lib.rs
carl-wallace: Changed.

PR #666, opened 2022-05-28
base32ct: constant-time Base32 decoder/encoder

tarcieri:
Rounding out the collection of RFC4648 implementations in this repo, this adds a constant-time Base32 decoder/encoder implementation.

Adapted from:

https://github.com/paragonie/constant_time_encoding/blob/master/src/Base32.php


tarcieri:
Note: I've chosen a structure more like `base64ct` both because I'd like to handle both padded and unpadded variants and also multiple alphabets (upper and lower case variants of both the RFC4648 and the Crockford alphabet)

PR #678, opened 2022-06-20
x509-cert: bump `rstest` to v0.14

tarcieri:


tarcieri:
No idea what's going on here

PR #680, opened 2022-06-27
base64ct: fix last block validation

### LINKED ISSUE #679: base64ct doesn't handle invalid last characters
I ran into this issue trying to use base64ct in arti, a Tor implementation in Rust. https://gitlab.torproject.org/tpo/core/arti/-/issues/154

This simple crate demonstrates the issue: https://github.com/pastly/base64ct-poc. From its README:

The code uses two short base64-encoded strings to demonstrate the discrepancy
in how base64ct handles the last character. Base64 is chosen as the point of
comparison because it handles *this* issue correctly IMO.

The two strings:

- Mi, which is `001100 100010`, or split "on the byte," `00110010 0010`.

- Mg, which is `001100 100000`, or split "on the byte," `00110010 0000`.

The former contains trailing bits that are non-zero, thus shouldn't be ignored,
but are ignored by base64ct.

The latter has only zero-valued trailing bits, so the bits are safely ignored
by both base64 and base64ct.

<https://en.wikipedia.org/wiki/Base64#Base64_table>

tarcieri:
Generalizes the internal `validate_padding` function into a `validate_last_block` function which can work on unpadded variants as well.

The function performs round-trip re-encoding the decoded last block, ensuring that it matches the provided input, or otherwise returns an `InvalidEncodingError`.

Closes #679

### DIFF
--- a/base64ct/src/encoding.rs
+++ b/base64ct/src/encoding.rs
@@ -96,7 +96,7 @@ impl<T: Alphabet> Encoding for T {
         dst_rem.copy_from_slice(&tmp_out[..dst_rem.len()]);
 
         if err == 0 {
-            validate_padding::<T>(src.as_ref(), dst)?;
+            validate_last_block::<T>(src.as_ref(), dst)?;
             Ok(dst)
         } else {
             Err(Error::InvalidEncoding)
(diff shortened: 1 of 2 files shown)


arturomf94:
Ah - sweet! I was about to open a PR with [these changes](https://github.com/arturomf94/formats/commit/e34c679b870fdf887832a6fea0f247f2b25aeb51)! Thanks so much for this @tarcieri!

PR #684, opened 2022-07-01
Fix len_len for TlsByteVecU{16,32}

CarsonHoffman:
Currently, `TlsByteVecU16::len_len` and `TleByteVecU32::len_len` both return 1, while they should return 2 and 4, respectively. (`len_len` represents the number of bytes taken by the prefixed vector length, and the `U16` and `U32` vectors, as their names imply, have their lengths take up 16 and 32 bits.) This is due to an issue in the implementation of `impl_tls_vec`, where the call to `impl_vec_members` passed in a static `1`, rather than the correct parameter of `$len_len`. This PR changes the implementation to pass through the correct length.

For reference, the following program:
```rust
use tls_codec::*;

fn main() {
    println!(
        "{} {} {}",
        TlsByteVecU8::len_len(),
        TlsByteVecU16::len_len(),
        TlsByteVecU32::len_len(),
    );
}
```
…gives `1 1 1` before the change, and `1 2 4` after it.

This never had an impact on internal serialization since it uses the `Size` trait which is correctly implemented by the `impl_byte_size` macro.

### DIFF
--- a/tls_codec/src/tls_vec.rs
+++ b/tls_codec/src/tls_vec.rs
@@ -564,7 +564,7 @@ macro_rules! impl_tls_vec {
         }
 
         impl $name {
-            impl_vec_members!(u8, 1);
+            impl_vec_members!(u8, $len_len);
         }
 
         impl core::hash::Hash for $name {


franziskuskiefer (changes requested):
ugh, thanks for fixing this.
Can you add a quick test to make sure it doesn't break again.

franziskuskiefer:
@CarsonHoffman are you still working on this?

PR #691, opened 2022-07-29
der: Rudimentary implementation of TeletexString and VideotexString

lumag:
Add basic support for TeletexString and VideotexString tags. For now treat them as if they were IA5Strings, disallowing chars >= 0x80.


der/src/asn1/videotex_string.rs
```diff
@@ -0,0 +1,154 @@
...
+
+impl<'a> VideotexStringRef<'a> {
+    /// Create a new ASN.1 `VideotexString`.
+    pub fn new<T>(input: &'a T) -> Result<Self>
+    where
+        T: AsRef<[u8]> + ?Sized,
+    {
+        let input = input.as_ref();
+
+        // Validate all characters are within IA5String's allowed set
```
tarcieri, line 37:
```suggestion
        // Validate all characters are within VideotexString's allowed set
```

tarcieri:
Between all of the similar-but-not-identical string types this is starting to feel like quite a bit of duplication.

Curious if this could be wrapped up in `macro_rules!` at the very least, but I'm fine with merging this PR as-is (at least, once it's been rustfmt'd)

der/src/asn1/videotex_string.rs
lumag: Done

lumag:
Updated to pass rustfmt.

carl-wallace:
I pulled and ran a path validator against a set of busted certs that were discussed a few months back (some of which had teletex). Those now parse and validate. I don't recall why we didn't pursue this approach (other than maybe that teletex is more or less deprecated by 5280).

lumag:
@carl-wallace as you've mentioned it. I couldn't find an existing pki/path validation library/code using x509-cert. Could you hopefully point me to it?

carl-wallace:
Bear in mind it’s not “done” and I’ve not worked with it other than quick tests like this for a few months, but I have thrown a variety of certs collected from the wild at it. If you find it useful, I’d welcome additional hands to ready for release. It’s here: https://github.com/carl-wallace/rust-pki. Note, I have some unpushed changes to add p384 support once that is released.

 

From: Dmitry Baryshkov ***@***.***>
Reply-To: RustCrypto/formats ***@***.***>
Date: Friday, July 29, 2022 at 6:32 PM
To: RustCrypto/formats ***@***.***>
Cc: Carl Wallace ***@***.***>, Mention ***@***.***>
Subject: Re: [RustCrypto/formats] der: Rudimentary implementation of TeletexString and VideotexString (PR #691)

 

@carl-wallace as you've mentioned it. I couldn't find an existing pki/path validation library/code using x509-cert. Could you hopefully point me to it?

PR #692, opened 2022-07-29
x509-cert: support `TeletexString`

lumag:
Enable using TeletexString in several places where it should be enabled.


tarcieri:
Thanks!

Issue #693, opened 2022-07-30
der: string support deficiencies


lumag:
Taking this out from #690 , #691 and few practical thoughts from my side:

The following string tags are unsupported:

- [x] BMPString
- [ ] UniversalString
- [ ] VisibleString
- [ ] GraphicString

Additionally:
- [ ] TeletexString lacks support for `chars >= 0x80`

Additionally current API lacks easy to use unified string API (at least for AnyRef). Currently when parsing the string from AnyRef (e.g. in RDN) one has to manually match all the string tags and to use corresponding `foo_string()` calls. Instead it might be useful to add the `is_string() -> bool` and to support  generic string wrapper around `::printable_string()`, `::utf8_string()`, `::ia5_string()`, etc.

tarcieri:
> support  generic string wrapper around `::printable_string()`, `::utf8_string()`, `::ia5_string()`, etc.

Yes, I mentioned as much in #692. I definitely want to address this before cutting another release of `der` as there is currently massive amount of duplication and copy-pasted code among the various string types.

There's an internal [`StrSlice`](https://github.com/RustCrypto/formats/blob/master/der/src/str_slice.rs) type it might make sense to expose publicly as a common representation of a string. It's a newtype wrapper for `&str` with an associated [`Length`].

With this as a common internal representation, a lot of the duplication between e.g. `Ia5StringRef`, `TeletexStringRef`, and `VideotexStringRef` could be eliminated, possibly via `Deref` coercions to replace the existing (same-shaped) inherent methods with a common undelrying type, with the core differences between the various types being largely relegated to the constructor.

It would be good to figure all of this out before trying to add any additional string types.

lumag:
@tarcieri the major issue with the `StrSlice` that I faced while trying to implement `TeletexString` in a more correct way (handling `byte >= 0x80` as if it was the win-1252), is that it can not be easily updated to support different `as_bytes()` implementation (or I could not come up with the way to achieve this). So all the strings currently work because either they are ASCII strings or Utf8Strings. For BMPString we'd need UTF-16BE encoding.

tarcieri:
@lumag yeah, `StrSlice` could only work for cases that are subsets of ASCII/UTF-8, but I think it would still make sense to use it for those

lumag:
Being a minor party here, I'm fine with either approach as long as StrSlice doesn't become the public API.

tarcieri:
If it were the `Deref` target for ASCII-like encodings it would have to be

[...]

PR #697, opened 2022-08-04
der/asn1: remove bolerplate code by using Deref (#693)

lumag:
Use Deref<StrSlice> to limit amount of boilerplace code in the
*StringRef implementations.

Signed-off-by: Dmitry Baryshkov <dmitry.baryshkov@linaro.org>

### DIFF
--- a/der/src/asn1/ia5_string.rs
+++ b/der/src/asn1/ia5_string.rs
@@ -4,7 +4,7 @@ use crate::{
     asn1::AnyRef, ord::OrdIsValueOrd, ByteSlice, DecodeValue, EncodeValue, Error, FixedTag, Header,
     Length, Reader, Result, StrSlice, Tag, Writer,
 };
-use core::{fmt, str};
+use core::{fmt, ops::Deref, str};
 
 /// ASN.1 `IA5String` type.
 ///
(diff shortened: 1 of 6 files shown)


der/src/asn1/ia5_string.rs
```diff
@@ -41,31 +41,13 @@ impl<'a> Ia5StringRef<'a> {
...
-        self.inner.len()
-    }
-
-    /// Is the inner string empty?
-    pub fn is_empty(&self) -> bool {
-        self.inner.is_empty()
-    }
 }
 
-impl AsRef<str> for Ia5StringRef<'_> {
```
tarcieri, line 66: This is a breaking change. I think it's also generally still good to have an `AsRef` impl for each of these.

tarcieri:
One nit about removing the `AsRef` impls. Otherwise this looks good.

der/src/asn1/ia5_string.rs
lumag: Ack, I can revert this change. However, I thought that Deref + AsRef<str> implementation for StrSplice should be enough. Is it not?

lumag:
@tarcieri done, added `AsRef`

der/src/asn1/ia5_string.rs
tarcieri: No, deref coercion doesn't apply to trait impls, since the trait needs to be impl'd for the parent type

tarcieri (approved):
Looks good, thanks!

PR #698, opened 2022-08-15
pkcs1: implement support for PSS Params structure

lumag:
Add support for parsing and generatign RSA PSS params.

Signed-off-by: Dmitry Baryshkov <dmitry.baryshkov@linaro.org>

### DIFF
--- a/pkcs1/src/params.rs
+++ b/pkcs1/src/params.rs
@@ -1,0 +1,103 @@
+//! PKCS#1 RSA parameters
+
+use crate::{Error, Result};
+use der::asn1::{AnyRef, ObjectIdentifier};
+use der::{Decode, Enumerated, Sequence, Tag};
+use spki::AlgorithmIdentifier;
+
+/// `TrailerField` as defined in [RFC 8017 Appendix 2.3].
+/// ```text
+/// TrailerField ::= INTEGER { trailerFieldBC(1) }
+/// ```
+/// [RFC 8017 Appendix 2.3]: https://datatracker.ietf.org/doc/html/rfc8017#appendix-A.2.3
+#[derive(Clone, Debug, Copy, PartialEq, Eq, Enumerated)]
+#[asn1(type = "INTEGER")]
+#[repr(u8)]
+pub enum TrailerField {
+    /// the only supported value (0xbc, default)
+    BC = 1,
+}
+
+impl Default for TrailerField {
+    fn default() -> Self {
+        Self::BC
+    }
+}
+
+const OID_SHA_1: ObjectIdentifier = ObjectIdentifier::new_unwrap("1.3.14.3.2.26");
+const OID_MGF_1: ObjectIdentifier = ObjectIdentifier::new_unwrap("1.2.840.113549.1.1.8");
+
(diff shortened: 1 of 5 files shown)


pkcs1/src/params.rs
```diff
@@ -0,0 +1,187 @@
...
+const SHA_1_AI: AlgorithmIdentifier<'_> = AlgorithmIdentifier {
+    oid: OID_SHA_1,
+    parameters: None,
+};
+
+fn default_mgf1_sha1() -> AlgorithmIdentifier<'static> {
+    AlgorithmIdentifier {
+        oid: OID_MGF_1,
+        parameters: Some(
+            AnyRef::new(Tag::Sequence, &[0x06, 0x5, 0x2b, 0xe, 0x3, 0x2, 0x1a]).unwrap(),
```
tarcieri, line 46:
```suggestion
            AnyRef::from(&OID_SHA_1)
```

lumag: Not quite. This will end up with the Any containing just `OID` rather than `SEQUENCE` with the single `OID` inside

lumag:
```
   0  48: SEQUENCE {
  2  13:   [0] {
  4  11:     SEQUENCE {
  6   9:       OBJECT IDENTIFIER sha-256 (2 16 840 1 101 3 4 2 1)
       :       }
       :     }
 17  26:   [1] {
 19  24:     SEQUENCE {
 21   9:       OBJECT IDENTIFIER pkcs1-MGF (1 2 840 113549 1 1 8)
 32  11:       SEQUENCE {
 34   9:         OBJECT IDENTIFIER sha-256 (2 16 840 1 101 3 4 2 1)
       :         }
       :       }
       :     }
 45   3:   [2] {
 47   1:     INTEGER 32
       :     }
       :   }
```

tarcieri: Aah. Annoying

lumag:
@tarcieri simplified the implementation to use `derive[Sequence]`. However I still could not simplify the `AnyRef` part.

pkcs1/src/params.rs
```diff
@@ -0,0 +1,103 @@
...
+///     saltLength         [2] INTEGER            DEFAULT 20,
+///     trailerField       [3] TrailerField       DEFAULT trailerFieldBC
+/// }
+/// HashAlgorithm ::= AlgorithmIdentifier
+/// MaskGenAlgorithm ::= AlgorithmIdentifier
+/// ```
+///
+/// [RFC 8017 Appendix 2.3]: https://datatracker.ietf.org/doc/html/rfc8017#appendix-A.2.3
+#[derive(Clone, Debug, Eq, PartialEq, Sequence)]
+pub struct RsaPSSParameters<'a> {
```
tarcieri, line 71: Per [RFC 430 naming conventions](https://rust-lang.github.io/api-guidelines/naming.html), this should be:

```suggestion
pub struct RsaPssParameters<'a> {
```

tarcieri (changes requested):
One nit otherwise looks good

tarcieri (approved):
I'll go ahead and fix up the name myself

PR #701, opened 2022-08-19
x509-cert: re-export spki

tomleavy:
SubjectPublicKeyInfo is required in order to create a CertReqInfo and was not currently exported


tarcieri:
Thanks!

PR #704, opened 2022-08-22
base64ct: fix io::Read impl

newpavlov:
Users of the trait expect that implementations should return `Ok(0)` to signal end of stream.

The issue was raised in the [Zulip chat](https://rustcrypto.zulipchat.com/#narrow/stream/300570-formats/topic/base64ct.20decoder).

### DIFF
--- a/base64ct/src/lib.rs
+++ b/base64ct/src/lib.rs
@@ -3,7 +3,7 @@
 #![doc(
     html_logo_url = "https://raw.githubusercontent.com/RustCrypto/meta/master/logo.svg",
     html_favicon_url = "https://raw.githubusercontent.com/RustCrypto/meta/master/logo.svg",
-    html_root_url = "https://docs.rs/base64ct/1.4.1"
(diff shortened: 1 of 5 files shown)


tarcieri (approved):

base64ct/src/lib.rs
```diff
@@ -3,7 +3,7 @@
 #![doc(
     html_logo_url = "https://raw.githubusercontent.com/RustCrypto/meta/master/logo.svg",
     html_favicon_url = "https://raw.githubusercontent.com/RustCrypto/meta/master/logo.svg",
-    html_root_url = "https://docs.rs/base64ct/1.4.1"
```
newpavlov, line 6: @tarcieri 
Considering that you forgot to bump it, I think it's worth to remove uses of `html_root_url` from all of our crates. :)

tarcieri: Yeah I’ve been slowly doing that as I cut new releases. Guess I didn’t do this one yet.

PR #709, opened 2022-08-28
x509-cert: bump `rstest` to v0.15

tarcieri:
Still confused why #682 isn't passing


tarcieri:
Per #715 we'd need to bump MSRV for this: to 1.59 for v0.13, and 1.60 for v0.14+.

PR #716, opened 2022-09-03
chore(spki): don't enable base64ct implicitly from alloc feature

npmccallum:


tarcieri:
Note: MSRV bumps like this require we bump minor versions as well, per our SemVer policy.

We’re probably at the point we can start a new release cycle but it’d be good to get any lingering non-breaking changes first.

Namespaced features in particular make these MSRV bumps a bit painful as they have to be done at the granularity of an entire workspace, and they prevent whole releases from being parsed by Cargo, and therefore you don’t have `rust-version` to inform the process.

npmccallum:
@tarcieri Yeah, no worries. I just ran across this problem today and figured I'd throw a patch across the wall. Then it required an MSRV bump, so I realized it was more than trivial.

tarcieri:
We can merge this whenever we're ready to begin making breaking changes

tarcieri:
@npmccallum if you rebase, this can be merged

tarcieri:
Rebased in #787

PR #723, opened 2022-09-16
Implement ValueOrd in more places

roblabla:
I'm currently working on implementing [Authenticode](https://www.symbolcrash.com/wp-content/uploads/2019/02/Authenticode_PE-1.pdf) (microsoft's codesigning system). As part of this, I've had to implement a lot more of PKCS#7, and a problem I've been hitting fairly consistently is how few types implement ValueOrd.

This MR does three things:

1. It adds manual ValueOrd implementations for der's FlagSet, UIntRef, and x509-cert's Version.
2. It allows deriving ValueOrd for Choice enums (note - I'm not 100% sure the implementation is correct, more on this below).
3. It derives ValueOrd on a lot of different types.

---

The biggest improvement here is implementing ValueOrd for Choice enums. It works by generating an implementation that looks roughly like this:

```rust
match (self, other) {
    (Self::A(this), Other::A(other)) => this.value_cmp(other),
    (Self::B(this), Other::B(other)) => this.value_cmp(other),
    _ => unreachable!()
}
```

The idea being that ValueOrd::value_cmp should only be called for enums that share a variants. Enums that don't share a variant should have a different tag, and thus `DerOrd` won't call the `value_cmp` function.

### DIFF
--- a/der/derive/src/value_ord.rs
+++ b/der/derive/src/value_ord.rs
@@ -21,6 +21,8 @@ pub(crate) struct DeriveValueOrd {
 
     /// Fields of structs or enum variants.
     fields: Vec<ValueField>,
+
+    is_enum: bool,
 }
 
 impl DeriveValueOrd {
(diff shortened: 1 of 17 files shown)


tarcieri (approved):
Thanks! This fills a major TODO / API gap.

I only have one very minor nit I'll leave as a line note.

der/derive/src/value_ord.rs
```diff
@@ -21,6 +21,8 @@ pub(crate) struct DeriveValueOrd {
 
     /// Fields of structs or enum variants.
     fields: Vec<ValueField>,
+
+    is_enum: bool,
```
tarcieri, line 25: I wonder if this would be semantically clearer as an `enum`

PR #726, opened 2022-09-20
feat(base64ct): add support for output streaming

npmccallum:
There are two reasons you might want to reject this PR.

1. Unlike the rest of the crate, output streaming is not secret-safe. I have documented this. But it still may lead to a mismatch of expectations.

2. Unlike the rest of the crate, encoded output is streamed as `&[u8]` rather than `&str`. This allows for the `Update` trait to be reused for both encoding and decoding. It also allows runtime enablement of encoding without changing the types. This is required to implement the JWS `b64` parameter. For example, you can do: `Either<T, Encoder<T>>` and the result still implements `Update`.

### DIFF
--- a/base64ct/src/lib.rs
+++ b/base64ct/src/lib.rs
@@ -75,6 +75,8 @@ extern crate alloc;
 #[cfg(feature = "std")]
 extern crate std;
 
+pub mod stream;
+
 mod alphabet;
 mod decoder;
 mod encoder;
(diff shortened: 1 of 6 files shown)


tarcieri:
The first point is a good one. `base64ct` is aimed at use cases and formats that may potentially contain secrets, so something like this doesn't really fit the spirit of the crate.

PR #728, opened 2022-09-23
Update Cargo.lock

newpavlov:
Also bumps regex to 1.6 to prevent "maybe insecure" warning on deps.rs.


newpavlov:
@tarcieri 
`time v0.3.14` has bumped MSRV to 1.59, which is bigger than `der`'s MSRV (1.57).

PR #733, opened 2022-10-10
pkcs1: implement support for OAEP Params structure

lumag:
Add support for parsing and generatign RSA OAEP params.

Note: I was not sure regarding the `pSpecifiedEmpty.parameters`, but RFC3560 clearly defines it as `OCTET STRING (SIZE(0))`.


lumag:
BTW: it would be nice to cut a point release of `pkcs1` crate after this PR is merged to be able to use Params structs.

tarcieri (approved):
Looks good, thanks!

PR #738, opened 2022-10-13
Add sha2-style base64 crypt format

ibotty:
This format is also known as Hash64 by python's passlib.

It is used e.g. by the `crypt(3)` schemes sha1_crypt, sha256_crypt, sha512_crypt and md5_crypt, but notably not `des_crypt`.

This also refactors Alphabet into two traits.
The translation part stays in `Alphabet`, the en- and decoding part get's its own trait `AlphabetEncoding`.

The default Encoding is the usual big endian encoding. To opt in to little endian, an `Alphabet` struct has to (blankly) implement `LittleEndianAlphabetEncoding` instead of `AlphabetEncoding`.

### DIFF
--- a/base64ct/src/alphabet.rs
+++ b/base64ct/src/alphabet.rs
@@ -7,6 +7,7 @@ use core::{fmt::Debug, ops::RangeInclusive};
 
 pub mod bcrypt;
 pub mod crypt;
+pub mod shacrypt;
 pub mod standard;
 pub mod url;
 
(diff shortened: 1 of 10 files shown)


tarcieri:
This seems like a fairly invasive change just for the purposes of supporting `sha-crypt`, which is ostensibly the only thing that works this way.

Perhaps the logic could be put there instead?

tarcieri:
@ibotty as a bit of an unfortunate secret, the blanket impl of `Encoding` for `Alphabet` leaks the `Alphabet` trait into the public API. Here's an example:

https://github.com/RustCrypto/JOSE/blob/master/jose-b64/src/stream/dec.rs#L46

...so you should be able to add this implementation directly to `sha-crypt` crate while still leveraging the `decode_6bits`/`encode_6bits` methods from `base64ct`.

ibotty:
I did not notice that Alphabet is public API.  That makes this change indeed more troublesome.

This encoding is used by some other crypt schemes like md5 (I have a patch coming) and sha1, afair.  des crypt and maybe argon (I have not looked into these) are the only crypt schemes supporting the other base64 encoding afaict.  So I'd rather have it in `base64ct`.

I guess the ShaCrypt `Alphabet` implementation could as well just overwrite the functions.  This way it could still be in `base64ct`.  What do you think?

tarcieri:
Yeah, if you can just add a new alphabet (possibly borrowing the `DECODE`/`ENCODE` constants from e.g. `Base64Crypt`) which overrides the relevant trait methods, that should work and be non-breaking.

ibotty:
surpassed by #742.  Closing

PR #740, opened 2022-10-13
tls_codec: Implement `Ord/PartialOrd` for VLBytes

nomaxg:
I am working on switching hash references in OpenMLS to [variable length vectors](https://github.com/openmls/openmls/issues/928#event-7556019970). The references will be stored as `VLBytes`. OpenMLS contains tree operations that depend on an `Ord` implementation for hash references.


franziskuskiefer (approved):
Thanks

PR #742, opened 2022-10-17
Add sha2-style base64 crypt format

ibotty:
This replaces #738.  It is minimally invasive.

### DIFF
--- a/base64ct/src/alphabet.rs
+++ b/base64ct/src/alphabet.rs
@@ -7,6 +7,7 @@ use core::{fmt::Debug, ops::RangeInclusive};
 
 pub mod bcrypt;
 pub mod crypt;
+pub mod shacrypt;
 pub mod standard;
 pub mod url;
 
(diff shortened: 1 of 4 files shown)


base64ct/src/alphabet.rs
```diff
@@ -70,7 +71,7 @@ pub trait Alphabet: 'static + Copy + Debug + Eq + Send + Sized + Sync {
         ret
     }
 
-    /// Encode 3-bytes of a Base64 message.
+    // Encode 3-bytes of a Base64 message.
```
tarcieri, line 74: This seems unnecessary

tarcieri (approved):
One nit if you care to fix it, otherwise I can merge as-is

base64ct/src/alphabet.rs
ibotty: Oh, I missed that.  Sorry.  Fixed in latest commit.

ibotty:
Thank you!

PR #751, opened 2022-10-31
tls_codec: fix empty variable-length vectors and add MLS limit

franziskuskiefer:
This change fixes a bug where empty variable-length encoded vectors lead to an endless loop because (triggered in https://github.com/openmls/openmls/pull/1082).

It also introduces an `mls` feature flag that limits the length of variable-length encoded vectors to 30 bits. Without this features the default behaviour from QUIC is used that allows 62 bits.

cc @raphaelrobert @kkohbrok

### DIFF
--- a/tls_codec/src/quic_vec.rs
+++ b/tls_codec/src/quic_vec.rs
@@ -6,7 +6,11 @@
 //! Note that we require, as the MLS specification does, that vectors have to
 //! use the minimum number of bytes necessary for the encoding.
 //! This ensures that encodings are unique.
-
+//!
+//! With the `mls` feature the length of variable length vectors can be limited
+//! to 30-bit values.
+//! This is in contrast to the default behaviour defined by RFC 9000 that allows
+//! up to 62-bit length values.
 use alloc::vec::Vec;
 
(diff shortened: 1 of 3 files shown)


tls_codec/src/quic_vec.rs
```diff
@@ -6,15 +6,26 @@
...
 use serde::{Deserialize as SerdeDeserialize, Serialize as SerdeSerialize};
 
 use crate::{Deserialize, Error, Serialize, Size};
 
+#[cfg(not(feature = "mls"))]
 const MAX_LEN: u64 = 0x3fff_ffff_ffff_ffff; // <= (1<<62)-1
+#[cfg(not(feature = "mls"))]
+const MAX_LEN_LEN: usize = 8;
+#[cfg(feature = "mls")]
+const MAX_LEN: u64 = 0x3fff_ffff; // <= (1<<30)-1
```
raphaelrobert, line 26: For clarity, I would use the expression from the comments: `(1 << 30) - 1`

The same also applies to the original length obviously.

raphaelrobert (approved):
Just a nit, the rest looks good

PR #756, opened 2022-11-09
feat(tls_codec): Implement `tls_codec(skip)` support.

duesee:
This PR implements support for a `skip` attribute on structs. When using ...

```rust
#[derive(TlsDeserialize, TlsSerialize, TlsSize)]
struct StructWithSkip {
  #[tls_codec(skip)]
  a: u8,
  b: u8,
}
```

... only `b` will be serialized. During deserialization `Default::default()` is called to populate `a`.

When `Default` is not implemented for an attribute we get something like ...

```
error[E0277]: the trait bound `NonDefaultField: Default` is not satisfied
  --> tls_codec/derive/tests/attributes.rs:34:14
   |
34 |     #[derive(TlsDeserialize, TlsSerialize, TlsSize)]
   |              ^^^^^^^^^^^^^^ the trait `Default` is not implemented for `NonDefaultField`
```

It is also not allowed to use `skip` twice on the same attribute. Open question: should we accept `skip` case-insensitive?

This is a first (unpolished) working version but it may have some problems I overlooked. There is also no documentation yet. First time doing something with procedural macros :-)


tls_codec/derive/src/lib.rs
```diff
@@ -151,13 +152,20 @@ enum TlsAttr {
...
 }
 
 impl TlsAttr {
     fn name(&self) -> &'static str {
         match self {
             TlsAttr::With(_) => "with",
             TlsAttr::Discriminant(_) => "discriminant",
+            TlsAttr::Skip => "skip",
         }
     }
```
franziskuskiefer, line 171: Can you update the comment on `parse`?

tls_codec/derive/src/lib.rs
```diff
@@ -232,16 +254,29 @@ fn function_prefix(field: &Field) -> Result<Prefix> {
...
-                format!("Unrecognized field attribute `{}`", attr.name()),
-            )),
+            (path, _) => Ok(path),
         })?
         .map(Prefix::Custom)
         .unwrap_or_else(|| Prefix::Type(field.ty.clone()));
     Ok(prefix)
 }
 
+fn function_skip(field: &Field) -> Result<bool> {
```
franziskuskiefer, line 264: Please add comments to function to exaplain what they are doing.

tls_codec/derive/src/lib.rs
```diff
@@ -424,7 +465,27 @@ fn impl_tls_size(parsed_ast: TlsStruct) -> TokenStream2 {
...
             member_prefixes,
+            member_skips,
         }) => {
+            // Filter out skipped members.
+            let (members, member_prefixes) = {
+                let mut tmp_members: Vec<Member> = Vec::new();
+                let mut tmp_member_prefixes: Vec<Prefix> = Vec::new();
+
+                for ((member, prefix), skip) in members
+                    .into_iter()
```
franziskuskiefer, line 476: Using `iter` everywhere here and only cloning the `member` when needed should be more efficient.

tls_codec/derive/src/lib.rs
```diff
@@ -497,7 +558,27 @@ fn impl_serialize(parsed_ast: TlsStruct) -> TokenStream2 {
             generics,
             members,
             member_prefixes,
+            member_skips,
         }) => {
+            // Filter out skipped members.
+            let (members, member_prefixes) = {
+                let mut tmp_members: Vec<Member> = Vec::new();
```
franziskuskiefer, line 565: Same code as above? Can we pull this out to avoid code duplication?

tls_codec/derive/src/lib.rs
```diff
@@ -587,7 +668,37 @@ fn impl_deserialize(parsed_ast: TlsStruct) -> TokenStream2 {
             generics,
             members,
             member_prefixes,
+            member_skips,
         }) => {
+            // Partition skipped and non-skipped members.
+            let (members, member_prefixes, members_default, member_prefixes_default) = {
+                let mut tmp_members: Vec<Member> = Vec::new();
```
franziskuskiefer, line 675: Almost the same here.

franziskuskiefer (changes requested):
Genereally looks good to me with a few comments inline.

duesee:
Thanks for the review. I'll cleanup, deduplicate, and add a bit of documentation.

duesee:
I used the `trybuild` crate. Not sure if this is too much. But it might be good to already have that in place in case we want to work on user-facing errors in the future. The alternative is, as you said, to use doc comments with `compile_fail` on private functions.

tls_codec/derive/Cargo.toml
```diff
@@ -20,3 +20,4 @@ proc-macro2 = "1.0"
 
 [dev-dependencies]
 tls_codec = { path = "../" }
+trybuild = "*"
```
franziskuskiefer, line 23: Let's use a version.

franziskuskiefer (changes requested):
I think using `trybuild` is fine. Only one nit.

franziskuskiefer (approved):
Thanks

duesee:
I should have mentioned that I intended to cleanup the commits before merging. It's okay but having "fixups" in the commit history is not great. Sorry for that.

franziskuskiefer:
It's getting squashed anyway 🙂

duesee:
Ah, true that :-)

PR #760, opened 2022-11-14
x509: `countryName` should always be `PrintableString`

tomleavy:
I was doing some testing and came across the fact that creating a CSR / Certificate with countryName set in the DN leads to a different output. I narrowed it down to formats producing UTF8String for countryName and OpenSSL 3.0 producing PrintableString. 

I also came across this stack overflow post on the subject. The countryName field is required to be PrintableString according to X.502. https://stackoverflow.com/questions/17215598/country-name-field-in-ca-generated-by-openssl-is-encoded-as-printablestring-whil

If this isn't the right place to put this logic please let me know, thanks!

### DIFF
--- a/x509-cert/src/attr.rs
+++ b/x509-cert/src/attr.rs
@@ -1,6 +1,7 @@
 //! Attribute-related definitions as defined in X.501 (and updated by RFC 5280).
 
 use alloc::vec::Vec;
+use const_oid::db::rfc4519::COUNTRY_NAME;
 use core::fmt::{self, Write};
 
 use const_oid::db::DB;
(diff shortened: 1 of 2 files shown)


carl-wallace:
I hit the same thing last week in DNs created from strings but had not looked into resolving it yet. This looks like it ought do but it may be worth adding DomainComponent as well, since it is an IA5String. Another approach would be to see if the value only contains PrintableString characters and default to that in those cases (except where PrintableString is not permitted, like DomainComponent).

tomleavy:
Thanks @carl-wallace , I added the DomainComponent being IA5String as well. I think going to PrintableString in the cases where it can has too many unforeseeable consequences, better to tackle these as a one off as that's what OpenSSL already does.

tarcieri:
Thanks!

PR #761, opened 2022-11-15
Implement Arbitrary for Certificate

stormshield-guillaumed:
Add an optional dependency on the arbitrary crate to implement Arbitrary for Certificate. This allows to generate arbitrary certificates for fuzzing.


tarcieri:
As a gut reaction, this is a surprisingly invasive change

const-oid/src/lib.rs
```diff
@@ -1,4 +1,4 @@
-#![no_std]
+#![cfg_attr(not(feature = "std"), no_std)]
```
tarcieri, line 1: Please don't change these. This makes the semantics non-additive. It's done this way deliberately.

I'm not sure what it's breaking in `arbitrary` not to have this but that seems like a poor use of macros if it's depending on e.g. preludes.

const-oid/Cargo.toml
```diff
@@ -16,12 +16,16 @@ readme = "README.md"
...
 
+[dependencies]
+arbitrary = { version = "1.2.0", features = ["derive"], optional = true }
+
 [dev-dependencies]
 hex-literal = "0.3"
 
 [features]
-std = []
+arbitrary = ["std", "dep:arbitrary"]
```
tarcieri, line 26: This requires an MSRV bump, which we're not ready to do yet.

I plan on landing #757 first.

stormshield-guillaumed: It's feasible to have the same semantic by changing the feature name and using the implicit `arbitrary` feature inside. However, with [this](https://github.com/RustCrypto/formats/pull/761/files#diff-4256d0986a772eb1e13583d5ddc3d60712c86ebf05dbd03b57a52329d0f17cc0R33), the removal of the `?` will unconditionally enable the `const-oid` dependency when the feature is enabled, and I don't know an other way of doing this.

const-oid/src/lib.rs
stormshield-guillaumed: Without this change, the code expanded by the macro doesn't compile because the `thread_local` macro cannot be found. I'll try to see if I can make a patch on the `arbitrary` side.

Why does this makes the semantics non-additive ? `serde` [does it this way](https://github.com/serde-rs/serde/blob/f41509261e32357e28a6bc6f46f58e98d841e7e8/serde/src/lib.rs#L86). If the feature `std` is disabled, `no_std` is applied, and if the feature `std` is enabled, `no_std` is not applied.

tarcieri: Because you no longer have a consistent baseline: with the current approach, a `no_std` profile is always used as a baseline, and then `std` added back with `extern crate`. Really it's `no_std` that's backwards to begin with, but this approach you're using absence of a feature to remove functionality.

Admittedly it's gotten better than before, when libcore was conditionally linked based on the *presence* of `no_std`, leading to ridiculous hacks like `use core as std`.

But really the problem here is `arbitrary` is depending on the `std` prelude, and yes, that should get fixed. (Why does it even need `thread_local` in the macro anyway?)

stormshield-guillaumed: Ok, I get the difference with the prelude now. I made a PR on rust-fuzz/arbitrary#131. I'll revert the `no_std` change and update this PR when the fix is released.

const-oid/Cargo.toml
tarcieri: We bumped MSRV to 1.60 so this is fine now

tarcieri:
@stormshield-guillaumed you can bump the clippy rustc version, although it might need some code changes

stormshield-guillaumed:
After a quick check, the `arbitrary` feature doesn't seem to be included with `cargo hack test --feature-powerset`. It seems to be because of the `cargo-hack` version used not supporting `dep:` features. They started supporting it with version [`0.5.15`](https://github.com/taiki-e/cargo-hack/blob/3b289aeb08b7779f825c78e4a464ccb5fa47c46a/CHANGELOG.md#0515---2022-07-18). I think more CI checks will be failing because I didn't pay attention, but the MSRV of `arbitrary` is [`1.63.0`](https://github.com/rust-fuzz/arbitrary/blob/8fabb537dd39903ccd5679600f25b6b0002be74b/Cargo.toml#L20), so it won't build with `1.60.0`.

Also, I saw that the `rust-version` of the `x509-cert` crate is `1.56.0` but the CI runs on `1.60.0`.

stormshield-guillaumed:
I checked and `arbitrary` really requires `1.63.0` because they use `array::from_fn`.

tarcieri:
FWIW we're looking at bumping MSRV to 1.65 in #797

tarcieri:
This bumps the `cargo-hack` version to v0.5.24: https://github.com/RustCrypto/actions/pull/22

der/src/asn1/any.rs
```diff
@@ -23,6 +23,7 @@ use crate::asn1::ObjectIdentifier;
 /// Nevertheless, this crate defines an `ANY` type as it remains a familiar
 /// and useful concept which is still extensively used in things like
 /// PKI-related RFCs.
+#[cfg_attr(feature = "arbitrary", derive(arbitrary::Arbitrary))]
```
tarcieri, line 26:
```suggestion
#[cfg_attr(
    feature = "arbitrary",
    derive(arbitrary::Arbitrary),
    allow(clippy::integer_arithmetic)
)]
```

der/src/asn1/any.rs
```diff
@@ -202,6 +203,7 @@ impl<'a> TryFrom<&'a [u8]> for AnyRef<'a> {
 /// backing data.
 #[cfg(feature = "alloc")]
 #[cfg_attr(docsrs, doc(cfg(feature = "alloc")))]
+#[cfg_attr(feature = "arbitrary", derive(arbitrary::Arbitrary))]
```
tarcieri, line 206:
```suggestion
#[cfg_attr(
    feature = "arbitrary",
    derive(arbitrary::Arbitrary),
    allow(clippy::integer_arithmetic)
)]
```

der/src/asn1/any.rs
stormshield-guillaumed: I tried this and a few other places for the `allow` but none suppresses the warning. I don't know how to put attributes on derive generated code and adding the `allow` to the module seems a bit extreme (suppressing more than wanted).

tarcieri: I guess the issue is it's annotating the struct and not the associated `impl` blocks.

You can add it at to the module if it's gated on `feature = "arbitrary"`

stormshield-guillaumed:
The only remaining errors are on the minimal version check, but I already specify `arbitrary = "1.2.0"`, so I don't know what I can do to solve this.

tarcieri:
I opened an issue on `arbitrary`: https://github.com/rust-fuzz/arbitrary/issues/134

If you'd like to get the build passing now you can temporarily disable the check in the crates that are failing, although I'd ask you do that in a separate PR/commit so it's easy to revert when the issue does get fixed.

tarcieri:
@stormshield-guillaumed can you either PR b4ac2a6 separately or squash all of the commits down to two, one which adds the `Arbitrary` impl, and the other with the contents of b4ac2a6 so it can be easily reverted?

tarcieri:
Thanks!

PR #769, opened 2022-11-25
spki: make `AlgorithmIdentifier` generic around `Params`

tarcieri:
NOTE: breaking change.

Previously `AlgorithmIdentifier::parameters` were always `AnyRef`. This commit changes them to a generic parameter `Params`.

An alias `AlgorithmIdentifierRef` provides a type identical to the original with `AnyRef` as its `parameters`, which is used in all of the other crates in this repo.

cc @baloo @lumag

### DIFF
--- a/spki/src/algorithm.rs
+++ b/spki/src/algorithm.rs
@@ -2,8 +2,10 @@
 
 use crate::{Error, Result};
 use core::cmp::Ordering;
-use der::asn1::{AnyRef, ObjectIdentifier};
-use der::{Decode, DecodeValue, DerOrd, Encode, Header, Reader, Sequence, ValueOrd};
+use der::{
+    asn1::{AnyRef, Choice, ObjectIdentifier},
+    Decode, DecodeValue, DerOrd, Encode, Header, Reader, Sequence, ValueOrd,
+};
 
 /// X.509 `AlgorithmIdentifier` as defined in [RFC 5280 Section 4.1.1.2].
 ///
@@ -15,16 +17,68 @@ use der::{Decode, DecodeValue, DerOrd, Encode, Header, Reader, Sequence, ValueOr
 ///
 /// [RFC 5280 Section 4.1.1.2]: https://tools.ietf.org/html/rfc5280#section-4.1.1.2
 #[derive(Copy, Clone, Debug, Eq, PartialEq, PartialOrd, Ord)]
-pub struct AlgorithmIdentifier<'a> {
+pub struct AlgorithmIdentifier<Params> {
     /// Algorithm OID, i.e. the `algorithm` field in the `AlgorithmIdentifier`
     /// ASN.1 schema.
     pub oid: ObjectIdentifier,
 
     /// Algorithm `parameters`.
-    pub parameters: Option<AnyRef<'a>>,
+    pub parameters: Option<Params>,
 }
 
-impl<'a> AlgorithmIdentifier<'a> {
+impl<'a, Params> DecodeValue<'a> for AlgorithmIdentifier<Params>
+where
+    Params: Choice<'a>,
+{
+    fn decode_value<R: Reader<'a>>(reader: &mut R, header: Header) -> der::Result<Self> {
+        reader.read_nested(header.length, |reader| {
+            Ok(Self {
+                oid: reader.decode()?,
+                parameters: reader.decode()?,
+            })
+        })
+    }
+}
+
+impl<'a, Params> Sequence<'a> for AlgorithmIdentifier<Params>
+where
+    Params: Choice<'a> + Encode,
+{
+    fn fields<F, T>(&self, f: F) -> der::Result<T>
+    where
+        F: FnOnce(&[&dyn Encode]) -> der::Result<T>,
+    {
+        f(&[&self.oid, &self.parameters])
+    }
+}
+
+impl<'a, Params> TryFrom<&'a [u8]> for AlgorithmIdentifier<Params>
+where
+    Params: Choice<'a> + Encode,
+{
+    type Error = Error;
+
+    fn try_from(bytes: &'a [u8]) -> Result<Self> {
+        Ok(Self::from_der(bytes)?)
+    }
+}
+
+impl<Params> ValueOrd for AlgorithmIdentifier<Params>
+where
+    Params: DerOrd,
+{
+    fn value_cmp(&self, other: &Self) -> der::Result<Ordering> {
+        match self.oid.der_cmp(&other.oid)? {
+            Ordering::Equal => self.parameters.der_cmp(&other.parameters),
+            other => Ok(other),
+        }
+    }
+}
+
+/// `AlgorithmIdentifier` reference which has `AnyRef` parameters.
+pub type AlgorithmIdentifierRef<'a> = AlgorithmIdentifier<AnyRef<'a>>;
+
+impl<'a> AlgorithmIdentifierRef<'a> {
     /// Assert the `algorithm` OID is an expected value.
     pub fn assert_algorithm_oid(&self, expected_oid: ObjectIdentifier) -> Result<ObjectIdentifier> {
         if self.oid == expected_oid {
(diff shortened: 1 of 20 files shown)


spki/src/spki.rs
```diff
@@ -36,7 +39,7 @@ use der::pem::PemLabel;
 #[derive(Copy, Clone, Debug, Eq, PartialEq)]
 pub struct SubjectPublicKeyInfo<'a> {
     /// X.509 [`AlgorithmIdentifier`] for the public key type
-    pub algorithm: AlgorithmIdentifier<'a>,
+    pub algorithm: AlgorithmIdentifierRef<'a>,
```
tarcieri, line 42: Note that this would be a lot more useful if `Params` were also added to `SubjectPublicKeyInfo`, i.e. this were changed to `AlgorithmIdentifier<Params>`.

It should probably get tackled as part of this PR.

spki/src/algorithm.rs
```diff
@@ -15,16 +17,68 @@ use der::{Decode, DecodeValue, DerOrd, Encode, Header, Reader, Sequence, ValueOr
...
+        }
+    }
+}
+
+/// `AlgorithmIdentifier` reference which has `AnyRef` parameters.
+pub type AlgorithmIdentifierRef<'a> = AlgorithmIdentifier<AnyRef<'a>>;
+
+impl<'a> AlgorithmIdentifierRef<'a> {
     /// Assert the `algorithm` OID is an expected value.
     pub fn assert_algorithm_oid(&self, expected_oid: ObjectIdentifier) -> Result<ObjectIdentifier> {
```
tarcieri, line 83: I left the inherent methods unchanged for now, however it's now possible to use `AlgorithmIdentifier<ObjectIdentifier>` which eliminates the need for many of them, since the main thing they provide is type conversions to/from `ObjectIdentifier`.

spki/src/spki.rs
tarcieri: Likewise it should be changed on `PrivateKeyInfo` in the `pkcs8` crate

tarcieri:
This is a weird failure:

https://github.com/RustCrypto/formats/actions/runs/3551054810/jobs/5964957749

`rstest` is failing? But Cargo.lock didn't change

spki/src/algorithm.rs
lumag: The assert_algorithm_oid should be implemented for the generic type, not for the AlgorithmIdentifierRef

x509-cert/src/crl.rs
```diff
@@ -75,7 +75,7 @@ pub struct RevokedCert<'a> {
 #[allow(missing_docs)]
 pub struct TbsCertList<'a> {
     pub version: Version,
-    pub signature: AlgorithmIdentifier<'a>,
+    pub signature: AlgorithmIdentifierRef<'a>,
```
lumag, line 78: This should also allow `AlgorithmIdentifier<...>` (and in `Certificate`/`CertificateList`) too.
I can not come with the good enough argument at this moment.

spki/src/spki.rs
lumag: I think it should be some type bound rather than generic `<Params>`

lumag:
@tarcieri please consider also merging https://pastebin.com/fNszVrjA

I tried generalizing `SubjectPublicKeyInfo`, but got stuck.

spki/src/spki.rs
lumag: After trying to implement generic arguments to SPKI or to Certificate/TBSCertificate, I think that `Params` might need to be bound with something like `Into<AnyRef>`, or any other generic enough type, so that there would be no need to add type parameters to these structures. I might be wrong with the SPKI, but handling type parameters for Certificate looks like a complete mess.

tarcieri: I don't follow what you mean by "type bound" if not a generic parameter.

If you need additional bounds you can always specify them. But requiring one would be rather onerous I think, especially as the conversion for e.g. `ObjectIdentifier` only works for `&ObjectIdentifier`.

tarcieri:
> @tarcieri please consider also merging https://pastebin.com/fNszVrjA

I think some of that looks good but would rather hold off on doing those sorts of updates in follow-up PRs. Maybe you can PR them separately after this is merged?

This PR isn't done yet and it's going to get more complex with changes to `SubjectPublicKeyInfo` and `PrivateKeyInfo`. And continuing to use `AnyRef` as `Params` should still be fine for now.

baloo:
> This is a weird failure:
> 
> https://github.com/RustCrypto/formats/actions/runs/3551054810/jobs/5964957749
> 
> `rstest` is failing? But Cargo.lock didn't change

https://github.com/RustCrypto/formats/pull/767#issuecomment-1327036134

nightly regression.

lumag:
@tarcieri sure, I can update `RsaPssParams` afterwards. The last chunk (regarding `assert_algorithm_oid`) can probably be a part of this PR.
I'm looking forward to seeing how you'd update SPKI/`PrivateKeyInfo`.

tarcieri:
@lumag ffb9092 makes `SubjectPublicKeyInfo` generic around `Params`, following much the same pattern as I used for `AlgorithmIdentifier`

Unfortunately this isn't enough to eliminate the lifetime and allow a fully owned form. That would require making the public key data generic around e.g. a `Bits` parameter.

spki/src/spki.rs
lumag: I'm still playing here. I have the feeling that TBSCertitificate & company should not have generic parameters per se. Otherwise handling them easily becomes a mess. I'm still trying to fit SPKI / SPKIRef properly into my code. Maybe this will flow better once I update RSA crate to return specific SPKI rather than generic SPKIRef.

tarcieri: Yes, they should use a concrete type. The only things I think it makes sense parameterizing are `AlgorithmIdentifier`, `SubjectPublicKeyInfo`, and `PrivateKeyInfo`.

lumag: Is there any kind of performing `Into<AnyRef>` or `TryInto<AnyRef>` conversion? The only way I have at this moment is to encode the `AlgorithmIndeifier<Params>` into DER format and then reparse it again into `AlgorithmIndentifierRef`.

baloo:
I've played a bit with the PR.
If I understand correctly the idea would be to get something like:
<details>
<summary>patch on `x509-cert/src/certificate.rs`</summary>

```
diff --git a/x509-cert/src/certificate.rs b/x509-cert/src/certificate.rs
index ec607eb..28cc2d6 100644
--- a/x509-cert/src/certificate.rs
+++ b/x509-cert/src/certificate.rs
@@ -8,7 +8,7 @@ use core::cmp::Ordering;
 use const_oid::AssociatedOid;
 use der::asn1::BitString;
 use der::{Decode, Enumerated, Error, ErrorKind, Sequence, ValueOrd};
-use spki::{AlgorithmIdentifierRef, SubjectPublicKeyInfoRef};
+use spki::{AlgorithmIdentifier, SubjectPublicKeyInfoRef};

 #[cfg(feature = "pem")]
 use der::pem::PemLabel;
@@ -73,7 +73,7 @@ impl Default for Version {
 /// [RFC 5280 Section 4.1]: https://datatracker.ietf.org/doc/html/rfc5280#section-4.1
 #[derive(Clone, Debug, Eq, PartialEq, Sequence, ValueOrd)]
 #[allow(missing_docs)]
-pub struct TbsCertificate<'a> {
+pub struct TbsCertificate<'a, S, K> {
     /// The certificate version
     ///
     /// Note that this value defaults to Version 1 per the RFC. However,
@@ -84,11 +84,11 @@ pub struct TbsCertificate<'a> {
     pub version: Version,

     pub serial_number: SerialNumber,
-    pub signature: AlgorithmIdentifierRef<'a>,
+    pub signature: AlgorithmIdentifier<S>,
     pub issuer: Name,
     pub validity: Validity,
     pub subject: Name,
-    pub subject_public_key_info: SubjectPublicKeyInfoRef<'a>,
+    pub subject_public_key_info: SubjectPublicKeyInfo<'a, K>,

     #[asn1(context_specific = "1", tag_mode = "IMPLICIT", optional = "true")]
     pub issuer_unique_id: Option<BitString>,
@@ -100,7 +100,7 @@ pub struct TbsCertificate<'a> {
     pub extensions: Option<crate::ext::Extensions>,
 }

-impl<'a> TbsCertificate<'a> {
+impl<'a, S, K> TbsCertificate<'a, S, K> {
     /// Decodes a single extension
     ///
     /// Returns an error if multiple of these extensions is present. Returns
@@ -147,15 +147,15 @@ impl<'a> TbsCertificate<'a> {
 /// [RFC 5280 Section 4.1]: https://datatracker.ietf.org/doc/html/rfc5280#section-4.1
 #[derive(Clone, Debug, Eq, PartialEq, Sequence, ValueOrd)]
 #[allow(missing_docs)]
-pub struct Certificate<'a> {
-    pub tbs_certificate: TbsCertificate<'a>,
-    pub signature_algorithm: AlgorithmIdentifierRef<'a>,
+pub struct Certificate<'a, S, K> {
+    pub tbs_certificate: TbsCertificate<'a, S, K>,
+    pub signature_algorithm: AlgorithmIdentifier<S>,
     pub signature: BitString,
 }

 #[cfg(feature = "pem")]
 #[cfg_attr(docsrs, doc(cfg(feature = "pem")))]
-impl PemLabel for Certificate<'_> {
+impl PemLabel for Certificate<'_, _, _> {
     const PEM_LABEL: &'static str = "CERTIFICATE";
 }

@@ -170,4 +170,4 @@ impl PemLabel for Certificate<'_> {
 /// ```
 ///
 /// [RFC 6066]: https://datatracker.ietf.org/doc/html/rfc6066#section-10.1
-pub type PkiPath<'a> = Vec<Certificate<'a>>;
+pub type PkiPath<'a, S, K> = Vec<Certificate<'a, S, K>>;
```

</details>

I'm not sure the `Vec<Certificate<'a, S, K>>` would make much sense as there is no obligation for a verification path to use the same signature or same key algorithm?

tarcieri:
@baloo I don't think it makes sense to parameterize `Certificate` around generic `Params`.

Instead they can use owned/allocating `Any` now, and runtime dispatch based on the provided `AlgorithmIdentifier`. That moves us one step closer to owned types.

baloo:
Ha, right, I missed the intention of the `Params` completely.

tarcieri:
I think it allows for a number of cross-cutting concerns, including using a concrete type like `ObjectIdentifier` when it makes sense.

We're also trying to solve the problem of having `AlgorithmIdentifier` constants where the `Params` are e.g. a SEQUENCE.

PR #791, opened 2022-12-12
tls_codec: rename `serde_serialize` feature => `serde`; MSRV 1.60

tarcieri:
Uses namespaced features introduced in Rust 1.60 to allow the `serde` feature name to be the same as the `serde` crate name without having to rename the `serde` crate.


tarcieri:
Note: this PR is something of an excuse to bump `tls_codec`'s MSRV to 1.60 so we can add it back to the workspace.

franziskuskiefer (approved):
I meant to bump the MSRV but somehow dropped the ball on it...
Thanks!

PR #799, opened 2022-12-17
Make PSS/OAEP params use generic AlgorithmIdentifier

lumag:


tarcieri (approved):

pkcs1/src/params.rs
```diff
@@ -84,7 +81,7 @@ pub struct RsaPssParams<'a> {
     pub hash: AlgorithmIdentifierRef<'a>,
 
     /// Mask Generation Function (MGF)
-    pub mask_gen: AlgorithmIdentifierRef<'a>,
+    pub mask_gen: AlgorithmIdentifier<AlgorithmIdentifierRef<'a>>,
```
baloo, line 84: Context: I've just submitted https://github.com/RustCrypto/formats/pull/1415 that rewrites the `AlgorithmIdentifier` to make sure the parameter is either `Any` or `AnyRef`.

I'm confused about this change. This doesn't seem to line up with https://datatracker.ietf.org/doc/html/rfc5280#section-4.1.1.2 (`spki::AlgorithmIdentifier`)
```
   AlgorithmIdentifier  ::=  SEQUENCE  {
        algorithm               OBJECT IDENTIFIER,
        parameters              ANY DEFINED BY algorithm OPTIONAL  }
```
 
Which, from my understand, ask that the parameters are encoded as an `Any`.

This change moves the implementation to remove the type prefix that `Any` would provide.

That said, this `AlgorithmIdentifier` would be defined by https://datatracker.ietf.org/doc/html/rfc8017#page-70:
```
   -- ================
   --   Useful types
   -- ================

   ALGORITHM-IDENTIFIER ::= CLASS {
       &id    OBJECT IDENTIFIER  UNIQUE,
       &Type  OPTIONAL
   }
       WITH SYNTAX { OID &id [PARAMETERS &Type] }
```

Which one is right here?
Could it be we need another definition of `AlgorithmIdentifier` altogether?

I think I would otherwise revert that change and get back the `SEQ_OID_SHA_1_DER` that was removed.

Issue #779, opened 2022-11-30
pkcs7+der: mixed BER/DER encoding with indefinite length


smndtrl:
Hi,

while looking into the `pkcs7` crate for a CMS usecase around the Apple world I discovered that their detached signatures use the BER indefinite length encoding for some of the `SEQUENCE` which of course is not supported by the `der` crate.

Has there been any thoughts/discussions around if that's something the formats repo should address or if that is out of scope (for now) and left to other crates.

[...]

PR #810, opened 2022-12-19
WIP: Parsing with `lax` mode

smndtrl:
This allows parsing DER with some "features" from BER that are in use (in the apple universe)

* Reading indefinite length tags by using `length = 0`
* Reading constructed/encapsulated OctetStrings (only if 1! OctetString is part of a indefinite length OctetString) as that is east without alloc

This is my first try to address [this issue](https://github.com/RustCrypto/formats/issues/779)

### DIFF
--- a/der/Cargo.toml
+++ b/der/Cargo.toml
@@ -30,6 +30,7 @@ proptest = "1"
 [features]
 alloc = []
 derive = ["der_derive"]
+lax = []
 oid = ["const-oid"]
 pem = ["alloc", "pem-rfc7468/alloc", "zeroize"]
 real = []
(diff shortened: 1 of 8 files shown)


tarcieri:
Since features are additive across crates, this really needs to be a separate "profile" which can be used concurrently with fully validating DER.

That will likely involve adding some kind of trait for profiles and a generic parameter to e.g. `Decode` and/or `Reader`

smndtrl:
Thanks for the info. Learned something new about features today.

However the approach you mentioned may be outside of what I'm comfortable doing with Rust atm.

tarcieri:
Going to close this then

Issue #765, opened 2022-11-19
x509-cert: move to all owned types?


tarcieri:
The `x509-cert` crate currently has a hard dependency on `alloc`, but several types still have a lifetime which borrows from the input, which precludes one-pass decoding from PEM (where Base64 is decoded directly into owned types) and makes writing things like certificate builders harder.

We could potentially move entirely to types which own their backing data, eliminating lifetimes from all types. We could still potentially have a corresponding set of borrowed e.g. `Ref` types (see https://github.com/RustCrypto/formats/issues/689), but using owned types would make usage more convenient at the cost of losing zero-copy decoding from DER bytes.

baloo:
I don't really see a way to make a better certificate builder API without this. The current version of the API kind of work, but it's pretty inconvenient to use.

carl-wallace:
I'd prefer to see owned and no copy be peers rather than owned replacing no copy (even if owned is why got the shorter crate name). The draft changes would generate a pile of work to catch up and each could be useful to have.

tarcieri:
@carl-wallace as I mentioned, we can still have `*Ref` types which would be useful for #689

baloo:
So do we want just duplicate all the types to add a `Foo` along with `FooRef` ?
For example the `Name<'a>`, the tree of types there is pretty heavy.

I'm okay with that, just want to make sure, for example the `impl` of `AnyRef` is pretty heavy, and duplicating that is a lot of code.
https://github.com/RustCrypto/formats/blob/c7c1fa73277007b83a92aa77c31d427a05cdf62b/der/src/asn1/any.rs#L35-L162

Or do I just throw a macro in there an factorize out both bodies?

tarcieri:
@baloo for now I’d say make everything owned, and circle back in `*Ref` types later as part of #689, which needs additional design work anyway

carl-wallace:
Could we have proc macros emit a second set of structures such that one set behaves the old way and one set uses owned? On Nov 26, 2022, at 11:19 AM, Tony Arcieri ***@***.***> wrote:﻿
@baloo for now I’d say make everything owned, and circle back in *Ref types later as part of #689, which needs additional design work anyway

tarcieri:
> Could we have proc macros emit a second set of structures 

Really if we want to have unified borrowed/owned types I would suggest evaluating `yoke` for this purpose (see #734)

It would allow us to define zero-copy types and owned wrappers which borrow from them.

[...]

PR #813, opened 2022-12-20
Adds SignedData type to `pkcs7`

smndtrl:
- Adds SignedData type to `ContentInfo`
- Two examples + test cases (Apple MDM Signature and SCEP Request from Apple device) originally BER encoded but reencoded in DER
- Requires `x509-cert` now for some types

Missing
- Some types (like `AttributeCertificateV2`) I haven't encountered 
- More asserts in the new tests. Currently only tests if `e_content` ambiguity in PKCS#7 vs CMS standard was handled
- ValueOrd implementation.. what makes sense?


pkcs7/src/certificate_choices.rs
```diff
@@ -0,0 +1,64 @@
...
+    #[asn1(context_specific = "3", tag_mode = "IMPLICIT")]
+    Other(OtherCertificateFormat<'a>),
+}
+
+// TODO: figure out what ordering makes sense - if any
+impl ValueOrd for CertificateChoices<'_>  {
+    fn value_cmp(&self, _other: &Self) -> der::Result<Ordering> {
+        Ok(Ordering::Equal)
+    }
+}
```
woodruffw, line 64: My suggestion would be to derive `ValueOrd`, since there's no defined ordering between these choices and the default behavior for that derive is just to order lexicographically.

woodruffw:
Thanks for opening this! I'll try it with some PKCS7 blobs I have locally. 

I had a similar line of work prepared (<https://github.com/RustCrypto/formats/compare/master...woodruffw-forks:formats:ww/signeddata?expand=1>), but this looks like it's closer to complete.

pkcs7/src/signed_data_content.rs
```diff
@@ -0,0 +1,54 @@
...
+
+/// ```text
+/// DigestAlgorithmIdentifier ::= AlgorithmIdentifier
+/// ```
+type DigestAlgorithmIdentifier<'a> = AlgorithmIdentifierRef<'a>;
+
+/// ```text
+/// DigestAlgorithmIdentifiers ::= SET OF DigestAlgorithmIdentifier
+/// ```
+type DigestAlgorithmIdentifiers<'a> = SetOfVec<DigestAlgorithmIdentifier<'a>>;
```
woodruffw, line 15: Maybe I did something weird locally, but I'm surprised this isn't complaining about the lack of the `alloc` feature from the `der` crate. Did you build with all features enabled locally?

pkcs7/src/encapsulated_content_info.rs
```diff
@@ -0,0 +1,30 @@
...
+/// content [0] EXPLICIT ANY DEFINED BY contentType OPTIONAL
+/// ```
+/// or
+/// ```text
+/// eContent [0] EXPLICIT OCTET STRING OPTIONAL
+/// ```
+#[derive(Clone, Copy, Debug, Eq, PartialEq, Sequence)]
+pub struct EncapsulatedContentInfo<'a> {
+    /// indicates the type of content.
+    pub e_content_type: ContentType,
```
woodruffw, line 24: I think this needs to be a `type ContentType = ObjectIdentifier;` alias instead, since `crate::ContentType` is a limited enum of valid CMS content types, not `SignedData` content types.

woodruffw: Per the RFC:

```asn1
      ContentInfo ::= SEQUENCE {
        contentType ContentType,
        content [0] EXPLICIT ANY DEFINED BY contentType }

      ContentType ::= OBJECT IDENTIFIER
```

woodruffw:
With the change in https://github.com/RustCrypto/formats/pull/813#discussion_r1053567241, this gets through most of the MS authroot PKCS7 blob I'm testing with:

```console
$ ./target/debug/examples/demo ~/Downloads/authroot.stl
thread 'main' panicked at 'failed to load CTL: Der(Error { kind: Length { tag: Tag(0x02: INTEGER) }, position: Some(Length(165291)) })', examples/demo.rs:12:29
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace

$ wc -c ~/Downloads/authroot.stl
  165282 /Users/william/Downloads/authroot.stl
```

It seems to jump off the end of the file, though.

For context, here's the file (zipped): [authroot.stl.zip](https://github.com/RustCrypto/formats/files/10270816/authroot.stl.zip)

pkcs7/src/signed_data_content.rs
smndtrl: Now that you say it I'm wondering as well. I ran everything without additional features enabled.

woodruffw: Yeah, it built for me locally without additional features as well, strange.

tarcieri: You're pulling in the `x509-cert` crate which unconditionally activates the `alloc` feature, and feature unification means it's activated here too.

It would probably be good to factor the relevant parts into a common crate which can be shared. We had the `x501` crate for this purpose at one point, although I'm not sure that's quite what you need.

woodruffw: Thanks for explaining!

If there *is* interest in maintaining a non-`alloc` part of `pkcs7`, this is how I did it in my own changes:

```toml
[features]
alloc = ["der/alloc"]
std = ["alloc"]
signed-data = ["alloc", "dep:x509-cert"]
```

(with `x509-cert` then being marked as optional.)

tarcieri: I'm guessing that, like #765, this crate should probably move to all owned types and make `alloc` a hard dependency, since heapless support probably won't be practical for real-world use cases

woodruffw:
> weird, I'll take a look when I get a chance, but can you put up the relevant document at https://lapo.it/asn1js/ or something and I can take a look?

It's around 165KB, so it looks like it won't fit into a shareable URL 😅 

I narrowed it down to the `SET OF SignerInfo`, which I've screencapped below:

Screenshot 2022-12-20 at 1 23 41 PM

(The only integers seem to be the `CMSVersion` for the `SignerInfo` itself and a 150-bit integer for the `SubjectKeyIdentifier`).

carl-wallace:
I was able to parse your blob using my fork after changing EncapsulatedContentInfo to use AnyRef instead of OctetStringRef: https://github.com/carl-wallace/formats/commit/698857883a7675200909495140d262baa9c068a5. 

 

From: William Woodruff ***@***.***>
Reply-To: RustCrypto/formats ***@***.***>
Date: Tuesday, December 20, 2022 at 1:26 PM
To: RustCrypto/formats ***@***.***>
Cc: Subscribed ***@***.***>
Subject: Re: [RustCrypto/formats] Adds SignedData type to `pkcs7` (PR #813)

 

weird, I'll take a look when I get a chance, but can you put up the relevant document at https://lapo.it/asn1js/ or something and I can take a look?

It's around 165KB, so it looks like it won't fit into a shareable URL 😅

I narrowed it down to the SET OF SignerInfo, which I've screencapped below:

(The only integers seem to be the CMSVersion for the SignerInfo itself and a 150-bit integer for the SubjectKeyIdentifier).

woodruffw:
> I was able to parse your blob using my fork after changing EncapsulatedContentInfo to use AnyRef instead of OctetStringRef: [carl-wallace@6988578](https://github.com/carl-wallace/formats/commit/698857883a7675200909495140d262baa9c068a5).

Thanks for testing. It's weird that it works there, but not here -- are you allowing BER encoding for integers?

smndtrl:
It's me using a u8 instead of UintRef in `IssuerAndSerialNumber`

woodruffw:
> It's me using a u8 instead of UintRef in `IssuerAndSerialNumber`

I can confirm that makes the parse succeed for me locally! Thanks a ton for debugging!

carl-wallace:
I see it’s OBE now, but I made no changes to tolerate BER (but that’s likely going to be necessary).

 

From: William Woodruff ***@***.***>
Reply-To: RustCrypto/formats ***@***.***>
Date: Tuesday, December 20, 2022 at 1:59 PM
To: RustCrypto/formats ***@***.***>
Cc: Carl Wallace ***@***.***>, Comment ***@***.***>
Subject: Re: [RustCrypto/formats] Adds SignedData type to `pkcs7` (PR #813)

 

I was able to parse your blob using my fork after changing EncapsulatedContentInfo to use AnyRef instead of OctetStringRef: ***@***.***

Thanks for testing. It's weird that it works there, but not here -- are you allowing BER encoding for integers?

pkcs7/src/signed_data_content.rs
smndtrl: What should we do for now? I don't want to mix this PR with the larger effort of doing something like #765 for this crate as well.

smndtrl:
I checked @carl-wallace's code and saw a lot of similar code in a new `cms` crate. Before I put this PR ready for review I wanted to check with you @tarcieri if we should get this PR going with the solution of using AnyRef to deal with the CMS vs PKCS#7 or use a `cms` crate and look to deal with all the common things in some way?

pkcs7/src/signed_data_content.rs
woodruffw: I'm +1 on doing the owned refactor in a follow-up (I can help with that) to keep the diff small 🙂

tarcieri:
@smndtrl add it to `pkcs7` for now and we can circle back on @carl-wallace's implementation

woodruffw:
I played around with this a bit, and I noticed that the current `EncapsulatedContentInfo` handling of `e_content` means that the `e_content`'s top-level `SEQUENCE` gets consumed if present, which results in a harder-to-parse slice if the contents are themselves DER-encoded.

On the other hand, if I change it to this:

```rust
#[derive(Clone, Copy, Debug, Eq, PartialEq, Sequence)]
pub struct EncapsulatedContentInfo<'a> {
    /// indicates the type of content.
    pub e_content_type: ObjectIdentifier,

    /// encapsulated content
    pub e_content: Option<AnyRef<'a>>,
}
```

I get back the interior `SEQUENCE` as expected.

Curious about thoughts on this -- it surprises me that the current behavior (i.e. with `#[asn1(context_specific = "0", optional = "true", tag_mode = "EXPLICIT")]` is to consume the `SEQUENCE`, but it's possible I'm misunderstanding something there.

tarcieri:
@woodruffw if that's true, that sounds like a bug.

Could you put together a minimal reproduction and file an issue?

woodruffw:
To clarify what I mean: without the changes above, the `AnyRef`'s slice points to the highlighted member:

Screenshot 2022-12-21 at 12 43 26 PM

whereas I *expect* it to point to the parent `SEQUENCE` of that memebr:

Screenshot 2022-12-21 at 12 47 17 PM



> @woodruffw if that's true, that sounds like a bug.
> 
> Could you put together a minimal reproduction and file an issue?

Yep, can do!

woodruffw:
Just answering my own question/confusion, based on investigation in #815: I think the behavior here is currently correct, and I was just holding the `AnyRef` API incorrectly. The correct way to use it is `decode_into`, not to grab the underlying `value()` slice and treat that as a complete DER stream.

woodruffw:
@roblabla this work may also be interesting to you (I saw you're also working on Authenticode support in the repo history).

pkcs7/src/lib.rs
```diff
@@ -8,14 +8,20 @@
 #![forbid(unsafe_code, clippy::unwrap_used)]
 #![warn(missing_docs, rust_2018_idioms, unused_qualifications)]
 
+mod certificate_choices;
+mod cms_version;
 mod content_info;
 mod content_type;
+mod revocation_info_choices;
+mod signer_info;
```
roblabla, line 16: Shouldn't this module be public? Otherwise, how can we access the SignerInfo data?

woodruffw: Yeah, these should probably be `pub` (or have specific re-exports).

pkcs7/src/signer_info.rs
```diff
@@ -0,0 +1,107 @@
...
+
+    #[asn1(context_specific = "0")]
+    SubjectKeyIdentifier(SubjectKeyIdentifier<'a>),
+}
+
+#[derive(Clone, Debug, Eq, PartialEq, Sequence)]
+#[allow(missing_docs)]
+pub struct IssuerAndSerialNumber<'a> {
+    pub name: Name,
+    pub serial_number: UintRef<'a>,
```
roblabla, line 50: Hmm, this is unfortunate. We have an `x509_cert::serial_number::SerialNumber` type, and I feel like this is what should be used here?

woodruffw (approved):
LGTM! I've been testing this locally for the past week, without problems.

I think the only thing that _could_ go into this is sanity-checking the `SignedData` content based on the `CmsVersion`, but it looks like `x509-cert` doesn't do that for its version either. So maybe that's a follow-on item.

pkcs7/src/cms_version.rs
```diff
@@ -0,0 +1,68 @@
...
+
+impl EncodeValue for CmsVersion {
+    fn value_len(&self) -> der::Result<Length> {
+        u8::from(*self).value_len()
+    }
+
+    fn encode_value(&self, writer: &mut dyn Writer) -> der::Result<()> {
+        u8::from(*self).encode_value(writer)
+    }
+}
```
tarcieri, line 68: I haven't tested this edit specifically to ensure it adds all of these same impls, but you should be able to eliminate much of the boilerplate code here using custom derive

```suggestion
#[derive(Clone, Copy, Debug, Enumerated, Eq, PartialEq)]
#[asn1(type = "INTEGER")]
#[repr(u8)]
pub enum CmsVersion {
    /// syntax version 0
    V0 = 0,
    /// syntax version 1
    V1 = 1,
    /// syntax version 2
    V2 = 2,
    /// syntax version 3
    V3 = 3,
    /// syntax version 4
    V4 = 4,
    /// syntax version 5
    V5 = 5,
}
```

tarcieri (changes requested):
This looks good aside from using custom derive for `CmsVersion`

jathan:
Would this patch properly parse certificate bundles like [http://repo.fpki.gov/fcpca/caCertsIssuedByfcpcag2.p7c](http://repo.fpki.gov/fcpca/caCertsIssuedByfcpcag2.p7c)?  As I was trying to parse it I ran across the same error It "found" the signed content.  Perhaps a good test case?  (See https://playbooks.idmanagement.gov/fpki/common/certificates/)

I'm new to Rust and was banging my head against the wall trying to figure out why this wasn't working...

woodruffw:
From a quick look, that does indeed look like a PKCS#7 SignedData. So this patch would parse it, although it would then be up to you to interpret the SignedData's body correctly (and to actually verify the signature on it).

pkcs7/src/cms_version.rs
smndtrl: Seems to work fine just as is. Pushed the change to use the macro

pkcs7/src/signer_info.rs
```diff
@@ -0,0 +1,105 @@
...
+    #[asn1(context_specific = "1", tag_mode = "IMPLICIT", optional = "true")]
+    pub unsigned_attributes: Option<UnsignedAttributes<'a>>,
+}
+
+// TODO: figure out what ordering makes sense - if any
+impl ValueOrd for SignerInfo<'_> {
+    fn value_cmp(&self, _other: &Self) -> der::Result<Ordering> {
+        Ok(Ordering::Equal)
+    }
+}
```
tarcieri, line 105: You should be able to derive `ValueOrd`

smndtrl: That'll end up requiring ValueOrd on CmsVersion and SignerIdentifier (which works nice with just adding ValueOrd there) but

```
the method `der_cmp` exists for enum `CmsVersion`, but its trait bounds were not satisfied
the following trait bounds were not satisfied:
`CmsVersion: ValueOrd`
which is required by `CmsVersion: DerOrd`
```

which ends up in when adding ValueOrd 

```
invalid `asn1` attribute (valid options are `tag_mode`)
```

on `CmsVersion`

tarcieri: Alright, I can go ahead and merge and see if I can figure it out

tarcieri (approved):

tarcieri:
@smndtrl if you could open a PR which adds test vectors for the original BER-encoded e.g. Apple MDM signature (or just paste the relevant bytes here as e.g. hex), I can see if I can find a solution for handling BER for this specific case

pkcs7/src/signer_info.rs
tarcieri: Opened #825 to address this

PR #814, opened 2022-12-21
tls_codec: feat: Add support for `Arbitrary` crate.

duesee:
This PR introduces an `arbitrary` feature that can be used to unlock `#[derive(Arbitrary)]` for `VLBytes` and `VLByteSlice`. The [`Arbitrary`](https://crates.io/crates/arbitrary) crate allows to create arbitrary instances of a struct that can be used, e.g., during fuzz-testing. I've also included a fuzz target to show how it works.

It is useful to have an `Arbitrary` implementation for `VLBytes` and `VLByteSlice` because `Arbitrary` can be automatically derived for structs/enums when all members implement `Arbitrary`. The reasons to provide support for `Arbitrary` in `tls_codec` pretty much align with the reasons to provide support for Serde in `tls_codec`.

Edit: I'm currently figuring out how to implement support for `TlsVecU8` etc. and should also limit the size for `VLBytes`.


duesee:
I've converted this PR to a draft because CI is not happy. Should I put `--exclude-features arbitrary` here https://github.com/RustCrypto/formats/blob/master/.github/workflows/tls_codec.yml#L41?

duesee:
I removed the `Arbitrary` derive for `VLByteSlice` and limited the length of `VLBytes` to `MAX_LEN`. So, this PR only adds `Arbitrary` support for `VLBytes` for now.

franziskuskiefer (approved):
Thanks, lgtm
Do you want to do this for the TLS vecs and other built-in types as well?

duesee:
> Do you want to do this for the TLS vecs and other built-in types as well?

I think that `tls_codec` should have all these implementations, yes. But they require a bit more thought which is why I will take a look next year.

PR #817, opened 2022-12-23
asn1/octet_string: add `OctetStringRef::decode_into`

woodruffw:
This is just a small API addition, designed to help with the (unfortunately) common case of DER nested within `OCTET STRING` slices.

This is effectively just a shorthand for `T::from_der(o.as_bytes())` for `T: Decode`.

(If this change isn't desired, please feel free to close!)


tarcieri:
Seems fine. I've encountered valid use cases.

PR #820, opened 2022-12-26
x509-cert: add `Display` impl for `SerialNumber`

woodruffw:
This adds a `Display` implementation for `SerialNumber`.

I went with a MAC-style format because that's the format OpenSSL uses.


tarcieri (approved):

tarcieri:
Thanks!

woodruffw:
Thanks as well!

PR #822, opened 2022-12-27
added `impl AsRef<[u8]> for SecretDocument`

StackOverflowExcept1on:
Hey, I'm using `std::fs::write` and its signature looks like that: (it has AsRef generic)
image

and I think I found one hole in your API:
```
error[E0277]: the trait bound `SecretDocument: AsRef<[u8]>` is not satisfied
   --> keypair/build.rs:25:55
    |
25  |     fs::write(keys_dir.join("private_key_pkcs1.der"), private_key_der)?;
    |     ---------                                         ^^^^^^^^^^^^^^^ the trait `AsRef<[u8]>` is not implemented for `SecretDocument`
    |     |
    |     required by a bound introduced by this call
```

 Now the developer needs to manually call `.as_bytes()` for `SecretDocument`, but `Document` already has `impl AsRef<[u8]> for Document`
image1

### DIFF
--- a/der/src/document.rs
+++ b/der/src/document.rs
@@ -287,6 +287,14 @@ impl SecretDocument {
         write_secret_file(path, self.to_pem(label, line_ending)?.as_bytes())
     }
 }
+
+#[cfg(feature = "zeroize")]
+impl AsRef<[u8]> for SecretDocument {
+    fn as_ref(&self) -> &[u8] {
+        self.as_bytes()
+    }
+}
+
 #[cfg(feature = "zeroize")]
 impl Debug for SecretDocument {
     fn fmt(&self, fmt: &mut fmt::Formatter<'_>) -> fmt::Result {


tarcieri:
This was intentional so as to avoid indiscriminately leaking secret values through implicit/inferred `AsRef` conversions.

That said, there is a "convenience" argument to getting access to key material. The question is how much of an issue unintentionally leaking key material via `AsRef` might come up in practice.

StackOverflowExcept1on:
@tarcieri can you explain me how it can leak? Look at like line `#[cfg(feature = "zeroize")]`

and also we have `Drop` that zeroes bytes:
```rust
impl Drop for SecretDocument {
    fn drop(&mut self) {
        self.0.der_bytes.zeroize();
    }
}
```

At least on Linux you can't read the memory of another process without root and `sudo sysctl kernel.yama.ptrace_scope=0`. If you are clearly not doing something damned, then it is unlikely that someone steals a DER private key in the userspace.

tarcieri:
Using inherent methods ensures that the caller explicitly wants the bytes of the secret accessed.

With `AsRef`, the caller may pass the secret by reference, only to have the callee invoke `AsRef` despite the caller not wanting the callee to, and that may not be immediately obvious to the caller.

StackOverflowExcept1on:
@tarcieri even if you pass a reference to the bytes containing the private key, what happens? Rust has borrow checker and when you have `&T` the reference points to valid `T`. Any attempt to move this reference out of scope will be disallowed by the rust compiler.

Let's say you did something stupid like this: `let ptr = document.as_ref() as *const u8;` In any case, we have `Drop` that will zero memory after object goes out of function scope.

tarcieri:
To back up a bit, the overall design goal here is [misuse resistance](https://news.ycombinator.com/item?id=14292223).

`SecretDocument` contains a cryptographic secret i.e. key material.

The APIs in the @RustCrypto libraries are designed to try to reduce potential misuses when handling key material.

Some potential threats related to misuses here are the logging of key material.

`AsRef` is glue which lets some function which doesn't know the concrete type the caller is passing get at the bytes of the underlying value. It may not be designed at all for the handling of sensitive key material.

The goal of using *explicit* APIs is to make sure the caller is aware that they are passing off the bytes of the secret, and not handing it off to some function which is going to broadcast secret key material to some kind of e.g. logging system, because it wasn't designed to handle key material.

Make sense?

StackOverflowExcept1on:
Yes, it makes some sense, but in terms of memory security, there is no problem. I would like a review from someone else

tarcieri:
I never said there was a memory security/safety issue.

I'm also the primary reviewer for this repository.

I'm not completely opposed to adding an `AsRef` impl, I was merely trying to spell out the rationale for not having one.

tarcieri:
Well, I haven't heard back, so I'm closing this with the rationale that the absence of the `AsRef` impl is deliberate to avoid accidental leakage of secret key material.

PR #827, opened 2022-12-30
WIP: add test for mixed DER/BER encoding CMS

smndtrl:
As requested in #813 a test for mixed encoding CMS.

@tarcieri I didn't find my original BER file from the Apple MDM Signature so I just checked the example at https://lapo.it/asn1js/# and it's very similar in that it starts with an indefinite length tag, uses multiple along the way and also uses constructed undefined length OctetStrings.

Both the DER/BER file have the same `e_content`. Currently the DER encoded file is used in the test and works fine.


pkcs7/tests/content_tests.rs
```diff
@@ -140,3 +140,27 @@ fn decode_signed_scep_example() {
...
+                    .unwrap()
+                    .decode_into::<OctetStringRef>()
+                    .unwrap()
+                    .as_bytes()
+                    .len(),
+                10034
+            )
+        }
+        _ => panic!("expected ContentInfo::SignedData(Some(_))"),
+    }
```
tarcieri, line 165:
```suggestion
    let bytes = include_bytes!("examples/cms_der.bin");

    let content = match ContentInfo::from_der(bytes) {
        Ok(ContentInfo::SignedData(Some(data))) => data,
        other => panic!("unexpected result: {:?}", other),
    };

    assert_eq!(
        content
            .encap_content_info
            .e_content
            .unwrap()
            .decode_into::<OctetStringRef>()
            .unwrap()
            .as_bytes()
            .len(),
        10034
    );
```

tarcieri:
Looking at the BER example, it looks like it's using indefinite lengths fairly pervasively for several (nested) fields.

Making this work will probably involve modifying custom derive to be able to generate `Decode` impls for BER-tolerant SEQUENCEs.

tarcieri:
Here's a stepping stone towards a solution: https://github.com/RustCrypto/formats/pull/830

tarcieri:
Merged in #840

PR #828, opened 2023-01-01
der: eliminate dynamism from encoding

tarcieri:
Replaces use of `dyn Writer` with `impl Writer`.

Removes the `Sequence::fields` method, which was used for a blanket impl of `EncodeValue`.

Instead, very small changes to the custom derive can be used to derive a monomorphic implementation of `EncodeValue` that doesn't rely on trait objects, where calls to the `Writer` can be potentially inlined.

### DIFF
--- a/der/LICENSE-MIT
+++ b/der/LICENSE-MIT
@@ -1,4 +1,4 @@
-Copyright (c) 2020-2022 The RustCrypto Project Developers
+Copyright (c) 2020-2023 The RustCrypto Project Developers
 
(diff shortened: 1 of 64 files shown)


tarcieri:
NOTE: `der` and `der_derive` have been updated. Everything else still needs to be updated.

tarcieri:
Thanks to its use of custom derive, no changes to `x509-cert` seem to be required, which is pretty neat.

PR #831, opened 2023-01-02
der, x509-cert: fix handling of negative integers

### LINKED ISSUE #826: x509-cert: another parse failure on publicly known cert
Like #821, although probably a different cause: [`d3eefbcbbcf49867838626e23bb59ca01e305db7`](www.download.windowsupdate.com/msdownload/update/v3/static/trustedr/en/d3eefbcbbcf49867838626e23bb59ca01e305db7.crt) in the Windows CTL fails to parse with:

```
ASN.1 DER message is too long at DER byte 60
```

The offending cert, as PEM:

```
-----BEGIN CERTIFICATE-----
MIIDcTCCAlmgAwIBAgIVAOYJ/nrqAGiM4CS07SAbH+9StETRMA0GCSqGSIb3DQEB
BQUAMFAxCzAJBgNVBAYTAlBMMSgwJgYDVQQKDB9LcmFqb3dhIEl6YmEgUm96bGlj
emVuaW93YSBTLkEuMRcwFQYDVQQDDA5TWkFGSVIgUk9PVCBDQTAeFw0xMTEyMDYx
MTEwNTdaFw0zMTEyMDYxMTEwNTdaMFAxCzAJBgNVBAYTAlBMMSgwJgYDVQQKDB9L
cmFqb3dhIEl6YmEgUm96bGljemVuaW93YSBTLkEuMRcwFQYDVQQDDA5TWkFGSVIg
Uk9PVCBDQTCCASIwDQYJKoZIhvcNAQEBBQADggEPADCCAQoCggEBAKxHL49ZMTml
6g3wpYwrvQKkvc0Kc6oJ5sxfgmp1qZfluwbv88BdocHSiXlY8NzrVYzuWBp7J/9K
ULMAoWoTIzOQ6C9TNm4YbA9A1jdX1wYNL5Akylf8W5L/I4BXhT9KnlI6x+a7BVAm
nr/Ttl+utT/Asms2fRfEsF2vZPMxH4UFqOAhFjxTkmJWf2Cu4nvRQJHcttB+cEAo
ag/hERt/+tzo4URz6x6r19toYmxx4FjjBkUhWQw1X21re//Hof2+0YgiwYT84zLb
eqDqCOMOXxvH480yGDkh/QoazWX3U75HQExT/iJlwnu7I1V6HXztKIwCBjsxffbH
3jOshCJtywcCAwEAAaNCMEAwDwYDVR0TAQH/BAUwAwEB/zAOBgNVHQ8BAf8EBAMC
AQYwHQYDVR0OBBYEFFOSo33/gnbwM9TrkmdHYTMbaDsqMA0GCSqGSIb3DQEBBQUA
A4IBAQA5UFWd5EL/pBviIMm1zD2JLUCpp0mJG7JkwznIOzawhGmFFaxGoxAhQBEg
haP+E0KR66oAwVC6xe32QUVSHfWqWndzbODzLB8yj7WAR0cDM45ZngSBPBuFE3Wu
GLJX9g100ETfIX+4YBR/4NR/uvTnpnd9ete7Whl0ZfY94yuu4xQqB5QFv+P7IXXV
lTOjkjuGXEcyQAjQzbFaT9vIABSbeCXWBbjvOXukJy6WgAiclzGNSYprre8Ryydd
fmjW9HIGwsIO03EldivvqEYL1Hv1w/Pur+6FUEOaL68PEIUovfgwIB2BAw+vZDuw
cH0mX548PojGyg434cDjkSXa3mHF
-----END CERTIFICATE-----
```

Happy to debug this again, just making the issue for visibility.

woodruffw:
This is my first attempt at a fix for the problem in #826.

I tried going with a `signed: bool` field at first, but ran into a lot of friction: I ended up having to essentially duplicate all of `uint.rs` in `int.rs`. So I scrapped that, and went with @baloo's suggestion of an enum: `IntRef` now varies over `SintRef` and `UintRef`, and provides a uniform interface by forwarding to each.

Advantages to this approach:

1. Less churn: I only had to make a small number of changes to `int.rs` to accommodate these changes, mostly around refusing to accept positive numbers and making sure that we handle the sign byte correctly.
2. Construction correctness in the API: `SintRef` (now poorly named) can be used to guarantee construction of a negative integer, similar to how `UintRef` can only construct positive integers.

Disadvantages:
1. `Sint` and `SintRef` are poorly named, since they only represent negative integers (and zero). I'm open to naming suggestions here; I left them this way just to get an initial PR up.
2. There's still some dual state around zero: `Int::Unsigned(0) != Int::Signed(0)`. This relates to the first problem, and I can fix it as part of naming `Sint` something better (like `Negative`) and forbidding zero by construction. 

Notably, that second disadvantage would still exist with the `signed: bool` approach, since care would need to be taken to ensure a canonical `signed: false` value when building `Int(0)`.

cc @baloo 

Fixes #826.


woodruffw:
(I can also confirm that these changes allow me to successfully parse and re-encode all 447 X.509 certs in the Windows trust list :slightly_smiling_face:)

x509-cert/src/serial_number.rs
```diff
@@ -34,13 +34,15 @@ impl SerialNumber {
...
+        let inner = Uint::new(bytes)?;
 
         if inner.len() > SerialNumber::MAX_LEN {
             return Err(ErrorKind::Overlength.into());
         }
 
-        Ok(Self { inner })
+        Ok(Self {
+            inner: inner.into(),
+        })
```
woodruffw, line 45: With this change, users will no longer be able to directly build a new negative serial number: the only way they'll be able to end up with a `SerialNumber` with a negative internal representation is by loading an existing one.

tarcieri:
I very strongly think signed integers should just use the 2's complement representation and not be any sort of enum which seems to be adding a lot of incidental complexity.

2's complement already solves the problem of "how do we store an integer which can be negative or positive" without resorting to an enum.

The `INTEGER` type is natively 2's complement, and the way "unsigned integers" are stored with them is to use the 2's complement representation but always add a leading 0 in the event the high bit of the MSB is set.

With that said, another approach is making `UInt` a newtype for `Int` which upholds that invariant and disallows the most significant bit of the most significant byte from being set, which would eliminate duplication.

I'm happy to take a crack at this if you're having trouble.

woodruffw:
That makes sense, thanks! I agree about the complexity (and I'm not happy with how big this change is).

I think the thing that tripped me up is over-thinking RFC 5280's length constraint on serial numbers, when really it's ambiguous: the RFC limits serial number values to 20 octets, but doesn't say whether that's 20 *with* the zero-extension byte or not.

It looks like this also caused some confusion in Go's `crypto/x509` (https://github.com/golang/go/issues/33310), and that their decision was to allow users to build serial numbers that *encode* to 21 bytes under the reasoning that their unsigned representation is only 20 bytes ([source ref](https://github.com/golang/go/blob/master/src/crypto/x509/x509.go#L1539-L1546)). 

The current `Int` type that's been merged is 2's complement internally, so I think all that really needs to be done is to (1) adjust `SerialNumber` so that it can only be constructed with positive integers, and (2) re-think its internal length check (in particular, we probably need to allow a length of 21 bytes *iff* the highest byte is a zero extension).

Does that make sense?

tarcieri:
Yeah sure. The existing `Int`/`IntRef` types are indeed already using 2's complement.

In parallel I can look at composing `Uint`/`UintRef` in terms of `Int`/`IntRef`, which should cut down on duplication.

woodruffw:
Sounds good, I can make those changes this afternoon/evening.

der/src/asn1/integer/int.rs
```diff
@@ -46,9 +46,9 @@ where
     writer.write(strip_leading_ones(bytes))
 }
 
-/// Get the encoded length for the given signed integer serialized as bytes.
+/// Get the encoded length for the given **negative** integer serialized as bytes.
 #[inline]
-pub(super) fn encoded_len(bytes: &[u8]) -> Result<Length> {
+pub(super) fn negative_encoded_len(bytes: &[u8]) -> Result<Length> {
```
woodruffw, line 51: N.B.: I renamed this to emphasize what this function is actually doing, as currently implemented: it doesn't calculate the encoded length of any signed integer represented as bytes, but only negative signed integers. Passing a positive signed integer here would result in the wrong length, since we only trim leading ones, not zeroes.

(I confirmed that the one remaining callsite for this only calls it on negative integers, and removed the other misleading callsites.)

der/src/asn1/integer/bigint.rs
```diff
@@ -261,6 +265,22 @@ mod allocating {
...
+            if value.value_len().expect("invalid Uint") > value.len() {
+                inner.push(0x00);
+            }
+
+            inner.extend_from_slice(value.as_bytes());
+            let inner = BytesOwned::new(inner).expect("invalid Uint");
+
+            Int { inner }
+        }
+    }
```
woodruffw, line 282: This provides the infallible `Uint -> Int` conversion, with infallibility under the assumption that the `Uint` was constructed correctly. This is the same as the other `From<...>` implementation directly above, but I can make it `TryFrom` instead if you'd prefer.

x509-cert/src/serial_number.rs
```diff
@@ -32,15 +32,27 @@ impl SerialNumber {
...
+        // `SerialNumber` *decodings* to have up to 21 bytes below.
+        if inner.value_len()? > SerialNumber::MAX_LEN {
             return Err(ErrorKind::Overlength.into());
         }
 
-        Ok(Self { inner })
+        Ok(Self {
+            inner: inner.into(),
+        })
     }
```
woodruffw, line 56: This ensures that we never create a *new* negative serial number, but the way it does that is potentially surprising because of how the `Int::new` API (unchanged) works. In particular, `Int::new` *assumes* that its input is positive, and encodes it correctly to enforce this assumption.

I've updated the tests below to show this behavior.

woodruffw: (This arguably violates POLA, since we're changing the serial number given to us by the user. But it's how the current `Int::new` API behaves, and I didn't want to make changes to that before putting something initial up here :slightly_smiling_face:)

baloo: curious: what does POLA stands for?

tarcieri: Principle of least astonishment

woodruffw: Yep! Sorry for the acronym abuse.

By POLA in this case I meant that users might consider it astonishing that the serial they provide doesn't end up in the cert, even if the *result* is technically more correct.

tarcieri:
@woodruffw can you rebase?

woodruffw:
> @woodruffw can you rebase?

Yep, can do this afternoon!

woodruffw:
Okay, should be up-to-date again. It looks like a bunch of `TryFrom<AnyRef>` impls were removed with #834, so I dropped the one here as well.

der/src/asn1/integer/bigint.rs
tarcieri: `From` should be fine here

tarcieri (approved):

tarcieri:
Thanks!

PR #836, opened 2023-01-03
WIP: Add support for pkcs12 decode

smndtrl:
This is a WIP. Putting it here to prevent duplicate work.

Missing:
* Docs
* Encoding


pkcs12/src/authenticated_safe.rs
```diff
@@ -0,0 +1,163 @@
...
+            AuthenticatedSafeItem::EnvelopedData(data) => f(&[
+                &self.content_type(),
+                &data.as_ref().map(|d| ContextSpecific {
+                    tag_number: CONTENT_TAG,
+                    tag_mode: TagMode::Explicit,
+                    value: d.clone(),
+                }),
+            ]),
+        }
+    }
```
tarcieri, line 103: FYI, this method was removed in #828 

Here's a similar example, although it could probably be cleaned up a bit / DRYed out: https://github.com/RustCrypto/formats/pull/828/files#diff-7eada31819a5738c81669b63773252d75c0ae287f02296637ee5862e36b18408

tarcieri:
We went with #1165 instead. Thanks though!

PR #837, opened 2023-01-03
Revert "const-oid+der+spki+x509-cert: temporarily disable CI check until rust-fuzz/arbitrary#134 is fixed"

tarcieri:
This reverts commit 517adb7b82ee3774156a80c256d0b26e14246f74.

The issue in `arbitrary` has been fixed: rust-fuzz/arbitrary#134


tarcieri:
Hmm, seems this wasn't actually fixed yet. Opened https://github.com/rust-fuzz/arbitrary/issues/136

PR #841, opened 2023-01-09
Misc. improvements for x509-certs 

npmccallum:


tarcieri:
Re: `TryFrom<&[u8]>`, I think it's nice to have on types which represent the toplevel `SEQUENCE`s of a document. I agree it doesn't have to be everywhere.

The advantage over `Decode::from_der` is in the 2021 edition `TryFrom` is in the prelude, so it skips an import.

@baloo it'd be good if you can take a look at this to see if it overlaps with what you've been working on

x509-cert/src/time.rs
```diff
@@ -62,7 +62,7 @@ impl Time {
 }
 
 impl fmt::Display for Time {
-    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> core::result::Result<(), fmt::Error> {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> Result<(), fmt::Error> {
```
baloo, line 65: nit:
```suggestion
    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
```
https://doc.rust-lang.org/stable/core/fmt/type.Result.html

baloo (approved):

baloo:
no overlap that I can see, all of it makes sense.

x509-cert/src/request.rs
```diff
@@ -73,31 +65,23 @@ impl<'a> TryFrom<&'a [u8]> for CertReqInfo<'a> {
...
     const PEM_LABEL: &'static str = "CERTIFICATE REQUEST";
 }
 
-impl<'a> TryFrom<&'a [u8]> for CertReq<'a> {
-    type Error = der::Error;
-
-    fn try_from(bytes: &'a [u8]) -> Result<Self, Self::Error> {
-        Self::from_der(bytes)
-    }
-}
```
tarcieri, line 99: I think this one should stay, as it's a toplevel type

x509-cert/src/time.rs
npmccallum: Fixed.

x509-cert/src/request.rs
npmccallum: Fixed.

PR #845, opened 2023-01-12
der: add `OctetString::into_bytes()`

abonander:
This makes it possible to extract the owned data without copying.


tarcieri:
@abonander needs rustfmt

abonander:
Fixed.

PR #849, opened 2023-01-19
Initial check-in of crmf crate

carl-wallace:
Note, EnvelopedData usage is temporarily commented out and will be restored when cms crate is pushed.


crmf/Cargo.toml
```diff
@@ -0,0 +1,39 @@
...
+# cms = { version = "0.1.0", path = "../cms" }
+flagset = { version = "0.4.3" }
+
+[dev-dependencies]
+hex-literal = "0.3"
+
+# NOTE: upgrading requires MSRV bumps
+# - rstest v0.13 requires 1.59 (for `async-global-executor`)
+# - rstest v0.14 requires a workspace-wide 1.60 MSRV (for namespaced features)
+rstest = "0.12"
```
tarcieri, line 30: These updates have already happened

carl-wallace: Deleted it as it was not required.

crmf/src/pop.rs
```diff
@@ -0,0 +1,354 @@
...
+//     }
+// }
+// impl<'a> ::der::Tagged for EncKeyWithIdChoice<'a> {
+//     fn tag(&self) -> ::der::Tag {
+//         match self {
+//             Self::String(_) => <Utf8StringRef<'a> as ::der::FixedTag>::TAG,
+//             Self::GeneralName(_) => <GeneralName as ::der::FixedTag>::TAG,
+//         }
+//     }
+// }
```
tarcieri, line 326: Can you omit this for now, rather than checking in commented code?

You can move the comment on it to e.g. the [`Choice` definition above](https://github.com/RustCrypto/formats/pull/849/files#diff-27e5055eda84edafd5049a59f707ea90673a0e090e1ef00fa1b942cd37bd7d02R226), or anywhere else that makes sense.

carl-wallace: Done

tarcieri:
I'm going to do a quick pass of some small fixups

PR #851, opened 2023-01-19
initial check-in of cmpv2 crate

carl-wallace:


cmpv2/src/certified_key_pair.rs
```diff
@@ -0,0 +1,99 @@
...
+/// ```text
+///  CertOrEncCert ::= CHOICE {
+///      certificate     [0] CMPCertificate,
+///      encryptedCert   [1] EncryptedValue }
+/// ```
+///
+/// [RFC 4210 Section 5.3.4]: https://www.rfc-editor.org/rfc/rfc4210#section-5.3.4
+#[derive(Clone, Debug, Eq, PartialEq, Choice)]
+#[allow(missing_docs)]
+#[allow(clippy::large_enum_variant)]
```
tarcieri, line 52: Might be worth boxing the large enum variant

cmpv2/src/status.rs
```diff
@@ -0,0 +1,233 @@
...
+///      -- to match this confirmation with the corresponding req/rep
+///      statusInfo  PKIStatusInfo OPTIONAL }
+/// ```
+///
+/// [RFC 4210 Section 5.2.18]: https://www.rfc-editor.org/rfc/rfc4210#section-5.2.18
+#[derive(Clone, Debug, Eq, PartialEq, Sequence)]
+#[allow(missing_docs)]
+pub struct CertStatus<'a> {
+    pub cert_hash: OctetString,
+    // using serial number to allow for large integers here
```
tarcieri, line 230: Use `Int`?

carl-wallace: Yep. Changed here and in several other spots.

cmpv2/src/certified_key_pair.rs
carl-wallace: Fixed for this struct. There are a few more. Will do those too with similar approach.

cmpv2/Cargo.toml
```diff
@@ -0,0 +1,38 @@
...
+spki = { version = "0.7.0-pre", path = "../spki" }
+flagset = { version = "0.4.3" }
+crmf = {version = "0.0", path = "../crmf" }
+
+[dev-dependencies]
+const-oid = { version = "0.10.0-pre", path = "../const-oid" }
+hex-literal = "0.3"
+#reqwest = "0.11.13"
+#tokio = { version = "1.19.2", features = ["full", "time", "rt-multi-thread"] }
+#tokio-test = {version = "0.4.2"}
```
tarcieri, line 29: Can you remove these if they're (currently) unused?

carl-wallace: Done. Also removed the commented out test case that these were associated with.

tarcieri (approved):
Approving this although I think we should try to get the `cmp` crate name.

AFAICT there isn't actually a Certificate Management Protocol V2 and the `cmpv2` name is only to avoid a crate name clash?

I will inquire with the current owner if they would be okay with giving it to us for this purpose.

PR #854, opened 2023-01-21
Fix tls_codec-ci badge

mcpherrinm:
The path needed to be updated per
https://github.com/badges/shields/issues/8671


tarcieri:
Thanks!

PR #856, opened 2023-01-23
removed #[allow(clippy::large_enum_variant)] for PkiArchiveOptions

carl-wallace:
- removed #[allow(clippy::large_enum_variant)] for PkiArchiveOptions in the crmf crate by changing definition of EncryptedKey to boxed fields and adding impls for Box\<EnvelopedData\> to the cms crate.
- fixed copy and paste error in PkiArchiveOptions content_specific attributes
- added manual test for PkiArchiveOptions change


cms/src/enveloped_data.rs
```diff
@@ -51,6 +52,86 @@ pub struct EnvelopedData {
...
+            .map(|field| ::der::asn1::ContextSpecificRef {
+                tag_number: ::der::TagNumber::N1,
+                tag_mode: ::der::TagMode::Implicit,
+                value: field,
+            })
+            .encode(writer)?;
+        Ok(())
+    }
+}
+impl<'__der_lifetime> ::der::Sequence<'__der_lifetime> for Box<EnvelopedData> {}
```
tarcieri, line 133: It seems like this implementation should just delegate to the respective trait impls on `EnvelopedData`.

Also it seems like we should just add an attribute that says the field needs to be deref'd. Offhand I'm not sure how e.g. `serde` handles that case.

carl-wallace: Pushed a refactored manual impl. Much smaller now. Thanks.

cms/src/enveloped_data.rs
```diff
@@ -51,6 +52,24 @@ pub struct EnvelopedData {
     pub unprotected_attrs: Option<Attributes>,
 }
 
+impl<'__der_lifetime> ::der::DecodeValue<'__der_lifetime> for Box<EnvelopedData> {
```
tarcieri, line 55: Can you change the lifetimes to `'a` and add a TODO comment:

```suggestion
// TODO: replace these handwritten impls with `Box` support in the proc macro
impl<'a> ::der::DecodeValue<'a> for Box<EnvelopedData> {
```

carl-wallace: Done. There is a draft macro with box support and one struct changed to use it on the box branch of my fork.

tarcieri (approved):

PR #859, opened 2023-01-24
Add BoxedSequence macro and modify PopoSigningKey to use macro

carl-wallace:
if this look OK will change other manual Box impls to use macro


tarcieri:
I don't think we need to add explicit support to the proc macro, or make any changes to the proc macro whatsoever.

Instead I think we just need to write the appropriate thunks for the "smart pointer" types in alloc/std, which can be blanket impls of various traits.

Here's what `serde` does:

https://docs.rs/serde/latest/src/serde/ser/impls.rs.html#415-466

PR #861, opened 2023-01-25
CryptoPAN

mikemcdermottredjack:
Move from RustCrypto/utils #831

CryptoPAN is a prefix-preserving IP address anonymization algorithm.


cryptopan/src/lib.rs
```diff
@@ -0,0 +1,120 @@
...
+//! ```
+use aes::cipher::BlockEncryptMut;
+use aes::cipher::{generic_array::GenericArray, InvalidLength, KeyInit};
+use bitvec::field::BitField;
+use bitvec::prelude::Msb0;
+use bitvec::{bitvec, vec::BitVec};
+use std::net::{Ipv4Addr, Ipv6Addr};
+
+/// The CryptoPAN anonymizer
+pub struct CryptoPAn {
```
tarcieri, line 29: Per [RFC 430](https://rust-lang.github.io/api-guidelines/naming.html#casing-conforms-to-rfc-430-c-case) I think this should just be `CryptoPan`:

```suggestion
pub struct CryptoPan {
```

cryptopan/Cargo.toml
```diff
@@ -0,0 +1,15 @@
...
+license = "MIT OR Apache-2.0"
+description = "CryptoPAN algorithm for IP address anonymization"
+repository = "https://github.com/RustCrypto/utils"
+keywords = ["cryptopan", "ip anonymization"]
+categories = ["cryptography"]
+edition = "2021"
+readme = "README.md"
+
+[dependencies]
+aes="0.8.1"
```
newpavlov, line 14:
```suggestion
aes = "0.8.1"
```

cryptopan/README.md
```diff
@@ -0,0 +1,37 @@
...
+[license-image]: https://img.shields.io/badge/license-Apache2.0/MIT-blue.svg
+[rustc-image]: https://img.shields.io/badge/rustc-1.41+-blue.svg
+[chat-image]: https://img.shields.io/badge/zulip-join_chat-blue.svg
+[chat-link]: https://rustcrypto.zulipchat.com/#narrow/stream/260052-utils
+
+[//]: # (general links)
+
+[RustCrypto]: https://github.com/rustcrypto
+
+
```
newpavlov, line 37:
```suggestion

```

newpavlov (commented):
As suggested in the previous PR, I think it's worth to make `CryptoPan` generic over the `cipher` traits. The original paper (unsurprisingly) defines the algorithm in a generic fashion as well and even mentions AES ciphers with other key lengths. in addition to supporting other ciphers, such change would also allow to use reference to cipher (i.e. `&Aes128`) instead of owned instance.

I would also remove dependency on `bitvec`, but we can do it in a separate PR.

tarcieri:
@redjack FYI, we're still interested in this, but there was some feedback which it would be good to address. If you don't want to or don't have time (you'll also need to upgrade to our new release series) I can take care of that for you

PR #867, opened 2023-02-02
const-oid: Add OIDs from rfc8410.

flihp:
This includes: Ed25519, Ed448, X25519 and X448.

This RFC & the relevant OIDs aren't included in the IANA LDAP CSV. The full text of the RFC is included as an alternative.


tarcieri (approved):

tarcieri:
Thanks!

PR #876, opened 2023-02-22
Replace Read::read with read_exact

pcapriotti:
Read::read is not guaranteed to fill the whole output buffer.

### DIFF
--- a/tls_codec/src/tls_vec.rs
+++ b/tls_codec/src/tls_vec.rs
@@ -56,24 +56,8 @@ macro_rules! impl_byte_deserialize {
             let mut result = Self {
                 vec: vec![0u8; len],
             };
-            let read = bytes.read(result.vec.as_mut_slice())?;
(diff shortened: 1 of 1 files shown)


tls_codec/src/tls_vec.rs
```diff
@@ -56,24 +56,8 @@ macro_rules! impl_byte_deserialize {
...
-                    );
-                }
-                Err(Error::DecodingError(format!(
-                    "{} bytes were read but {} were expected",
-                    read, len
-                )))
-            } else {
-                Ok(result)
-            }
+            bytes.read_exact(result.vec.as_mut_slice())?;
```
franziskuskiefer, line 59: This looks like I was trying to do something specific here but I can't recall what that was and don't see a reason to keep it as is.

franziskuskiefer (approved):
Thanks.

PR #910, opened 2023-02-28
tls_codec: feat: Append a comment with the hex representation of `VLBytes` during `Debug`-printing.

duesee:
This PR makes it so that the `Debug`-print of `VLBytes` and `VLByteSlice` will have an additional comment `/* 0x... */` with the hex representation of the bytes.

Example: `Debug`-print of `VLBytes { vec: [0] }`:

```
VLBytes { vec: [0] /* 0x00 */ }

// "Pretty"
VLBytes {
    vec: [
        0,
    ] /* 0x00 */,
}
```

I decided against `[0x00, 0x01, ...]` because `0x0001` seems more useful to copy&paste (for use in hex editors, Python's binascii.unhexlify, JSON, ...). Also, having both representations at hand might be useful as well.

### DIFF
--- a/tls_codec/src/quic_vec.rs
+++ b/tls_codec/src/quic_vec.rs
@@ -12,7 +12,7 @@
 //! This is in contrast to the default behaviour defined by RFC 9000 that allows
 //! up to 62-bit length values.
 use alloc::vec::Vec;
-use std::fmt;
+use core::fmt;
 
 #[cfg(feature = "arbitrary")]
 use arbitrary::{Arbitrary, Unstructured};
--- a/tls_codec/src/quic_vec.rs
+++ b/tls_codec/src/quic_vec.rs
@@ -197,26 +197,17 @@ impl<T: Size> Size for &[T] {
 
 // === Vec<u8> and &[u8]
 
-/// Helper type used for `Debug`-printing.
-///
-/// Wrap a `&[u8]` with this type to append a comment with its hex representation
-/// during `Debug`-printing, i.e., `/* 0x... */`.
-struct SliceU8HexDebug<'a>(pub &'a [u8]);
-
-impl<'a> fmt::Debug for SliceU8HexDebug<'a> {
-    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
-        self.0.fmt(f)?;
-
-        if !self.0.is_empty() {
-            write!(f, " /* 0x")?;
-            for byte in self.0 {
-                write!(f, "{:02x}", byte)?;
-            }
-            write!(f, " */")?;
+fn write_hex(f: &mut fmt::Formatter<'_>, data: &[u8]) -> fmt::Result {
+    if !data.is_empty() {
+        write!(f, "0x")?;
+        for byte in data {
+            write!(f, "{:02x}", byte)?;
         }
-
-        Ok(())
+    } else {
+        write!(f, "b\"\"")?;
     }
+
+    Ok(())
 }
 
 /// Variable-length encoded byte vectors.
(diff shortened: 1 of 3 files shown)


tls_codec/src/quic_vec.rs
```diff
@@ -196,15 +197,45 @@ impl<T: Size> Size for &[T] {
...
+/// Helper type used for `Debug`-printing.
+///
+/// Wrap a `&[u8]` with this type to append a comment with its hex representation
+/// during `Debug`-printing, i.e., `/* 0x... */`.
+struct SliceU8HexDebug<'a>(pub &'a [u8]);
+
+impl<'a> fmt::Debug for SliceU8HexDebug<'a> {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        self.0.fmt(f)?;
+
```
franziskuskiefer, line 209: Is this really useful? This will get pretty messy for larger slices. I'd prefer skipping this.

tls_codec/src/quic_vec.rs
```diff
@@ -432,3 +471,32 @@ impl<'a> Arbitrary<'a> for VLBytes {
...
+            VLBytes::new(vec![0xAA]),
+            VLBytes::new(vec![0xFF]),
+            VLBytes::new(vec![0x00, 0x00]),
+            VLBytes::new(vec![0x00, 0xAA]),
+            VLBytes::new(vec![0x00, 0xFF]),
+            VLBytes::new(vec![0xff, 0xff]),
+        ];
+
+        for test in tests.into_iter() {
+            println!("# {:?}", test.vec);
```
franziskuskiefer, line 494: Can you make these actual tests checking the resulting string.

franziskuskiefer (changes requested):
Thanks, lgtm with a few nits (I think this is a little too noisy).

tls_codec/src/quic_vec.rs
```diff
@@ -196,15 +197,45 @@ impl<T: Size> Size for &[T] {
 
 // === Vec<u8> and &[u8]
 
+/// Helper type used for `Debug`-printing.
+///
+/// Wrap a `&[u8]` with this type to append a comment with its hex representation
+/// during `Debug`-printing, i.e., `/* 0x... */`.
```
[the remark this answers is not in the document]
duesee: Often, `Debug` output is almost valid Rust code allowing to instantiate types for replay. Using a comment preserves a bit of this property. The inline comment also makes it more clear that this is not a second value but only printed for readability. I can remove it if you feel strongly about it but would prefer to keep it.

tls_codec/src/quic_vec.rs
duesee: Could you elaborate? I'm not sure I understand what you propose to skip.

franziskuskiefer: You print things twice right now. And I don't see how that's useful.
The output is, as you also say in the description
```
VLBytes { vec: [0] /* 0x00 */ }
```
while
```
VLBytes { vec: 0x00 }
```
or even better
```
VLBytes: 0x00
```
would be sufficient and less cluttered.

franziskuskiefer (changes requested):

tls_codec/src/quic_vec.rs
duesee: Seeing this... let's go with the last option :-) We can revisit later when we think that having both representations -- decimal and hexadecimal -- would be useful.

duesee:
This is the new (checked) output:

```
# []
VLByteSlice { b"" }
VLBytes { b"" }

# [0]
VLByteSlice { 0x00 }
VLBytes { 0x00 }

# [170]
VLByteSlice { 0xaa }
VLBytes { 0xaa }

# [255]
VLByteSlice { 0xff }
VLBytes { 0xff }

# [0, 0]
VLByteSlice { 0x0000 }
VLBytes { 0x0000 }
```

I kept the braces and empty bytes slices are printed as `b""`.

duesee:
Pushed two more commits to fix Clippy lints. The last commit resolves all other tls_codec lints while the first one only resolves the warnings generated due to the PR changes. I could drop the last commit or quickly do a new PR.

franziskuskiefer (approved):
Thanks

PR #920, opened 2023-03-07
`der`: Make `zeroize`'s `alloc` feature conditional

nick-mobilecoin:
Previously `zeroize`'s `alloc` feature would be enabled if `zeroize` was enabled in `der`. Now `zeroize`'s `alloc` feature is only enabled if this `der`'s `alloc` feature is enabled.

---

We were hoping to use p256 in a no-alloc environment. Moving p256 from 0.12.0->0.13.0 results in `alloc` getting pulled in.

It looks like when `sec1` went from 0.3.0->0.7.1, https://github.com/RustCrypto/formats/compare/sec1/v0.3.0...sec1/v0.7.1#diff-fe7c93733c51ed30397c61754e6d22958291335b1d61470ce78edcdf49a09399R36 its `der` dependency now uses `zeroize`. The `zeroize` dependency in `der` was calling out `alloc`.

Looking through `der` and `sec1` it seems that `alloc` + `zerioze` is only needed for `der`'s [document.rs](https://github.com/RustCrypto/formats/blob/master/der/src/document.rs).

Making this proposed change may be a breaking change as `der` exposes `zeroize`, https://github.com/RustCrypto/formats/blob/master/der/src/lib.rs#L394. If someone was using the Vec, String, etc features exposed via `der::zeroize` and the `alloc` feature.

### DIFF
--- a/der/Cargo.toml
+++ b/der/Cargo.toml
@@ -22,14 +22,14 @@ der_derive = { version = "0.7", optional = true, path = "derive" }
 flagset = { version = "0.4.3", optional = true }
 pem-rfc7468 = { version = "0.7", optional = true, features = ["alloc"], path = "../pem-rfc7468" }
 time = { version = "0.3.4", optional = true, default-features = false }
-zeroize = { version = "1.5", optional = true, default-features = false, features = ["alloc"] }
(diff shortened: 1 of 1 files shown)


nick-mobilecoin:
Thanks so much!!

PR #923, opened 2023-03-09
tls_codec: feat: Allow const exprs in discriminant attrs

OtaK:
You can now provide paths to `#[tls_codec(discriminant = ...)]`, like so `#[tls_codec(discriminant ="path::to::const::or::enum::Variant")]`

I'm going to do a pass on documenting the new behavior and will undraft this PR.

Changes: 

* Allowed `PathExpr`s in `#[tls_codec(discriminant = ...)]` attributes that resolve to constants or enum variants that can be coerced to the enum's repr.
* Fixed the proc macros panicking instead of returning a compile time error

### DIFF
--- a/tls_codec/derive/src/lib.rs
+++ b/tls_codec/derive/src/lib.rs
@@ -54,6 +54,7 @@
 //!
 //! ```text
 //! #[tls_codec(discriminant = 123)]
+//! #[tls_codec(discriminant = "path::to::const::or::enum::Variant")]
 //! ```
 //!
 //! This attribute may be applied to an enum variant to specify the discriminant to use when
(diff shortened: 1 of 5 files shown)


tls_codec/derive/src/lib.rs
```diff
@@ -477,17 +529,45 @@ fn define_discriminant_constants(
...
+                    DiscriminantValue::Literal(value_u32) => {
+                        implicit_discriminant = value_u32;
+                        quote! {
+                            const #constant_id: #repr = #value_u32 as #repr;
+                        }
+                    }
+                    DiscriminantValue::Path(pathexpr) => {
+                        discriminant_has_paths = true;
+                        quote! {
+                            const #constant_id: #repr = #pathexpr as #repr;
```
franziskuskiefer, line 549: You are forcing the path value into the representation of the enum. But this will lead to unexpected behaviour.
Taking the test in `encode.rs` for example you you can change `Potato = 0x0004,` to `Potato = 0x1000_0004`, i.e. u32, which is forced into `u16`. So
```rust
    let serialized = EnumWithDataAndConstDiscriminant::B
        .tls_serialize_detached()
        .unwrap();
    assert_eq!(vec![0, 4], serialized);
```
still holds, while I'd expect a compiler error that tells me that `Potato` is `u32` but `u16` was expected.

OtaK: I could just remove the `as #repr` part out of `#pathexpr as #repr` so that the compiler would error out because types differ instead of silently casting integers for convenience. Would that sound fine to you?

OtaK:
> Here I'd like to be able to add something like` #[tls_codec(discriminant = "ValueType"]` to the `Values` enum to map all the values at once.
> 
> Ideally we could do both. But we can get this one in first when it's fixed.

I thought about doing this, but for this one we could imagine an enum-level attr instead of variant-level; But this is much more involved as it would require deriving a trait on the "raw" enum to emit metadata at proc-macro time and another derive on the target enum to catch that metadata and copy value literals. 

I initially intended to basically have `Value` (the enum holding other information) emit `ValueType` (the enum holding raw discriminants) through the derive but hygiene-wise it's not so great as well. So I settled on the middleground that is easy to implement but still yields great usability/maintenability gains :)

tls_codec/derive/src/lib.rs
OtaK: After checking with a u64::MAX casted on a repr(u8) enum, you're absolutely right, this is not only unexpected behaviour but also Undefined Behaviour as well. Good catch, that's 100% a bug.

OtaK: By the way, the same issue exists with explicit literal discriminants as they are converted to `u32` first and *then* coerced into the enum's `repr`.

tls_codec/derive/src/lib.rs
```diff
@@ -64,18 +65,48 @@
 //! For enumerations with non-unit variants, if no variant has this attribute, the serialization
 //! discriminants will start from zero. If this attribute is used on a variant and the following
 //! variant does not have it, its discriminant will be equal to the previous variant discriminant
-//! plus 1.
+//! plus 1. This behavior is refered to as "implicit discriminants".
```
franziskuskiefer, line 68:
```suggestion
//! plus 1. This behavior is referred to as "implicit discriminants".
```

franziskuskiefer (approved):
Thanks for fixing!
I think this should work for now.

PR #946, opened 2023-03-23
Updates to syn-2

maurer:
Additionally has a commit with some clippy fixes, I can drop it if you don't like them.

### DIFF
--- a/const-oid/oiddbgen/Cargo.toml
+++ b/const-oid/oiddbgen/Cargo.toml
@@ -8,7 +8,7 @@ convert_case = "0.5.0"
 proc-macro2 = "1.0.36"
 quote = "1.0.15"
 regex = "1.5.5"
-syn = "1.0.86"
+syn = "2"
 
 [workspace]
 members = ["."]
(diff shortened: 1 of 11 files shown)


franziskuskiefer (approved):
Thanks!
Maybe someone should double-check the der crate (I'm not too familiar with the code). But everything else looks fine to me.

const-oid/oiddbgen/Cargo.toml
```diff
@@ -8,7 +8,7 @@ convert_case = "0.5.0"
 proc-macro2 = "1.0.36"
 quote = "1.0.15"
 regex = "1.5.5"
-syn = "1.0.86"
+syn = "2.0.6"
```
tarcieri, line 11: Is this the minimum compatible version?

I ask because if it's actually `2` I'd prefer to use that instead. We run `minimum-version` checks, and including an explicit minor version like that triggers Dependabot runs.

maurer: 2 works fine, I've relaxed it to 2.

tarcieri:
Thank you!

PR #949, opened 2023-03-25
x509-cert: add `FromStr` impls for name-related functionality

tarcieri:
Adds `FromStr` impls for the following types:

- `RdnSequence`
- `RelativeDistingiushedName`
- `AttributeTypeAndValue`

All of these types have `encode_from_string` methods which would parse the string and serialize the resulting type as a DER byte vector.

The `encode_from_string` methods now call the relevant `FromStr` impls.


npmccallum (approved):
Seems same. I don’t know why we didn’t implement it that way to begin with…

PR #953, opened 2023-03-27
pkcs1: export RSAPSS_SALT_LEN_DEFAULT

lumag:
Export default RSA PSS salt length.

### DIFF
--- a/pkcs1/src/params.rs
+++ b/pkcs1/src/params.rs
@@ -17,8 +17,6 @@ const SHA_1_AI: AlgorithmIdentifierRef<'_> = AlgorithmIdentifierRef {
     parameters: None,
 };
 
-const SALT_LEN_DEFAULT: u8 = 20;
-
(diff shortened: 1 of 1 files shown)


pkcs1/src/params.rs
```diff
@@ -17,7 +17,8 @@ const SHA_1_AI: AlgorithmIdentifierRef<'_> = AlgorithmIdentifierRef {
     parameters: None,
 };
 
-const SALT_LEN_DEFAULT: u8 = 20;
+/// Default RSA PSS Salt length in RsaPssParams
+pub const RSAPSS_SALT_LEN_DEFAULT: u8 = 20;
```
tarcieri, line 21: Maybe use an inherent constant like `RsaPssParams::SALT_LEN_DEFAULT`?

lumag:
ack, sounds good

pkcs1/src/params.rs
lumag: Done

tarcieri:
Thanks!

PR #954, opened 2023-03-27
spki: define trait to get associated AlgorithmIdentifier

lumag:
It is useful to get AlgorithmIdentifiers for defined keys. Add a trait implementing that functionality.

### DIFF
--- a/spki/src/traits.rs
+++ b/spki/src/traits.rs
@@ -3,6 +3,9 @@
 use crate::{Error, Result, SubjectPublicKeyInfoRef};
 
 #[cfg(feature = "alloc")]
+use crate::AlgorithmIdentifierOwned;
+
+#[cfg(feature = "alloc")]
 use der::Document;
 
 #[cfg(feature = "pem")]
@@ -92,4 +95,12 @@ pub trait EncodePublicKey {
         let doc = self.to_public_key_der()?;
         Ok(doc.write_pem_file(path, SubjectPublicKeyInfoRef::PEM_LABEL, line_ending)?)
     }
+}
+
+/// Returns AlgorithmIndentifier associated with the structure. This is mostly useful for Signing
+/// and Validation keys.
+#[cfg(feature = "alloc")]
+pub trait DynAssociatedAlgorithmIdentifier {
+    /// Returns corresponding AlgorithmIndentifier
+    fn algorithm_identifier(&self) -> AlgorithmIdentifierOwned;
 }
--- a/spki/src/lib.rs
+++ b/spki/src/lib.rs
@@ -57,8 +57,9 @@ pub use der::{self, asn1::ObjectIdentifier};
 #[cfg(feature = "alloc")]
 pub use {
     crate::{
-        algorithm::AlgorithmIdentifierOwned, spki::SubjectPublicKeyInfoOwned,
-        traits::EncodePublicKey,
+        algorithm::AlgorithmIdentifierOwned,
+        spki::SubjectPublicKeyInfoOwned,
+        traits::{DynAssociatedAlgorithmIdentifier, EncodePublicKey},
     },
     der::Document,
 };


spki/src/traits.rs
```diff
@@ -93,3 +96,11 @@ pub trait EncodePublicKey {
         Ok(doc.write_pem_file(path, SubjectPublicKeyInfoRef::PEM_LABEL, line_ending)?)
     }
 }
+
+/// Returns AlgorithmIndentifier associated with the structure. This is mostly useful for Signing
+/// and Validation keys.
+#[cfg(feature = "alloc")]
+pub trait AssociatedAlgorithmIdentifier {
+    /// Returns corresponding AlgorithmIndentifier
+    fn get_algo_id(&self) -> AlgorithmIdentifierOwned;
```
tarcieri, line 105: `const_oid::AssociatedOid` uses an associated const.

I think both patterns are useful, although naming them seems hard. I was thinking `DynAssociatedOid` for `const_oid`.

There can be a blanket impl which connects them too.

lumag: Associated const is not really possible since for RSA PSS we have to take `salt_length` into account (which is a `SigningKey` property).

lumag: @tarcieri see https://github.com/RustCrypto/RSA/pull/278

tarcieri: I was saying that there are two patterns that are useful: this one, which takes `&self` and is therefore object safe, and a pattern that uses an associated const.

Though an associated const doesn't work for that, there are many places where `AlgorithmIdentifier` is effectively static, e.g. just a single OID or OID 2-tuple, for example all of the elliptic curve `AlgorithmIdentifier`s.

spki/src/traits.rs
```diff
@@ -93,3 +96,11 @@ pub trait EncodePublicKey {
         Ok(doc.write_pem_file(path, SubjectPublicKeyInfoRef::PEM_LABEL, line_ending)?)
     }
 }
+
+/// Returns AlgorithmIndentifier associated with the structure. This is mostly useful for Signing
+/// and Validation keys.
+#[cfg(feature = "alloc")]
+pub trait AssociatedAlgorithmIdentifier {
```
tarcieri, line 103: As a more specific recommendation:

```suggestion
pub trait DynAssociatedAlgorithmIdentifier {
```

...leaving `AssociatedAlgorithmIdentifier` for a trait which uses an associated constant.

spki/src/traits.rs
tarcieri, line 105: Also while I'm here, how about this name for this method:

```suggestion
    fn algorithm_identifier(&self) -> AlgorithmIdentifierOwned;
```

spki/src/traits.rs
lumag: ack, pushing an updated version

lumag:
I marked this as a draft too. Let's settle on the implementation details first (including ecdsa crates).

spki/src/traits.rs
```diff
@@ -93,3 +96,11 @@ pub trait EncodePublicKey {
...
     }
 }
+
+/// Returns AlgorithmIndentifier associated with the structure. This is mostly useful for Signing
+/// and Validation keys.
+#[cfg(feature = "alloc")]
+pub trait DynAssociatedAlgorithmIdentifier {
+    /// Returns corresponding AlgorithmIndentifier
+    fn algorithm_identifier(&self) -> AlgorithmIdentifierOwned;
+}
```
tarcieri, line 106: As an alternative to being `alloc`-gated, it could potentially take params generically ala `AlgorithmIdentifier`:

```suggestion
pub trait DynAssociatedAlgorithmIdentifier<Params> {
    /// Returns corresponding AlgorithmIndentifier
    fn algorithm_identifier(&self) -> AlgorithmIdentifier<Params>;
}
```

This is still object safe, and allows the trait to work without `alloc`.

Though there is a convenience argument for the current implementation.

tarcieri: This approach would probably simplify writing a blanket impl for a prospective `AssociatedAlgorithmIdentifier` trait that uses an associated constant.

lumag: I settled on `Owned` because I needed this crate to fill out the `Certificate` fields (which are `Owned` nowadays). I fear that defining it with `<Params>` would result either in duplication or in some ugly type conversions.

tarcieri: You'd have to use `DynAssociatedAlgorithmIdentifier<Any>` to get `AlgorithmIdentifierOwned`

tarcieri: Okay, let's start with this and I'll try to add a corresponding `AssociatedAlgorithmIdentifier` trait and an blanket impl.

lumag: @tarcieri I think I'll be able to use associated const implementation to simplify ECDSA support. Internally I developed this as a separate source module, so there were no locality issues. And splitting things to proper source files brings some interesting problems.

tarcieri: Okay, if you'd prefer to define that trait that's fine too

tarcieri:
Merged as #962

PR #962, opened 2023-03-31
spki: define trait to get associated `AlgorithmIdentifier`

tarcieri:
This is an alternative to/continuation of #954, which defines both `AssociatedAlgorithmIdentifier` and `DynAssociatedAlgorithmIdentifier`, where the former provides an associated constant `AlgorithmIdentifierRef`, and the latter a trait/object-safe way to obtain an `AlgorithmIdentifierOwned` which can be computed from `&self`.

A blanket impl links the two.


tarcieri:
cc @lumag

PR #966, opened 2023-04-02
spki: add `AssociatedAlgorithmIdentifier::Params` type

tarcieri:
Adds an associated `Params` type which is passed as `AlgorithmIdentifier<Params>`, which makes it *much* easier to define `AlgorithmIdentifier` constants.

Without this, `const fn` type conversions are required to `AnyRef<'static>`. This isn't currently possible for `ObjectIdentifier` for example.

cc @baloo @lumag

### DIFF
--- a/spki/src/traits.rs
+++ b/spki/src/traits.rs
@@ -1,11 +1,11 @@
 //! Traits for encoding/decoding SPKI public keys.
 
-use crate::{AlgorithmIdentifierRef, Error, Result, SubjectPublicKeyInfoRef};
+use crate::{AlgorithmIdentifier, Error, Result, SubjectPublicKeyInfoRef};
 
 #[cfg(feature = "alloc")]
 use {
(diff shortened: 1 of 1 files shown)


spki/src/traits.rs
```diff
@@ -115,8 +118,15 @@ pub trait DynAssociatedAlgorithmIdentifier {
 }
 
 #[cfg(feature = "alloc")]
-impl<T: AssociatedAlgorithmIdentifier> DynAssociatedAlgorithmIdentifier for T {
+impl<T> DynAssociatedAlgorithmIdentifier for T
+where
+    T: AssociatedAlgorithmIdentifier,
+    T::Params: Into<Any>,
```
tarcieri, line 124: Debatably we could get rid of this bound by using `Any::encode_msg`, however that's fallible so we'd have to add an `expect` and panic on encoding failure.

PR #967, opened 2023-04-02
spki: add (Dyn)SignatureAlgorithmIdentifier traits

lumag:
Add SignatureAlgorithmIdentifier and DynSignatureAlgorithmIdentifier which are useful for specifying the Signature AlgorithmIdentifier for Signing/Verifying keys.


spki/src/traits.rs
```diff
@@ -120,3 +120,29 @@ impl<T: AssociatedAlgorithmIdentifier> DynAssociatedAlgorithmIdentifier for T {
         Self::ALGORITHM_IDENTIFIER.ref_to_owned()
     }
 }
+
+/// Returns `AlgorithmIdentifier` associated with the signature system.
+///
+/// Unlike AssociatedAlgorithmIdentifier this is intended to be implemented for public and/or
+/// private keys.
+pub trait SignatureAlgorithmIdentifier {
```
tarcieri, line 128: This should probably be parameterized by `Params` ala #966, otherwise it's difficult to define constants

lumag: done

spki/src/traits.rs
```diff
@@ -130,3 +130,39 @@ where
...
+    /// Algorithm parameters.
+    type Params: der::Encode;
+
+    /// `AlgorithmIdentifier` for the corresponding singature system.
+    const ALGORITHM_IDENTIFIER: AlgorithmIdentifier<Self::Params>;
```
tarcieri, line 143: This is ambiguous with `AssociatedAlgorithmIdentifier`. Perhaps this?
```suggestion
    const SIGNATURE_ALGORITHM_IDENTIFIER: AlgorithmIdentifier<Self::Params>;
```

spki/src/traits.rs
```diff
@@ -130,3 +130,39 @@ where
...
+}
+
+/// Returns `AlgorithmIdentifier` associated with the signature system.
+///
+/// Unlike AssociatedAlgorithmIdentifier this is intended to be implemented for public and/or
+/// private keys.
+#[cfg(feature = "alloc")]
+pub trait DynSignatureAlgorithmIdentifier {
+    /// `AlgorithmIdentifier` for the corresponding singature system.
+    fn algorithm_identifier(&self) -> AlgorithmIdentifierOwned;
```
tarcieri, line 153:
```suggestion
    fn signature_algorithm_identifier(&self) -> AlgorithmIdentifierOwned;
```

spki/src/traits.rs
lumag: Ack, I committed both suggestions.

tarcieri (approved):

PR #974, opened 2023-04-04
spki: use Result for Dyn*AlgorithmIdentifier traits

lumag:
There is a possibility that creating the AlgorithmIdentifier results in the Error. Change corresponding methods to return
Result<AlgorithmIdentifier>.

### DIFF
--- a/spki/src/traits.rs
+++ b/spki/src/traits.rs
@@ -5,7 +5,7 @@ use crate::{AlgorithmIdentifier, Error, Result, SubjectPublicKeyInfoRef};
 #[cfg(feature = "alloc")]
 use {
     crate::AlgorithmIdentifierOwned,
-    der::{Any, Document},
+    der::{Any, Document, EncodeValue, Tagged},
 };
 
 #[cfg(feature = "pem")]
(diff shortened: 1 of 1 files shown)


tarcieri (approved):
I'm fine with this. It would make the blanket impl more flexible, since we can use the `Encode` trait to build the `Any` value for `Params` (though that needs a bit more work).

@baloo are you okay with this?

spki/src/traits.rs
```diff
@@ -123,11 +123,11 @@ where
     T: AssociatedAlgorithmIdentifier,
     T::Params: Into<Any>,
 {
-    fn algorithm_identifier(&self) -> AlgorithmIdentifierOwned {
-        AlgorithmIdentifierOwned {
+    fn algorithm_identifier(&self) -> Result<AlgorithmIdentifierOwned> {
+        Ok(AlgorithmIdentifierOwned {
             oid: T::ALGORITHM_IDENTIFIER.oid,
             parameters: T::ALGORITHM_IDENTIFIER.parameters.map(Into::into),
```
tarcieri, line 129: With #976 landed you can get rid of the `T::Params: Into<Any>` bound and instead do:

```suggestion
            parameters: T::ALGORITHM_IDENTIFIER.parameters.as_ref().map(Any::encode_from).transpose()?,
```

spki/src/traits.rs
```diff
@@ -159,10 +159,10 @@ where
     T: SignatureAlgorithmIdentifier,
     T::Params: Into<Any>,
 {
-    fn signature_algorithm_identifier(&self) -> AlgorithmIdentifierOwned {
-        AlgorithmIdentifierOwned {
+    fn signature_algorithm_identifier(&self) -> Result<AlgorithmIdentifierOwned> {
+        Ok(AlgorithmIdentifierOwned {
             oid: T::SIGNATURE_ALGORITHM_IDENTIFIER.oid,
             parameters: T::SIGNATURE_ALGORITHM_IDENTIFIER.parameters.map(Into::into),
```
tarcieri, line 165: Same here:


```suggestion
            parameters: T::SIGNATURE_ALGORITHM_IDENTIFIER.parameters.as_ref().map(Any::encode_from),
```

spki/src/traits.rs
lumag: Implemented in a slightly different way.

spki/src/traits.rs
lumag: same

spki/src/traits.rs
tarcieri: That seems way more verbose, but I guess it gets the job done

lumag: Maybe it's a matter of preference of simple `if let` compared to long chains of function calls.

tarcieri: I imagine clippy will complain about it at some point in the future: https://github.com/rust-lang/rust-clippy/issues/4171

I'd consider the functional approach a bit more idiomatic.

lumag: that lint talks about iterables. If you prefer, I can use your form, but IMHO it is less clear.

tarcieri: If nothing else it's more consistent with the other uses of `transpose()` in this repo

lumag: done

tarcieri (approved):
Still curious to hear from @baloo

lumag:
> Still curious to hear from @baloo

Sure, me too.

baloo:
Yeah that should work for me.

PR #975, opened 2023-04-04
der: support converting a Sequence into Any

lumag:
It is useful to be able to convert a Tag+Value into Any. However directly implementing the TryFrom trait is not possible because of a clash with the default blanked implementation. Use special AnyWrapper newtype to handle such conversions.


tarcieri:
There's a much better API possible for this which doesn't require introducing another type. I can open a PR.

lumag:
yes, your proposal is definitely better.

PR #976, opened 2023-04-04
der: add `Any::encode_from`

tarcieri:
Support for encoding an `Any` value from any type which impls `Tagged` + `EncodeValue`.


tarcieri:
This could probably use a test

PR #988, opened 2023-04-05
x509-cert: use `ErrorKind::Value` for overlength serial

tarcieri:
This at least surfaces that the error has to do with an integer field, which would've been helpful with tracing it down to SerialNumber.

`Overlength` has a bit of a specific meaning in the context of the `ErrorKind`, namely it's used for framing errors, which isn't what's occurring here.

Perhaps the `der` documentation should help clarify this.


lumag:
thanks, it is now much more clear. Otherwise I assumer a lowlevel DER parsing issue.

PR #991, opened 2023-04-06
Creating PKCS #7 messages

bkstein:
This is a proposal (say: a trial ballon) for how PKSC #7 messages could be created. The work is not done with this request. Signing and encryption is missing, but by making some types public, we can at least get a PKSC #7 scaffolding with certificates.
Is this a good way? Should I proceed with it? Or do you already have other plans?

### DIFF
--- a/pkcs7/Cargo.toml
+++ b/pkcs7/Cargo.toml
@@ -17,7 +17,8 @@ rust-version = "1.65"
 [dependencies]
 der = { version = "0.7", features = ["oid"], path = "../der" }
 spki = { version = "0.7", path = "../spki" }
-x509-cert = { version = "0.2", default-features = false, path = "../x509-cert" }
+# x509-cert = { version = "0.2", default-features = false, path = "../x509-cert" }
+x509-cert = { version = "0.2", path = "../x509-cert" }
 
 [dev-dependencies]
 hex-literal = "0.3"
--- a/pkcs7/Cargo.toml
+++ b/pkcs7/Cargo.toml
@@ -20,9 +20,9 @@ spki = { version = "0.7" }
 x509-cert = { version = "0.2", default-features = false }
 
 [dev-dependencies]
-der = { version = "0.7", features = ["oid", "pem"], path = "../der" }
+der = { version = "0.7", features = ["oid", "pem"] }
 hex-literal = "0.4"
-x509-cert = { version = "0.2", default-features = false, features = ["pem"], path = "../x509-cert" }
+x509-cert = { version = "0.2", default-features = false, features = ["pem"] }
 
 [package.metadata.docs.rs]
 all-features = true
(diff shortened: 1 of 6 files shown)


bkstein:
Oh oh, looks like I activated std. I'll check that.

pkcs7/Cargo.toml
```diff
@@ -17,7 +17,8 @@ rust-version = "1.65"
 [dependencies]
 der = { version = "0.7", features = ["oid"], path = "../der" }
 spki = { version = "0.7", path = "../spki" }
-x509-cert = { version = "0.2", default-features = false, path = "../x509-cert" }
+# x509-cert = { version = "0.2", default-features = false, path = "../x509-cert" }
+x509-cert = { version = "0.2", path = "../x509-cert" }
```
tarcieri, line 21: It's this change that's activating `std`.

Instead you should activate the specific features you need.

bkstein: Thank's. I activated PEM handling for the added test to see, if I can add certificates (using existing example cert.pem). CI is green now.

pkcs7/tests/content_tests.rs
```diff
@@ -139,6 +143,10 @@ fn decode_signed_scep_example() {
         }
         _ => panic!("expected ContentInfo::SignedData(Some(_))"),
     }
+
+    let mut buf = vec![0u8; bytes.len()];
+    let encoded_content = encode_content_info(&content, &mut buf);
+    println!("{:?}", encoded_content);
```
tarcieri, line 149: Looks like you might've been doing some debugging here?

tarcieri:
Looks good overall. One nit about what seems like it was leftover debugging code.

pkcs7/Cargo.toml
```diff
@@ -20,7 +20,9 @@ spki = { version = "0.7" }
 x509-cert = { version = "0.2", default-features = false }
 
 [dev-dependencies]
+der = { version = "0.7", features = ["oid", "pem"], path = "../der" }
 hex-literal = "0.4"
+x509-cert = { version = "0.2", default-features = false, features = ["pem"], path = "../x509-cert" }
```
tarcieri, line 25: After #997 this needs to be:

```suggestion
der = { version = "0.7", features = ["oid", "pem"] }
hex-literal = "0.4"
x509-cert = { version = "0.2", default-features = false, features = ["pem"] }
```

pkcs7/tests/content_tests.rs
```diff
@@ -139,6 +143,9 @@ fn decode_signed_scep_example() {
         }
         _ => panic!("expected ContentInfo::SignedData(Some(_))"),
     }
+
+    let mut buf = vec![0u8; bytes.len()];
+    let encoded_content = encode_content_info(&content, &mut buf);
```
tarcieri, line 148: It's complaining about the unused variable:

```suggestion
    encode_content_info(&content, &mut buf);
```

PR #993, opened 2023-04-06
x509-cert: fix `KeyUsage` bit tests

### LINKED ISSUE #992: x509-cert: incorrect bit tests for KeyUsage
The recently added bit tests method on `KeyUsage` doesn't give the expected result.

# Reproduction

```rust
let key_usage = KeyUsage(KeyUsages::KeyEncipherment.into());

println!("{}", key_usage.digital_signature());
println!("{}", key_usage.0.contains(KeyUsages::DigitalSignature));
```

## Output

```
true
false
```

## Expected

```
false
false
```

This is due to the usage of `KeyUsages::X as u16` which is not the same as the flag value. I'm preparing a PR. Would you like to keep the same API or have a wrapper function `fn contains(&self, rhs: impl Into<FlagSet<KeyUsages>>)` or maybe both ?

stormshield-guillaumed:
closes #992

### DIFF
--- a/x509-cert/src/ext/pkix/keyusage.rs
+++ b/x509-cert/src/ext/pkix/keyusage.rs
@@ -56,43 +56,43 @@ impl_newtype!(KeyUsage, FlagSet<KeyUsages>);
 impl KeyUsage {
     /// The subject public key is used for verifying digital signatures
     pub fn digital_signature(&self) -> bool {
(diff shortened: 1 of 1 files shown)


tarcieri:
cc @Xynnn007

tarcieri (approved):

PR #1009, opened 2023-04-18
der: add `Decode` and `Encode` impls for `PhantomData`

tarcieri:
Allows `PhantomData` fields to be used with custom derive.

cc @baloo


der/tests/derive.rs
```diff
@@ -266,6 +267,9 @@ mod sequence {
             tag_mode = "IMPLICIT"
         )]
         pub only_contains_attribute_certs: bool,
+
+        /// Test handling of PhantomData.
+        pub phantom: PhantomData<()>,
```
tarcieri, line 272: Added this as a spot check. It seems to work?

PR #1013, opened 2023-04-19
Fix feature `alloc` and `pem` in `pkcs1`

NobodyXu:
Avoid pulling in dep `pkcs8` or `zeroize` just because feature `alloc` or `pem` is enabled.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -977,7 +977,6 @@ dependencies = [
  "pkcs8",
  "spki",
  "tempfile",
- "zeroize",
 ]
 
 [[package]]
--- a/pkcs1/Cargo.toml
+++ b/pkcs1/Cargo.toml
@@ -20,7 +20,6 @@ spki = { version = "0.7" }
 
 # optional dependencies
 pkcs8 = { version = "0.10", optional = true, default-features = false }
-zeroize = { version = "1", optional = true, default-features = false }
 
 [dev-dependencies]
 const-oid = { version = "0.9", features = ["db"] } # TODO: path = "../const-oid"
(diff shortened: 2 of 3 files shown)


tarcieri:
This is technically a breaking change, although probably an OK one.

You'll need to update the import gating accordingly, however, since the tests are now failing.

tarcieri:
(whoops sorry, didn't mean to close it)

NobodyXu:
@tarcieri Thanks, I've fixed the error and I believe this PR is backwards compatible, since
 - no feature is removed
 - no API change under cfg

NobodyXu:
Sorry that somehow I did not receive any notification for your reply.

NobodyXu:
I've also run `cargo semver-checks check-release` and it confirms that this PR does not include any breaking change.

NobodyXu:
@tarcieri All checks have passed, this PR is ready for another review.

tarcieri:
@NobodyXu the breaking change would be features being implicitly enabled before and not any longer. That's not going to be caught by a tool like `cargo semver-checks`.

But it's minor enough it's probably a non-issue.

NobodyXu:
> @NobodyXu the breaking change would be features being implicitly enabled before and not any longer. That's not going to be caught by a tool like `cargo semver-checks`.
> 
> But it's minor enough it's probably a non-issue.

Yeah you are right, I kind of ignore that because I thought it was wrong to enable `pkcs8` in feature `alloc`.
In fact that was one of the motivation for this PR since this bumps the msrv to 1.65.0 in sccache https://github.com/mozilla/sccache/pull/1742

@tarcieri Can you make a new release for `pkcs1`?

PR #1029, opened 2023-04-27
Derive Zeroize and ZeroizeOnDrop for VLBytes

kkohbrok:
In contrast to the generic tls_vec implementations, there is no `Secret...` variant for VLBytes. Since the overhead is relatively low, this PR blanket-implements ZeriozeOnDrop for VLBytes. In the future, it might be worth exposing a dedicated `SecretVLBytes` struct.

### DIFF
--- a/tls_codec/Cargo.toml
+++ b/tls_codec/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "tls_codec"
-version = "0.3.0-pre.2"
+version = "0.3.0-pre.3"
 authors = ["RustCrypto Developers"]
 license = "Apache-2.0 OR MIT"
 documentation = "https://docs.rs/tls_codec/"
--- a/tls_codec/Cargo.toml
+++ b/tls_codec/Cargo.toml
@@ -11,7 +11,7 @@ edition = "2021"
 rust-version = "1.60"
 
 [dependencies]
-zeroize = { version = "1", default-features = false, features = ["alloc"] }
+zeroize = { version = "1", default-features = false, features = ["alloc", "zeroize_derive"] }
 
 # optional dependencies
 arbitrary = { version = "1", features = ["derive"], optional = true }
--- a/tls_codec/Cargo.toml
+++ b/tls_codec/Cargo.toml
@@ -11,7 +11,7 @@ edition = "2021"
 rust-version = "1.60"
 
 [dependencies]
-zeroize = { version = "1", default-features = false, features = ["alloc", "zeroize_derive"] }
+zeroize = { version = "1.6", default-features = false, features = ["alloc", "zeroize_derive"] }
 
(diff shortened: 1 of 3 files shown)


tls_codec/src/quic_vec.rs
```diff
@@ -396,6 +397,94 @@ impl Size for &VLBytes {
...
 }
 
+/// A wrapper struct around [`VLBytes`] that implements [`ZeroizeOnDrop`]. It
+/// behaves just like [`VLBytes`], except that it doesn't allow conversion into
+/// a [`Vec<u8>`].
+#[cfg_attr(feature = "serde", derive(SerdeSerialize, SerdeDeserialize))]
+#[derive(Clone, PartialEq, Eq, Hash, Ord, PartialOrd, Zeroize, ZeroizeOnDrop)]
+pub struct SecretVLBytes(VLBytes);
+
+impl fmt::Debug for SecretVLBytes {
```
franziskuskiefer, line 407: Can you throw all of the implementation (that makes sense) into a macro so we don't need to duplicate the code?

tls_codec/src/quic_vec.rs
```diff
@@ -214,7 +215,7 @@ fn write_hex(f: &mut fmt::Formatter<'_>, data: &[u8]) -> fmt::Result {
 /// Use this struct if bytes are encoded.
 /// This is faster than the generic version.
 #[cfg_attr(feature = "serde", derive(SerdeSerialize, SerdeDeserialize))]
-#[derive(Clone, PartialEq, Eq, Hash, Ord, PartialOrd)]
+#[derive(Clone, PartialEq, Eq, Hash, Ord, PartialOrd, Zeroize)]
```
franziskuskiefer, line 218: So only zeroize doesn't make a performance difference?

tls_codec/Cargo.toml
```diff
@@ -11,7 +11,7 @@ edition = "2021"
 rust-version = "1.60"
 
 [dependencies]
-zeroize = { version = "1", default-features = false, features = ["alloc"] }
+zeroize = { version = "1", default-features = false, features = ["alloc", "zeroize_derive"] }
```
franziskuskiefer, line 14: Can you bump the version to `-pre.3`? Then we can release it right away.
I'd like to release this as `0.3` then next week.

tls_codec/src/quic_vec.rs
kkohbrok: Benching locally I don't see a meaningful difference. The way I understand it, deriving `Zeroize` only gives you the ability to zeroize manually. Automatically zeroizing on drop seems to cause performance issues.

franziskuskiefer (commented):
lgtm but looks like there's an issue with the msrv on ci

Issue #1031, opened 2023-04-28
x509-cert: Generate certificate signing request?


jstayton:
👋🏻 Thanks for the great work you're doing here!

I'm wondering if the `x509-cert` crate is able to generate a certificate signing request at this point? I know development is pretty early on it. The [tests](https://github.com/RustCrypto/formats/blob/master/x509-cert/tests/certreq.rs) seem to be focused on reading/modeling an existing CSR.

Any pointers/examples would be much appreciated. Thanks!

baloo:
Hey! thanks for raising the issue.
Certificate requests were not on my mind when I pushed the initial draft of the certificate builder. A request builder is not too hard to add, although I needed to redesign a bit the `AsExtension` trait.
I sent a rough draft on https://github.com/RustCrypto/formats/pull/1034, would love to get your feedback and opinion.

jstayton:
Hey @baloo – Wow, that's great news! I appreciate your work.

I'm new to Rust, so I could definitely be missing something obvious here, but when I try to run your example/test code in  a new crate, I get this:

```
error[E0277]: the trait bound `sha2::OidSha256: AssociatedOid` is not satisfied
   --> src/main.rs:26:52
    |
26  |     let mut builder = RequestBuilder::new(subject, &signer).unwrap();
    |                       -------------------          ^^^^^^^ the trait `AssociatedOid` is not implemented for `sha2::OidSha256`
    |                       |
    |                       required by a bound introduced by this call
    |
    = help: the following other types implement trait `AssociatedOid`:
              AuthorityInfoAccessSyntax
              AuthorityKeyIdentifier
              BaseCrlNumber
              BasicConstraints
              CertificatePolicies
              CoreWrapper<T>
              CrlDistributionPoints
              CrlNumber
            and 19 others
    = note: required for `CtVariableCoreWrapper<sha2::core_api::Sha256VarCore, UInt<UInt<UInt<UInt<UInt<UInt<UTerm, B1>, B0>, B0>, B0>, B0>, B0>, sha2::OidSha256>` to implement `AssociatedOid`
    = note: 2 redundant requirements hidden
    = note: required for `ecdsa::Signature<NistP256>` to implement `AssociatedOid`
    = note: required for `ecdsa::Signature<NistP256>` to implement `AssociatedAlgorithmIdentifier`
    = note: required for `ecdsa::signing::SigningKey<NistP256>` to implement `SignatureAlgorithmIdentifier`
[...]

PR #1034, opened 2023-05-01
x509-cert: adds a `CertReq` builder

baloo:
Fixes https://github.com/RustCrypto/formats/issues/1031

Generates CSR like:
```
---- certificate_request stdout ----
Certificate Request:
    Data:
        Version: 1 (0x0)
        Subject: CN = service.domination.world
        Subject Public Key Info:
            Public Key Algorithm: id-ecPublicKey
                Public-Key: (256 bit)
                pub:
                    04:1c:ac:ff:b5:5f:2f:2c:ef:d8:9d:89:eb:37:4b:
                    26:81:15:24:52:80:2d:ee:a0:99:16:06:81:37:d8:
                    39:cf:7f:c4:81:a4:44:92:30:4d:7e:f6:6a:c1:17:
                    be:fe:83:a8:d0:8f:15:5f:2b:52:f9:f6:18:dd:44:
                    70:29:04:8e:0f
                ASN1 OID: prime256v1
                NIST CURVE: P-256
        Attributes:
            Requested Extensions:
                X509v3 Subject Alternative Name:
                    IP Address:192.0.2.0
    Signature Algorithm: ecdsa-with-SHA256
    Signature Value:
        73:29:74:17:23:a1:12:f4:94:82:3e:b8:8b:eb:f4:21:f4:80:
        7e:ba:1a:0e:63:b4:ea:03:5e:6c:87:43:f1:01:6d:19:34:3c:
        1c:f8:e7:0f:d0:1b:c4:5b:20:b5:80:46:dd:73:0a:84:b4:94:
        00:5e:a3:31:81:21:70:3b:57:a5
-----BEGIN CERTIFICATE REQUEST-----
MIH5MIGnAgEAMCMxITAfBgNVBAMMGHNlcnZpY2UuZG9taW5hdGlvbi53b3JsZDBZ
MBMGByqGSM49AgEGCCqGSM49AwEHA0IABBys/7VfLyzv2J2J6zdLJoEVJFKALe6g
mRYGgTfYOc9/xIGkRJIwTX72asEXvv6DqNCPFV8rUvn2GN1EcCkEjg+gIjAgBgkq
hkiG9w0BCQ4xEzARMA8GA1UdEQQIMAaHBMAAAgAwCgYIKoZIzj0EAwIDQQBzKXQX
I6ES9JSCPriL6/Qh9IB+uhoOY7TqA15sh0PxAW0ZNDwc+OcP0BvEWyC1gEbdcwqE
tJQAXqMxgSFwO1el
-----END CERTIFICATE REQUEST-----
```

### DIFF
--- a/x509-cert/tests/builder.rs
+++ b/x509-cert/tests/builder.rs
@@ -8,7 +8,7 @@ use sha2::Sha256;
 use spki::SubjectPublicKeyInfoOwned;
 use std::{str::FromStr, time::Duration};
 use x509_cert::{
-    builder::{CertificateBuilder, Profile, RequestBuilder},
+    builder::{Builder, CertificateBuilder, Profile, RequestBuilder},
     ext::pkix::{name::GeneralName, SubjectAltName},
     name::Name,
     serial_number::SerialNumber,
--- a/x509-cert/tests/builder.rs
+++ b/x509-cert/tests/builder.rs
@@ -8,7 +8,8 @@ use sha2::Sha256;
 use spki::SubjectPublicKeyInfoOwned;
 use std::{str::FromStr, time::Duration};
 use x509_cert::{
-    builder::{CertificateBuilder, Profile},
+    builder::{CertificateBuilder, Profile, RequestBuilder},
+    ext::pkix::{name::GeneralName, SubjectAltName},
     name::Name,
     serial_number::SerialNumber,
     time::Validity,
@@ -233,4 +234,22 @@ const PKCS8_PRIVATE_KEY_DER: &[u8] = include_bytes!("examples/p256-priv.der");
 fn ecdsa_signer() -> ecdsa::SigningKey<NistP256> {
     let secret_key = p256::SecretKey::from_pkcs8_der(PKCS8_PRIVATE_KEY_DER).unwrap();
     ecdsa::SigningKey::from(secret_key)
+}
+
+#[test]
+fn certificate_request() {
+    use std::net::{IpAddr, Ipv4Addr};
+    let subject = Name::from_str("CN=service.domination.world").unwrap();
+
+    let signer = ecdsa_signer();
+    let mut builder = RequestBuilder::new(subject, &signer).expect("Create certificate request");
+    builder
+        .add_extension(&SubjectAltName(vec![GeneralName::from(IpAddr::V4(
+            Ipv4Addr::new(192, 0, 2, 0),
+        ))]))
+        .unwrap();
+
+    let cert_req = builder.build::<ecdsa::Signature<NistP256>>().unwrap();
+    let pem = cert_req.to_pem(LineEnding::LF).expect("generate pem");
+    println!("{}", openssl::check_request(pem.as_bytes()));
 }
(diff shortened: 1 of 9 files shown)


baloo:
cc @lumag

x509-cert/tests/builder.rs
```diff
@@ -234,3 +235,20 @@ fn ecdsa_signer() -> ecdsa::SigningKey<NistP256> {
...
 }
+
+#[test]
+fn certificate_request() {
+    let subject = Name::from_str("CN=service.domination.world").unwrap();
+
+    let signer = ecdsa_signer();
+    let mut builder = RequestBuilder::new(subject, &signer).expect("Create certificate request");
+    builder.add_extension(&SubjectAltName(vec![GeneralName::IpAddress(
+        OctetString::new(vec![192, 0, 2, 0]).unwrap(),
```
baloo, line 246: note to self: get a conversion between https://doc.rust-lang.org/std/net/enum.IpAddr.html and `GeneralName`

lumag:
@baloo maybe I miss something, why do you have to pass extensions to the `to_extension()`?

baloo:
> @baloo maybe I miss something, why do you have to pass extensions to the `to_extension()`?

Because some extensions may require that, like `Basic Constraints` (https://datatracker.ietf.org/doc/html/rfc5280#section-4.2.1.9):
```
Conforming CAs MUST include this extension in all CA certificates
   that contain public keys used to validate digital signatures on
   certificates and MUST mark the extension as critical in such
   certificates.  This extension MAY appear as a critical or non-
   critical extension in CA certificates that contain public keys used
   exclusively for purposes other than validating digital signatures on
   certificates.  Such CA certificates include ones that contain public
   keys used exclusively for validating digital signatures on CRLs and
   ones that contain key management public keys used with certificate
   enrollment protocols.  This extension MAY appear as a critical or
   non-critical extension in end entity certificates.
```

for now, the behavior of the crate is to mark `BasicConstraints` as critical, but I prefer to leave the door open if we need to revisit that.

lumag:
I think we should check/fix that at the `build` time. For now I'd suggest skipping that (and it definitely isn't a part of CSR builder patchset).

baloo:
Not a fan of breaking the API in the future if we ever need to :/

lumag:
Anyway, it should be a separate change (separate PR or patch inside the PR). Having everything spliced into a single patch makes review very hard. I'd suggest landing the RequestBuilder first and then taking a glance on the to_extension in a separate PR.

Regarding the quoted part. I think we should check that at the build time, depending on the  profile/whatsoever, rather than enforcing it at the add_extensions time.

baloo:
@lumag split that in its own commit.

baloo:
:facepalm: and I did push to `origin` not my fork

baloo:
@lumag last commit is what I meant with the `trait Builder`

lumag:
@baloo the request builder an the trait looks good. As for the fist commit, I'd leave that up to you and @tarcieri .

tarcieri (approved):

Issue #1045, opened 2023-05-04
`cms` and `pkcs7` overlap


tarcieri:
The `cms` and `pkcs7` crates seem to have a lot of overlap.

Should we pick one to be the place for CMS-related functionality and retire the other?

cc @carl-wallace

carl-wallace:
Possibly. The cms crate is more complete with regard to CMS structures. I prefer how ContentInfo is handled there as well (unsurprisingly). The cms crate more consistent with how structures are defined in x509, ocsp, cmpv2 and crmf. The pkcs7 crate may predate x509, which is when lots of stuff was added to der. The pkcs7 crate is mostly SignedData focused. Not sure offhand how much the structures other than ContentInfo differ.

tarcieri:
I added a deprecation notice to `pkcs7` in #1062.

Let's use `cms` as the basis for all work on Cryptographic Message Syntax.

Issue #812, opened 2022-12-20
pkcs7: support for RFC5652 5.2.1


woodruffw:
I'm trying to parse one of Microsoft's custom PKCS#7 containers, and I'm running up against the following error:

```
thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: Der(Error { kind: TagUnexpected { expected: Some(Tag(0x04: OCTET STRING)), actual: Tag(0x30: SEQUENCE) }, position: None })', examples/demo.rs:12:29
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

This is happening because CMS as defined in RFC 5652 defines `eContent` as follows:

```asn1
      eContent [0] EXPLICIT OCTET STRING OPTIONAL
```

whereas PKCS7 defines `content` (the predecessor to `eContent`) as follows:

```asn1
      content [0] EXPLICIT ANY DEFINED BY contentType OPTIONAL
```

Microsoft naturally makes use of this, and chooses to make their `content` a `SEQUENCE` instead of an `OCTET STRING`:

```
SEQUENCE (2 elem)
        OBJECT IDENTIFIER 1.3.6.1.4.1.311.10.1 certTrustList (Microsoft contentType)
        [0] (1 elem)
          SEQUENCE (5 elem)
```

I'll work on a fix for this; just filing an issue for visibility :slightly_smiling_face:

woodruffw:
Whoops, forgot the link: here's where the RFC allows both forms: <https://datatracker.ietf.org/doc/html/rfc5652#section-5.2.1>

smndtrl:
Do you have a specification of the content type format? I couldn't find anything with a quick google. 

If so it should be easy to create a struct for it based on the `der` crate. Like in my [gist](https://gist.github.com/smndtrl/c2e06faf3c70e7e01fd563be9cfb1553)

I don't think vendor specific content types should be supported with the `pkcs7` crate. However there should be a standard way of exposing opaque content

woodruffw:
Oh, just to clarify -- I don't expect `pkcs7` to support the `certTrustList` format at all! That would be unreasonable 🙂 

Instead, I expect it to support both PKCS#7v1.5 *and* IETF CMS-style `SignedData` contents -- that means that it should return a `&[u8]` or similar for PKCS#7v1.5, and an `OctetString` for IETF CMS. Right now it only supports the latter.

woodruffw:
[...]

PR #1051, opened 2023-05-05
Build SignedData CMS (and PKCS #7)

bkstein:
This branch adds a builder for SignedData PKCS #7 messages.


pkcs7/Cargo.toml
```diff
@@ -15,15 +15,29 @@ edition = "2021"
...
 
+# optional dependencies
+sha1 = { version = "0.10"}
+sha2 = { version = "0.10"}
+sha3 = { version = "0.10"}
+signature = { version = "2.1.0", features = ["digest", "alloc"]}
+
 [dev-dependencies]
 der = { version = "0.7", features = ["oid", "pem"] }
+rsa = { version = "0.9.0-pre.1", features = ["sha2"] }
```
tarcieri, line 31: 0.9.0 final is out.

tarcieri:
cc @carl-wallace 

See also #1045 where we are discussing the future of the `pkcs7` crate

pkcs7/Cargo.toml
baloo: (and I would recommend the 0.9.1, it has a fix for signature padding)

pkcs7/src/builder.rs
```diff
@@ -0,0 +1,541 @@
...
+        self.is_signed = true;
+
+        Ok(())
+    }
+
+    /// Return finalized `SignedData` struct.
+    /// This method returns a `ContentInfo` of type `signedData`. After this call, the builder cannot
+    /// be used any more. However, as the signature(s) are stored in the builder, the builder object
+    /// must be kept until the message is dropped.
+    pub fn build(&'s mut self) -> Result<ContentInfo<'s>> {
```
baloo, line 313: Just wondering, would it make sense to reuse the [`Builder` trait](https://docs.rs/x509-cert/0.2.2-pre/x509_cert/builder/trait.Builder.html) from `x509-cert`, it's a dependency already.

bkstein: This would also hold for [`CertificateBuilder`](https://github.com/RustCrypto/formats/blob/bee8a6ea34f978ca474e4e42fceb158f9a39d6dd/x509-cert/src/builder.rs#L258), right?
As for the PKCS #7 builder: how would we handle multiple signers? My current solution is to add a `SignerInfo` including the signature with each call of `SignedDataBuilder::sign()`. Idea: I could possibly rename `sign()` to `add_signer()` and just add the `SignerInfo` without the signature. The signatures would then be calculated when building the message with `SignedDataBuilder::build()`.

pkcs7/Cargo.toml
bkstein: Updated.

bkstein:
> cc @carl-wallace
> 
> See also #1045 where we are discussing the future of the `pkcs7` crate

Yes, I will definitely check the cms crate. But first I'd like to solve my lifetime problem.

tarcieri:
@bkstein the `ContentInfo` type in the `cms` crate is owned and doesn't have a lifetime, which may make it easier to work with:

https://docs.rs/cms/0.2.1/cms/content_info/struct.ContentInfo.html

pkcs7/src/builder.rs
baloo: Yeah, I guess you would need an owned structure to be able to use it.

bkstein:
> @bkstein the `ContentInfo` type in the `cms` crate is owned and doesn't have a lifetime, which may make it easier to work with:
> 
> https://docs.rs/cms/0.2.1/cms/content_info/struct.ContentInfo.html

I had a look into the `cms` crate and I wonder, why the `struct`s own their content while the `pkcs7` `structs` borrow everything. Is there a special reason for this?
However, I think that building `cms`s types will be easier. I will try this tomorrow.

tarcieri:
@bkstein some past discussion here: #765 

We ended up moving many of the crates to use all owned types. This makes things like builders easier (as you are experiencing), but also makes one-pass deserialization from PEM possible.

The `pkcs7` crate predates that decision. We should maybe consider retiring it and moving everything to `cms`.

bkstein:
That sounds reasonable to me, too. The only difference between CMS and PKCS #7 I could find ist the coding of encapsulated content (RFC 5652, 5.2.1) and that should be manageable.

tarcieri:
I've officially deprecated the `pkcs7` crate: #1062 

@bkstein do you think you could update this PR to add this same functionality to `cms` instead?

bkstein:
Yes, I worked exactly on this during the last week. But I'm not ready, yet.

pkcs7/src/builder.rs
bkstein: @baloo I'm currently moving my implementations to the `cms` crate, which uses owned types. It looks like I will be able to reuse your `Builder` trait for the `SignerInfo` objects. 🙂

bkstein:
I moved the code from crate pkcs7 (which is deprecated) to cms. I used `Builder` from x509-cert to create the `SignerInfo`. `Builder` should probably be moved to a more central location in the future. I also moved `der::asn1::set_of::der_sort()` to make it public for some other types. The implementation is not yet usable for creating PKCS #7 messages, as enveloped_data must also be created. This will be next.

x509-cert/src/name.rs
```diff
@@ -166,6 +170,25 @@ impl FromStr for RelativeDistinguishedName {
...
+        for i in 0..slice.len() {
+            let mut j = i;
+
+            while j > 0 && slice[j - 1].der_cmp(&slice[j])? == Ordering::Greater {
+                slice.swap(j - 1, j);
+                j -= 1;
+            }
+        }
+        Ok(RelativeDistinguishedName(SetOfVec::try_from(vec)?))
+    }
```
tarcieri, line 189:
```suggestion
    fn try_from(mut vec: Vec<AttributeTypeAndValue>) -> der::Result<RelativeDistinguishedName> {
        Ok(RelativeDistinguishedName(SetOfVec::try_from(vec)?))
    }
```

pkcs7/src/signer_info.rs
```diff
@@ -15,12 +15,12 @@ use x509_cert::{
 /// ```text
 /// SignedAttributes ::= SET SIZE (1..MAX) OF Attribute
 /// ```
-type SignedAttributes<'a> = SetOfVec<Attribute>;
+pub type SignedAttributes<'a> = SetOfVec<Attribute>;
```
baloo, line 18: I believe changes in pkcs7 need to be dropped (deprecated).

baloo:
looks great! Thank you for doing that!

x509-cert/src/name.rs
bkstein: I agree, that's much better. I will move `der_sort()` back.

pkcs7/src/signer_info.rs
bkstein: Right, I will check pkcs7 for remainders of my changes.

bkstein:
Question to the audience: `SignerInfoBuilder` uses generics for the signing algorithms. But the user also needs to pass values for digest and signature algorithms in `SignerinfoBuilder::new()`. It's up to the user to set theses values consistent with the used signer. Do you see a chance to derive this info so users don't have to pass them?
In test `build_signed_data()` I create `signer_info_builder_2` and I had to guess, how to set `signature_algorithm_2`.

cms/Cargo.toml
```diff
@@ -16,17 +16,27 @@ rust-version = "1.65"
 [dependencies]
 der = { version = "0.7", features = ["alloc", "derive", "oid", "pem"] }
 spki = { version = "0.7" }
-x509-cert = { version = "0.2", default-features = false, features = ["pem"] }
+x509-cert = { version = "0.2.2-pre", default-features = false, features = ["pem"], path = "../x509-cert"}
 const-oid = { version = "0.9", features = ["db"] } # TODO: path = "../const-oid"
 
+# optional dependencies
```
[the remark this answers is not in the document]
tarcieri: They appear to be optional now

tarcieri:
@bkstein it looks like you're already bounding on `DynSignatureAlgorithmIdentifier` which can be used to get the algorithm identifier for the signature algorithm. Take a look at `x509-cert` for examples.

bkstein:
Not sure, if this is the right place to discuss, but it affects the PR:
`EncapsulatedContentInfo::econtent` is currently implemented as `Option<Any>`, while RFC 5652 wants an OctetString:
```rust
/// The `EncapsulatedContentInfo` type is defined in [RFC 5652 Section 5.2].
///
/// ```text
///   EncapsulatedContentInfo ::= SEQUENCE {
///       eContentType       CONTENT-TYPE.&id({ContentSet}),
///       eContent           [0] EXPLICIT OCTET STRING
///               ( CONTAINING CONTENT-TYPE.
///                   &Type({ContentSet}{@eContentType})) OPTIONAL }
/// ```
///
/// [RFC 5652 Section 5.2]: https://www.rfc-editor.org/rfc/rfc5652#section-5.2
#[allow(missing_docs)]
pub struct EncapsulatedContentInfo {
    pub econtent_type: ObjectIdentifier,
    #[asn1(context_specific = "0", tag_mode = "EXPLICIT", optional = "true")]
    pub econtent: Option<Any>,
}

```
Is this for compatibility with the PKCS #7 specification?

tarcieri:
@bkstein see #812 re `eContent`

bkstein:
@tarcieri This looks like the customer has to take care of what he or she adds as `eContent`? That's OK at least for now. I need to create PKCS#7 messages and I will try to write another builder for enveloped data. Perhaps I can add some comfort there.

tarcieri:
@bkstein yeah, we can add a convenience method for `OctetString` per  RFC5652 but it should at least be possible to provide an `Any` to satisfy PKCS#7 use cases

baloo:
The PR looks great, but I think we need to reintroduce min-versions first (#1071) and split that in individual PR / minor releases on all the various crates this affects (just so the next person cutting a release of `cms` doesn't make a mistake).
I'd be happy to follow up and cut the various releases in the individual PRs.

pkcs7/src/lib.rs
```diff
@@ -15,6 +15,9 @@
     unused_qualifications
 )]
 
+#[cfg(feature = "std")]
+extern crate std;
```
tarcieri, line 19: This seems irrelevant now?

tarcieri:
@baloo AFAICT the only relevant changes are to `x509-cert` and `cms`?

@bkstein FYI, the `const-oid` in-tree is targeting the next breaking release (0.10.0-pre) and not the one used by e.g. `der`/`x509-cert` (v0.9), so changes added here won't immediately be available

pkcs7/src/lib.rs
bkstein: Yes. I thought I had reverted all changes in pkcs7. Don't know why this was still in.

bkstein:
@baloo 

> I'd be happy to follow up and cut the various releases in the individual PRs.

You're welcome, please go ahead. I'm not familiar with the versioning and release process anyway. I also didn't touch the version numbers. Will this also cover the `const-oid` issue @tarcieri mentioned above?

tarcieri:
No, bumping the `const-oid` version is a breaking change

baloo:
> @baloo AFAICT the only relevant changes are to `x509-cert` and `cms`?

it uses the changes on `const-oid` in the tests, that just means we have to send those in a separate PR and bump the version to `0.10` when a release is cut.

baloo:
@bkstein could you split [the changes on `x509-cert`](https://github.com/RustCrypto/formats/pull/1051/files#diff-4469cf07082e1e146c849e0397c7eace4bb6cb5efe55f1b0a95aae921ddd97ac) and send that in a separate PR?
I'll merge that and cut a patch release for you. (I just want you to keep authorship on the change)

tarcieri:
@baloo as I mentioned earlier, we can't bump `const-oid` because it's a v0.10 prerelease which would be a breaking change to everything to upgrade.

However, I can backport the changes to v0.9 if need be.

baloo:
I'm confused why can't we wait to merge this until 0.10 is released?

tarcieri:
That's probably going to be awhile. It's a breaking change to everything in-tree, and also a breaking change to `digest` and all of the hash function implementations.

It would be nice to get to 1.0 soon, but for now it's one of the most painful upgrades in the project.

baloo:
alright then :) 0.9 backport it is! :D

bkstein:
> @bkstein could you split [the changes on `x509-cert`](https://github.com/RustCrypto/formats/pull/1051/files#diff-4469cf07082e1e146c849e0397c7eace4bb6cb5efe55f1b0a95aae921ddd97ac) and send that in a separate PR? I'll merge that and cut a patch release for you. (I just want you to keep authorship on the change)

Hi @baloo, I will do the split tomorrow.

tarcieri:
@bkstein can you split out the `const-oid` changes? That will make them a lot easier to cherry-pick in a backport

bkstein:
Sure, I already expected this 🙂.

cms/Cargo.toml
```diff
@@ -16,17 +16,27 @@ rust-version = "1.65"
...
+sha1 = { version = "0.10", optional = true}
+sha2 = { version = "0.10", optional = true}
+sha3 = { version = "0.10", optional = true}
+signature = { version = "2.1.0", features = ["digest", "alloc"], optional = true}
+
 [dev-dependencies]
 hex-literal = "0.4"
+pem-rfc7468 = "0.7.0"
 pkcs5 = { version = "0.7" }
+rsa = { version = "0.9.1", features = ["sha2"] }
```
baloo, line 32:
```suggestion
rsa = { version = "0.9.2", features = ["sha2"] }
```

cms/Cargo.toml
```diff
@@ -16,17 +16,27 @@ rust-version = "1.65"
 [dependencies]
 der = { version = "0.7.6", features = ["alloc", "derive", "oid", "pem"] }
 spki = { version = "0.7" }
-x509-cert = { version = "0.2", default-features = false, features = ["pem"] }
+x509-cert = { version = "0.2.2-pre", default-features = false, features = ["pem"], path = "../x509-cert"}
```
baloo, line 19:
```suggestion
x509-cert = { version = "0.2.3", default-features = false, features = ["pem"], path = "../x509-cert"}
```

Fresh off the press!

bkstein:
Just realized, that resolving does not automatially accept the suggested changes. 🙄

baloo:
we'll still need you to send the `const-oid` changes in a new PR. Once merged we'll backport that to `const-oid 0.9` release you can use here.

bkstein:
> we'll still need you to send the `const-oid` changes in a new PR. Once merged we'll backport that to `const-oid 0.9` release you can use here.
> 
> Sorry for the extra work.

I checked the `const-oid` changes and found out, that all OIDs are now available with RFC 5911. Thus, we don't need a new `const-oid`. I will check OID usage in my branch, change the `const_oid::db::rfc5652::...` to the corresponding `const_oid::db::rfc5911::...` and remove rfc5652.txt.

bkstein:
@tarcieri @baloo I reverted the changes in `const-oid`, so we don't need a new version here.

const-oid/oiddbgen/Cargo.lock
```diff
@@ -36,18 +36,18 @@ dependencies = [
 
 [[package]]
 name = "proc-macro2"
-version = "1.0.36"
```
baloo, line 39: Could you revert changes on [const-oid/oiddbgen/Cargo.lock](https://github.com/RustCrypto/formats/pull/1051/files#diff-33d26b33538c98cd911ffec8d3685569bc7c07c8fa9cb58171b58c08f89ed50c) as well?

baloo:
(lastly, I think we'll want you to squash those commits if you could)

baloo (approved):

carl-wallace:
Sorry I missed this request somehow last month (was very busy then and just missed it). I reviewed now in prep of reviewing the EnvelopedData changes and had a couple of comments.  None of this is essential (though fixing the test case probably ought be done).

SignerInfoBuilder.assemble
- Ought the signer be interrogated for the SignerIdentifier (as it is for signature algorithm)? Same could apply to yielding a certificate (or certification path) that is consistent with the signer as well. This was not appropriate for a certificate builder but given this is a little higher level it may be worth another trait.
- Should there be a consistency check here re: digest algorithm and signature algorithm?

SignedDataBuilder.build
- Same question re: interrogating SignerInfos for digest algorithms and for consistency check where external is used.

Test comments
- The test_build_signed_data uses SHA256 and SHA512 in two different SignerInfo structures but only lists SHA256 in the SignedData.digestAlgorithms field. Section 5.3 indicates hash algorithms in SignerInfo.digestAlgorithm SHOULD be in the SignedData.digestAlgorithms field. 
- SHA512 is inconsistent with ecdsaWithSHA256 as well (which may suggest the consistency check mentioned above is needed). 
- May be worth including a certs-only test (which worked fine).

Questions
- Why is the signing time attribute required? This seems like something that ought be optional. RFC6019 defines an alternative signing time, for example. Some protocols (like SCEP) use a nonce. A brief survey of things that reference RFC5652 doesn't show much in the way of requiring signingTime (RFC8551 and RFC8995 intimate it is required but there's no MUST).

Nits
- The word signed is presented in quotes in documentation comments, but unsigned is not. Suggest dropping the quotes.

bkstein:
@carl-wallace Thanks for reviewing this.

I have some questions/remarks.

SignedDataBuilder.build/assemble

Could you give an example, how you would interrogate the info? 

Test comments

- If I understand CMS correctly, digest algorithm and signature algorithm are independent. The digest algorithm is used to hash the message content. The hash value is then signed together with other signed attributes. Please correct me, if I'm wrong.
- A "Degenerate certificates-only CMS SignedData"  test is a very good idea. 🙂

Questions

- As for the signing time attribute: I think you are right and it is not required. We should probably leave it out. Also to remove unnecessary dependencies. The builder is currently not `no_std` and I have to admit, that I didn't have that on the agenda. Understanding RFC 5652 was hard enough.

Nits

- The quotes definitely MUST be dropped.

carl-wallace:
Re: interrogation, adding a trait bound to S is what I was thinking. At present it is `Keypair + DynSignatureAlgorithmIdentifier`. Adding some trait that returns a SignerIndentifier here (or RecipientIdentifier for EnvelopedData) might close a gap. Where S is a P12 or a hardware module, this info ought be available. I've no surveyed where this would go but it seems doable.

Re: hash algorithm and signature algorithm, you're right they can move independently, but accumulating the full list in the SignedData later is required (as a SHOULD). The consistency check suggestion is likely too much, just ignore that.

Here's a certs only test case:

```
#[test]
fn test_build_certs_only_signed_data() {
    // Make some content
    let content = EncapsulatedContentInfo {
        econtent_type: const_oid::db::rfc5911::ID_DATA,
        econtent: None,
    };

    let certificate_buf = include_bytes!("examples/ValidCertificatePathTest1EE.pem");
    let certificate = x509_cert::Certificate::from_pem(certificate_buf).unwrap();

    let mut builder = SignedDataBuilder::new(&content);

    let signed_data_pkcs7 = builder
        .add_certificate(CertificateChoices::Certificate(certificate))
        .expect("error adding certificate")
        .build()
        .expect("building signed data failed");
    let signed_data_pkcs7_der = signed_data_pkcs7
        .to_der()
        .expect("conversion of signed data to DER failed.");
    println!(
        "{}",
        pem_rfc7468::encode_string("PKCS7", LineEnding::LF, &signed_data_pkcs7_der)
            .expect("PEM encoding of signed data DER failed")
    );
}
```

PR #1052, opened 2023-05-05
Add notes about `UTCTime` range being 1970-2049

nick-mobilecoin:
It looks like `UtcTime` currently has a lower bound of 1970.
I'm very open to wording tweaks IFF a documentation change ends up being the way to address this.

RFC 5280 says that `UTCTime` is 1950-2049, however due the common use of
computer times being based off of `UNIX_EPOCH` the implementation of
[`UtcTime`](https://docs.rs/der/latest/der/asn1/struct.UtcTime.html) is limited to 1970.

This limitation is due to the `MIN_YEAR` bounding valid
[`DateTime`](https://docs.rs/der/latest/der/struct.DateTime.html)
structs.
The `DateTime` struct is the underlying implementation of `UTCTime`.

Thanks!

### DIFF
--- a/der/src/asn1/utc_time.rs
+++ b/der/src/asn1/utc_time.rs
@@ -25,6 +25,9 @@ use std::time::SystemTime;
 /// >   interpreted as `19YY`; and
 /// > - Where `YY` is less than 50, the year SHALL be interpreted as `20YY`.
 ///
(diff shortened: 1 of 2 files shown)


tarcieri (approved):
Looks good, thanks

PR #1054, opened 2023-05-07
Add `gss-api` crate for GSS-API related structs

smndtrl:
I came across the need to parse GSS API-related structures in my current work. As these are based on ASN1 DER, I built it using the `der` crate. 
I'm not sure this is a crate you would want to have as part of this repo. Let me know.

Side note: [[MS-SPNG] ](https://learn.microsoft.com/en-us/openspecs/windows_protocols/ms-spng/8e71cf53-e867-4b79-b5b5-38c92be3d472) specifies the hint_name as `GeneralString`, I exposed it as `AnyRef` as there's no implementation for `GeneralString` and the value should be static according to the spec.


tarcieri:
This is a welcome contribution, thanks!

gssapi/Cargo.toml
```diff
@@ -0,0 +1,32 @@
+[package]
+name = "gssapi"
```
tarcieri, line 2: The `gssapi` crate name is taken but I was able to get `gss-api`

Cargo.toml
```diff
@@ -10,6 +10,7 @@ members = [
     "crmf",
     "der",
     "der/derive",
+    "gssapi",
```
tarcieri, line 13:
```suggestion
    "gss-api",
```

gssapi/Cargo.toml
```diff
@@ -0,0 +1,32 @@
...
+name = "gss-api"
+version = "0.0.0"
+description = """
+Pure Rust implementation for Generic Security Service Application Program Interface
+"""
+authors = ["RustCrypto Developers"]
+license = "Apache-2.0 OR MIT"
+repository = "https://github.com/RustCrypto/formats/tree/master/gssapi"
+categories = ["cryptography", "data-structures", "encoding", "no-std", "parser-implementations"]
+keywords = ["crypto", "key", "gssapi", "rfc1509", "rfc2478", "rfc4178", "ms-spng"]
```
tarcieri, line 11: crates.io allows a maximum of 5 keywords

smndtrl: I'm fine with `gss` and `gssapi` in there. Rest is optional/not necessary in my opinion. Any keywords you would still like to have? Maybe leave `crypto` in?

tarcieri: `kerberos` would probably make sense.

Re: categories, perhaps `authentication` would make sense. You could possibly get rid of `data-structures`, as I believe that's more for `std::collections`-style data structures than protocol messages.

tarcieri:
Also if you're basically done with an initial crate, can you remove draft?

Issue #1053, opened 2023-05-06
der: Support custom error types.


turbocool3r:
I'm currently developing a library for parsing the [Image4](https://www.theiphonewiki.com/wiki/IMG4_File_Format) format found in boot images on Apple devices which is based on DER. This format has some constraints not related to the DER and I'd like to provide a custom error type that implements `From<der::Error>` and has some additional variants I need. Is this something that would be possible to add?

I think the best way to do it is to add associated `Error` types in the `Decode` and `Encode` traits and wherever needed. I'll be happy to provide a PR.

[...]

PR #1055, opened 2023-05-08
Add custom error types to the Decode and DecodeValue traits.

turbocool3r:
Adds support for specifying custom error types for the `Encode`, `Decode`, `EncodeValue`, `DecodeValue`, `DerOrd` and `ValueOrd` traits (see #1053).

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -64,3 +64,7 @@ x509-ocsp         = { path = "./x509-ocsp" }
 p256 = { git = "https://github.com/RustCrypto/elliptic-curves.git" }
 # Pending a release of 0.11.0-pre.2
 whirlpool = { git = "https://github.com/RustCrypto/hashes.git" }
+
+# https://github.com/RustCrypto/formats/pull/1055
+# https://github.com/RustCrypto/signatures/pull/809
+ecdsa = { git = "https://github.com/turbocool3r/signatures", branch = "custom-errors" }
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -562,8 +562,7 @@ dependencies = [
 [[package]]
 name = "ecdsa"
 version = "0.17.0-pre.5"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "d7e045ee5c360512162782f3d4cb07d2f4ce8c4ef9bf7c77ec16d1cf60b3d5ca"
+source = "git+https://github.com/turbocool3r/signatures?branch=custom-errors#d9d394ad7a87de8943a0eb39b85f316149ebb025"
 dependencies = [
  "der",
  "digest 0.11.0-pre.8",
@@ -1272,8 +1271,7 @@ checksum = "c707298afce11da2efef2f600116fa93ffa7a032b5d7b628aa17711ec81383ca"
 [[package]]
 name = "rfc6979"
 version = "0.5.0-pre.3"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "045972f2f66b9467a2f6834b7fd0f9b23ca214b4a8700b880c36edb726e96da6"
+source = "git+https://github.com/turbocool3r/signatures?branch=custom-errors#d9d394ad7a87de8943a0eb39b85f316149ebb025"
 dependencies = [
  "hmac 0.13.0-pre.3",
  "subtle",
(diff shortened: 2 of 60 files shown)


tarcieri:
Just to set expectations: it will probably be a few months before we're ready to start making breaking changes again

der/src/ord.rs
```diff
@@ -11,27 +11,36 @@ use core::{cmp::Ordering, marker::PhantomData};
 /// This is used by the DER encoding for `SET OF` in order to establish an
 /// ordering for the elements of sets.
 pub trait DerOrd {
+    /// An error returned when the comparison fails.
+    type Error: From<Error>;
```
tarcieri, line 15: Per my comments on #1053, I think things would be simpler without this associated type, keeping the error type as `der::Error`, and leveraging the `From<der::Error>` bound in e.g. `Decode`/`Encode` to handle converting error types.

Same goes for `ValueOrd`.

turbocool3r:
> Just to set expectations: it will probably be a few months before we're ready to start making breaking changes again

That's ok, I'll leave this PR in place until it can be accepted.

der/src/ord.rs
turbocool3r: Fixed this.

tarcieri:
It's possible to make breaking changes now.

That said, I've opened #1328 which provides a less invasive alternative: it allows `der::Error` to carry with it an arbitrary error source when the `std` feature is enabled.

That said, I think this approach is still interesting, and closer to e.g. what `serde` does.

tarcieri:
If you can get this PR green we can consider it over #1328.

turbocool3r:
Ok, I'll try to do this by Wednesday

turbocool3r:
It seems this won't be completely green because there are some dependencies on external crates like `signature` which should be modified separately.

There also was an issue with the `DecodePem` trait: adding a generic error type there causes the `from_pem` method to fail the borrow check even if both `T` and `Error` are marked as `'static`. Forcing the error to be `der::Error` there fixes the issue.

tarcieri:
> It seems this won't be completely green because there are some dependencies on external crates like signature which should be modified separately.

Err, how so?

turbocool3r:
Like here https://github.com/RustCrypto/formats/actions/runs/7631920556/job/20790968317?pr=1055#step:4:297.

tarcieri:
Aah, in `ecdsa`. I can open a PR for that if you'd like.

turbocool3r:
> Aah, in `ecdsa`. I can open a PR for that if you'd like.

It would be nice. I think we need to first get this as it goes to see the complete list of crates that need to be modified.

turbocool3r:
Yes, it seems it's only ecdsa, not signature.

turbocool3r:
Hey, how about that pull request? I could probably do that myself to save some time.

tarcieri:
Sorry, I've been meaning to get to it and I'd love to land this, but if you'd like to go for it that works too.

It's going to be oddly circular so you need to make a PR to https://github.com/rustcrypto/formats which uses a `[patch.crates-io]` directive to point to this branch, then you'll need to reference the branch in this PR to complete the cycle.

turbocool3r:
Did you mean the https://github.com/RustCrypto/signatures repo? By this branch do you mean my fork or some other branch?

tarcieri:
Err sorry, yes https://github.com/RustCrypto/signatures

You'll need to point to your branch on your fork for both this repo when updating signatures and again in a circular manner in your fork of `signatures` to update this PR

turbocool3r:
Seems all green now, the `signatures` PR is https://github.com/RustCrypto/signatures/pull/809.

tarcieri:
Thanks, will review and try to get this merged ASAP

turbocool3r:
Thanks!

Cargo.toml
```diff
@@ -64,3 +64,6 @@ x509-ocsp         = { path = "./x509-ocsp" }
 p256 = { git = "https://github.com/RustCrypto/elliptic-curves.git" }
 # Pending a release of 0.11.0-pre.2
 whirlpool = { git = "https://github.com/RustCrypto/hashes.git" }
+
+# PR #1055
```
baloo, line 68:
```suggestion
# https://github.com/RustCrypto/formats/pull/1055
# https://github.com/RustCrypto/signatures/pull/809
```

turbocool3r: Fixed this.

tarcieri (approved):

PR #1065, opened 2023-05-15
der: add `SetOfVec::{extend, from_iter}`

tarcieri:
Adds inherent methods to `SetOfVec` shaped like the `Extend` and `FromIterator` traits.

However, the methods are fallible, so they don't quite fit the existing traits, hence why inherent methods are used instead.

### DIFF
--- a/der/src/asn1/set_of.rs
+++ b/der/src/asn1/set_of.rs
@@ -214,6 +214,19 @@ where
         }
     }
 
+    /// Create a new [`SetOfVec`] from the given iterator.
+    ///
+    /// Note: this is an inherent method instead of an impl of the
+    /// [`FromIterator`] trait in order to be fallible.
+    pub fn from_iter<I>(iter: I) -> Result<Self>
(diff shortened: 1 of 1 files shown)


der/src/asn1/set_of.rs
```diff
@@ -214,6 +214,19 @@ where
...
 
+    /// Create a new [`SetOfVec`] from the given iterator.
+    ///
+    /// Note: this is an inherent method instead of an impl of the
+    /// [`FromIterator`] trait in order to be fallible.
+    pub fn from_iter<I>(iter: I) -> Result<Self>
+    where
+        I: IntoIterator<Item = T>,
+    {
+        let mut ret = Self::new();
```
tarcieri, line 225: This could use `Vec::with_capacity` if it were constrained to an `ExactSizeIterator`

tarcieri: Went ahead and did that in 8b55b41

PR #1071, opened 2023-05-16
CI: re-enable all minimal-versions checks

tarcieri:
All of the blockers on these checks should now be resolved


baloo:
To fix the min-versions on `cmpv2` we can either cut a new pre-release of `crmf` (`0.2.0-pre.1`) with:
```
x509-cert = { version = "0.2", default-features = false }
```
it complains because of a conflict of dependencies:
https://crates.io/crates/crmf/0.2.0-pre.0/dependencies

tarcieri:
Maybe keep it disabled with a TODO to re-enable it after the next release

PR #1082, opened 2023-05-25
Added cms and a drepcation notice for pkcs7

raphaelahrens:
Since `pkcs7` was deprecated in favor of `cms`  a note was added to the overview. cms was added, with description and links to the RFCs.


tarcieri:
Thanks!

PR #1094, opened 2023-06-02
const_oid: add RFC 6962 OIDs

imor:
This PR adds OIDs from [RFC 6962](https://www.rfc-editor.org/rfc/rfc6962).


tarcieri:
Thanks!

PR #1095, opened 2023-06-04
initial rfc3161

enri1196:
With this PR I'd like to introduce RFC3161 TimeStamp Protocol to the available formats.


tarcieri:
re: names, how about `x509-ocsp`?

tarcieri:
re: crate names, how about `x509-tsp`?

enri1196:
Yeah, it’s good

carl-wallace:
I reviewed the code and exercised it a bit. I can push edits for the items below along with two test cases that use OpenSSL-generated artifacts if you like.

- To be consistent with other crates in the project, some names should be camel case, i.e., TsaPolicyId, PkiStatusInfo, PkiFreeText, PkiStatus, PkiFailureInfo, TstInfo (instead of TSAPolicyId, PKIStatusInfo, etc)
- Some structures are defined in the cmpv2 crate and could be reused here: PkiStatusInfo, PkiFreeText, PkiStatus, PkiFailureInfo
- Two fields should use large integer support not u64 or i32: nonce and serial_number
- Accuracy fields could be smaller than u64 since range is 1 to 999 for millis and micros
- The cert_req and ordering fields should feature a DEFAULT value
- IMPLICIT tag mode should be added in a few spots (and EXPLICIT maybe added to the tsa field for clarity)

Separately, while reviewing this I noticed there is a typo in PkiFailureInfo in CMPv2 that causes a number of values to be off by one (15 is repeated for UnacceptedExtension and values above that are off). I will send a PR to fix this separately.

carl-wallace:
Yes. I wanted to affirm the changes I thought were needed for default and tags worked. The changes were small. Just pushed changes to your repo. Thanks for preparing this.

carl-wallace (approved):
Changes that were identified were made.

tarcieri:
@enri1196 can you:

1) rename the crate to `x509-tsp`
2) rename the corresponding directory to `x509-tsp`

enri1196:
sure, renamed to `x509-tsp`

x509-tsp/Cargo.toml
```diff
@@ -0,0 +1,24 @@
+[package]
+name = "x509-tsp"
+version = "0.1.0"
+edition = "2021"
+description = """
+Pure Rust implementation of the Timestamping Protocol as described in RFC 3161.
+"""
+authors = ["RustCrypto Developers"]
+license = "Apache-2.0 OR MIT"
+repository = "https://github.com/RustCrypto/formats/tree/master/cms"
```
imor, line 10: Looks like this should be renamed to `https://github.com/RustCrypto/formats/tree/master/x509-tsp`

x509-tsp/src/lib.rs
```diff
@@ -0,0 +1,227 @@
...
+///     accuracy                     Accuracy                 OPTIONAL,
+///     ordering                     BOOLEAN             DEFAULT FALSE,
+///     nonce                        INTEGER                  OPTIONAL,
+///       -- MUST be present if the similar field was present
+///       -- in TimeStampReq.  In that case it MUST have the same value.
+///     tsa                          [0] GeneralName          OPTIONAL,
+///     extensions                   [1] IMPLICIT Extensions   OPTIONAL  }
+/// ```
+#[derive(Clone, Debug, Eq, PartialEq, Sequence)]
+pub struct TSTInfo {
```
imor, line 99: Shouldn't this be named `TstInfo`?

enri1196: You're right! Fixed both, thanks!

tarcieri (approved):

PR #1102, opened 2023-06-10
Add Hash implementation to EncodedPoint

nresare:
It would be nice to be able to have public ssh keys implement Hash so that they can be used as HashMap keys or be put in HashSets. 

This minimal change will in turn make it possible to do the analogous change in [ssh-key](https://github.com/RustCrypto/SSH/tree/master/ssh-key)


tarcieri:
Released in `sec1` v0.7.3 (#1155)

PR #1103, opened 2023-06-12
tls_codec: use fully qualified name in derive macro

imor:
Using `#[derive(TlsSerialize)]` on a struct forced the user to include `std::format`.

### DIFF
--- a/tls_codec/derive/src/lib.rs
+++ b/tls_codec/derive/src/lib.rs
@@ -728,7 +728,7 @@ fn impl_serialize(parsed_ast: TlsStruct) -> TokenStream2 {
                             let expected_written = tls_codec::Size::tls_serialized_len(&self);
                             debug_assert_eq!(written, expected_written, "Expected to serialize {} bytes but only {} were generated.", expected_written, written);
                             if written != expected_written {
-                                Err(tls_codec::Error::EncodingError(format!("Expected to serialize {} bytes but only {} were generated.", expected_written, written)))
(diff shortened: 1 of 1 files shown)


franziskuskiefer (commented):
Do you have a minimal example where this makes trouble? Generally this shouldn't be necessary.

imor:
Just checked the output of `cargo expand` and it expands the format to `::alloc::fmt::format`:

```rust
                  Err(
                        tls_codec::Error::EncodingError({
                            let res = ::alloc::fmt::format(
                                format_args!(
                                    "Expected to serialize {0} bytes but only {1} were generated.",
                                    expected_written, written
                                ),
                            );
                            res
                        }),
                    )
```

Not sure why that shouldn't be sufficient to compile the code.

imor:
@tarcieri thanks that makes sense, I'll add a `use alloc::format` in my code.

The question about whether `tls_codec_derive` should refer to items in stdlib still remains. IMHO it should not unconditionally depend upon stdlib because it restricts the derive macros to be usable only in stdlib environments. If you guys agree I can open another issue to make `tls_codec_derive` no_std compatible.

I will also be closing this PR since no_std compatibility is more work than what is done in this PR.

franziskuskiefer:
Yes, the derive crate currently requires `std`.
Making it `no_std` compatible would be nice but indeed a little more work.

imor:
Closing this PR. A new issue for `no_std` support is added [here](https://github.com/RustCrypto/formats/issues/1106)

Issue #1111, opened 2023-06-20
serdect: is it actually constant-time?


fjarri:
The binary serializer uses `serializer.serialize_tuple()` and `serialize_element()` which, in some formats at least, makes it data-dependent. E.g. MessagePack prepends every element greater than 127 with `0xCC`. 

Also, this contradicts the documentation claim:
> When using a binary format, the data is serialized as-is into binary.

What was the reason behind not using `serialize_bytes()`? Seems like it would provide better constant-time guarantees?

tarcieri:
The README describes it as follows:

> This crate provides "best effort" constant-time helper methods for reducing the amount of timing variability involved in serializing/deserializing data when using serde, Rust's standard serialization framework.

Really the goal is to be less noisy than e.g. a JSON serialization of bytes as an integer array. PRs accepted to improve the wording in the README.

> What was the reason behind not using serialize_bytes()? Seems like it would provide better constant-time guarantees?

The slice serializer just uses the `Serialize::serialize` impl on `[u8]`. I'm not familiar with how that differs from `serialize_bytes`, but if you think there's a compelling case I'd suggest opening a PR. We test with a number of different serde format impls so perhaps you can point out specific deficiencies with a particular one.

`serialize_tuple` is used by the array serializer, which avoids length prefixing the data because it's known a priori by the type system. That implementation avoids a problematic deficiency with serde which prevents it from working with const generics: https://github.com/serde-rs/serde/issues/1937

fjarri:
> The slice serializer just uses the Serialize::serialize impl on [u8]. I'm not familiar with how that differs from serialize_bytes

It seems that it is different, since `serialize_bytes` serializes bytes verbatim in MessagePack, while `serialize` doesn't.

> serialize_tuple is used by the array serializer, which avoids length prefixing the data because it's known a priori by the type system. 

That's true, but then binary formats instead prefix separate elements, which kind of defeats the point.

tarcieri:
Can you point to a specific implementation, and ideally, write some test cases which capture the issue?

tarcieri:
cc @daxpedda

fjarri:
> Can you point to a specific implementation, and ideally, write some test cases which capture the issue?

Let me make an MRE.

Also, speaking of `serialize_tuple()`, I would argue that while it would be nice to have an array serialization with no length specifier, it is secondary to the intent of `serdect` which is to a) strive for constant-timeness, and b) serialize bytes unchanged.

[...]

PR #1112, opened 2023-06-21
serdect: safer bytes handling

fjarri:
See #1111 for the discussion leading to this PR.

**Note:** this is a breaking change in the ABI.

Previous behavior: byte slices were serialized with the default `serde` implementation for `[T]` with `T = u8`. For some formats like MessagePack or CBOR it leads to array serialization, and separate `u8` values get treated as packed integers - that is, every value over 127 is prefixed (0xCC in MessagePack, 0x18 in CBOR). This leads to non-constant-time behavior, and also is counter-intuitive, because it results in variable serialized length for the byte slices of the same length.

Changes:
- Use `serialized_bytes` in the `slice` module, funneling to the specific "bytestring" pathways in corresponding formats, which don't do packing. 
- `slice::deserialize_hex_bin_vec()` changed accordingly. Note that `slice::deserialize_hex_or_bin()` deserialized from a bytestring already, so it was inconsistent with the serializer (which could be revealed in CBOR tests, if it was used there).
- Added a table with supported formats to Readme.md.

In tests, I replaced the 0x0F value with 0xFF to check if the packing occurs. CBOR and MessagePack do packing, so their tests include the packing markers in the tests for array serialization, and the lack of constant-timeness is reflected in the table in Readme.


tarcieri:
> In tests, I replaced the 0x0F value with 0xFF to check if the packing occurs. Currently the CBOR array tests fail because of it.

Perhaps you can leave it alone for now and we can examine arrays separately in a followup.

I'd also be curious to know which formats we currently test against this impacts and what the impact is, i.e. is this a mostly backwards compatible change or are there formats other than MessagePack where this is a breaking change.

Note that if this is a breaking change it will take quite awhile (i.e. months) to roll out as `serdect` is a very deep dependency.

fjarri:
It is a breaking change for CBOR because the tags for an array and a bytestring are different (see the change in the test). Bincode seems to be unaffected.

serdect/README.md
```diff
@@ -29,6 +29,18 @@ While this crate can't ensure that format implementations don't perform
 other kinds of data-dependent branching on the contents of the serialized data,
 using a constant-time hex serialization with human-readable formats should
 help reduce the overall timing variability.
+The table below lists the formats `serdect` is tested against, and
+whether the serialization via `array` or `slice` is constant-time for them.
```
daxpedda, line 33: I think we should be careful when using the word "constant-time", as pointed out above, this is best effort, we can't actually guarantee it.
```suggestion
The table below lists the formats `serdect` is tested against, and
whether the serialization via `array` or `slice` is data-independent.
```

serdect/README.md
```diff
@@ -29,6 +29,18 @@ While this crate can't ensure that format implementations don't perform
...
+
+| Format                          | `array`  | `slice`  |
+|---------------------------------|----------|----------|
+| Bincode (`bincode=^1`)          | &#x2705; | &#x2705; |
+| CBOR (`ciborium=^0.2`)          | &#x274c; | &#x2705; |
+| MessagePack (`rmp-serde`)       | &#x274c; | &#x2705; |
+| JSON (`serde-json-core=^0.5`)   | &#x2705; | &#x2705; |
+| JSON (`serde-json=^1`)          | &#x2705; | &#x2705; |
+| TOML (`toml=^0.7`)              | &#x2705; | &#x2705; |
```
daxpedda, line 42: Why is `rmp-serde` missing a verrsion? Also I don't think it makes sense to talk about the individual formats, we should just say which crate, because we don't actually test the formats, but only their implementations.

I also added links to each crate and centered the other columns.
```suggestion
| Format                                                             | `array`  | `slice`  |
|--------------------------------------------------------------------|:--------:|:--------:|
| [`bincode`](https://crates.io/crates/bincode) v1                   | &#x2705; | &#x2705; |
| [`ciborium`](https://crates.io/crates/ciborium) v0.2               | &#x274c; | &#x2705; |
| [`rmp-serde`](https://crates.io/crates/rmp-serde) v0.2             | &#x274c; | &#x2705; |
| [`serde-json-core`](https://crates.io/crates/serde-json-core) v0.5 | &#x2705; | &#x2705; |
| [`serde-json`](https://crates.io/crates/serde-json) v1             | &#x2705; | &#x2705; |
| [`toml`](https://crates.io/crates/toml) v0.7                       | &#x2705; | &#x2705; |
```

serdect/src/slice.rs
```diff
@@ -147,7 +150,32 @@ where
...
+            }
+
+            fn visit_bytes<E>(self, v: &[u8]) -> Result<Self::Value, E>
+            where
+                E: Error,
+            {
+                Ok(v.into())
+            }
+
+            #[cfg(feature = "alloc")]
```
daxpedda, line 169:
```suggestion
```
The function this is in is already gated by `feature = "alloc"`.

serdect/tests/cbor.rs
```diff
@@ -9,13 +9,18 @@ use serde::Serialize;
...
 
 /// Example input to be serialized.
-const EXAMPLE_BYTES: [u8; 16] = hex!("000102030405060708090A0B0C0D0E0F");
+/// Last byte is `0xFF` to test that no packing is performed for values under 128.
+const EXAMPLE_BYTES: [u8; 16] = hex!("000102030405060708090A0B0C0D0EFF");
 
 /// CBOR serialization of [`EXAMPLE_BYTES`] as a slice.
-const CBOR_SLICE: [u8; 17] = hex!("90000102030405060708090A0B0C0D0E0F");
+/// (first three bits, `0b010` denote a byte string, and the last five, `0b10000` denote the length)
+/// Note the 0x18 marker before 0xFF, denoting that the integers are dynamically sized.
```
daxpedda, line 17: There is no `0x18` marker here.

daxpedda (changes requested):
Could we also add an `assert!()` in the `proptest!`s checking that the length remains the same?

@tarcieri this is definitely a breaking change.
Also apparently CBOR exhibits the same issue as MessagePack. I think we should definitely fix that for arrays as well.

tarcieri:
Since it's a breaking change, please bump the `serdect` version to v0.3.0-pre.

Also please be aware since it's a breaking change that we can't update `elliptic-curve` until `elliptic-curve` v0.14. That will also involve cutting a breaking `crypto-bigint` v0.6 release. All of that is several months away at least.

Alternatively consider adding new APIs and deprecating the old ones.

tarcieri:
Re: array changes, ugh, that just really really sucks. It is just very very unfortunate have to break everyone's serializations to add a completely redundant length prefix to work around deficient format designs/implementations.

That's going to be a very ugly breaking change with wide-ranging impact, which will needlessly make a whole slew of messages longer.

Is there really no better solution? It seems like an absolute last resort.

tarcieri:
Also what's the solution for people who have existing message serializations? Are they just screwed and have to serialize everything?

daxpedda:
> It is just very very unfortunate have to break everyone's serializations to add a completely redundant length prefix to work around deficient format designs/implementations.

Just to clarify, imho it's not a deficiency in the implementations but in Serde itself. If we communicate we want a tuple, they will use their tuple encoding. Serde just doesn't offer a way to communicate homogeneous fixed-size sequences.

> Perhaps it's worth inquiring upstream for potential alternative solutions?

As pointed out above as far as I understand the implementations themselves can't do much about it. I was initially thinking they could detect that the tuple is homogeneous and use their array encoding ... but that's not right either, because some users require specific formats e.g. implementing a specific protocol, that then would break.

It could otherwise be done by making it a non-standard option, e.g. CBOR/MessagePack could have an option to turn off varint encoding in specific cases.

Otherwise it's up to Serde, which isn't gonna happen anytime soon I assume.

> Any of you volunteering to support and explain the changes to people who will inevitably ask why we're breaking all existing serializations of arrays?

I'm happy to respond to issues and help with support (as long as it's Rust related).

> Also what's the solution for people who have existing message serializations? Are they just screwed and have to serialize everything?

I think so? But I guess that's the point of breaking changes?

---

- We could just not do this and leave a note in the Readme stating compatibility issues with certain protocols, e.g. CBOR and MessagePack.
- Alternatively we could just offer two sets of APIs, but this doesn't really make a whole lot of sense because e.g. if you are using `crypto-bigint`, you can't really select which one to use. Unless we use crate features or some other not so viable workaround.

I honestly don't see a good way around this.
The only viable one is just to leave it incompatible with certain formats.

tarcieri:
Since the array serializer under this scheme would lose the advantage of not having a length prefix, perhaps it should be deprecated (or made into a thin wrapper around a slice deserializer).

fjarri:
> Just to clarify, imho it's not a deficiency in the implementations but in Serde itself.

As far as I understand it's a deficiency of specific formats. For MessagePack at the very least, probably for CBOR too. `serde` can't do much about it.

> I think the argument that @fjarri made before is that the array serializer does bound checking, which the slice one does not.

Yes, I would like something that I can use with arrays with the least amount of boilerplate. Ideally, something I can just put into `#[serde(with)]` annotation, for an array, a vector, or a `Box<[u8]>` (this currently works for serializers but not for deserializers). 

> It's a bleak outlook I agree :/

There won't be that many users affected after all, the only thing that breaks is CBOR + `slice` (in this PR specifically, assuming we don't change arrays to slice behavior). The other tested formats are unchanged.

By the way, I did not want to cause all this frustration. I just wanted something I could use in place of https://github.com/nucypher/rust-umbral/blob/master/umbral-pre/src/serde_bytes.rs for my crates without copying it everywhere. I thought `serdect` could be that after some adjustments, but if that does not work, I have no problems just making my own crate.

serdect/README.md
fjarri: That whole paragraph is largely a placeholder, if someone can suggest a better wording for it, I will happily accept.

tarcieri:
I guess we should try to fix it, including arrays, though it will take quite some time to get it rolled out

serdect/README.md
fjarri: > Why is rmp-serde missing a verrsion? 

Forgot to add. Fixed now, and added the links as well.

serdect/tests/cbor.rs
fjarri: Fixed.

serdect/src/slice.rs
fjarri: Fixed.

serdect/README.md
fjarri: Rephrased.

serdect/tests/messagepack.rs
```diff
@@ -0,0 +1,61 @@
+//! bincode-specific tests.
```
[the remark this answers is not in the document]
fjarri: Fixed.

daxpedda:
> > Just to clarify, imho it's not a deficiency in the implementations but in Serde itself.
> 
> As far as I understand it's a deficiency of specific formats. For MessagePack at the very least, probably for CBOR too. `serde` can't do much about it.

I'm pretty sure this is incorrect, MessagePack only uses varint encoding because Serde communicates this array as a tuple, which MessagePack handles differently. The same can be said about CBOR, which functions as desired when we are using the proper encoding that CBOR does provide.

> By the way, I did not want to cause all this frustration. I just wanted something I could use in place of https://github.com/nucypher/rust-umbral/blob/master/umbral-pre/src/serde_bytes.rs for my crates without copying it everywhere. I thought `serdect` could be that after some adjustments, but if that does not work, I have no problems just making my own crate.

I consider this whole issue a bug. We misused Serde's API, is how I see it. So thanks for figuring this all out, reporting it and helping us understand it!

serdect/README.md
```diff
@@ -29,6 +29,18 @@ While this crate can't ensure that format implementations don't perform
...
+
+| Crate                                                              | `array`  | `slice`  |
+|--------------------------------------------------------------------|:--------:|:--------:|
+| [`bincode`](https://crates.io/crates/bincode) v1                   | &#x2705; | &#x2705; |
+| [`ciborium`](https://crates.io/crates/ciborium) v0.2               | &#x274c; | &#x2705; |
+| [`rmp-serde`](https://crates.io/crates/rmp-serde) v0.2             | &#x274c; | &#x2705; |
+| [`serde-json-core`](https://crates.io/crates/serde-json-core) v0.5 | &#x2705; | &#x2705; |
+| [`serde-json`](https://crates.io/crates/serde-json) v1             | &#x2705; | &#x2705; |
+| [`toml`](https://crates.io/crates/toml) v0.7                       | &#x2705; | &#x2705; |
+
```
daxpedda, line 43: I think we are good to go to fix arrays in this PR as well, please correct me if I'm wrong @tarcieri, so this table could probably just list crates that we test against.

fjarri:
> I'm pretty sure this is incorrect, MessagePack only uses varint encoding because Serde communicates this array as a tuple, which MessagePack handles differently.

How would you represent a fixed-size `u8` array without a length specifier in MessagePack format?

daxpedda:
> How would you represent a fixed-size `u8` array without a length specifier in MessagePack format?

I assume with "length specifier" you mean varint encoding? That would be the "bin 8" format. Which is the format `rmp-serde` uses for `encode_bytes()`. This encoding does not use any varint encoding.

The problem is that `encode_tuple()` can't tell `rmp-serde` that this is a homogeneous array of bytes, that is why it chooses an inappropriate MessagePack format.
In the case of `rmp-serde`, it uses either the "fixarray", "array 16" or "array 32" format and then encodes every element with it's own format, which in our case is `u8`. For encoding an individual `u8` `rmp-serde` chooses either "positive fixint", "uint 8", "uint 16", "uint 32" or "uint 64", which is what creates the varint encoding.

So what I'm trying to get here is, that `encode_tuple` doesn't give the correct information to `rmp-serde` to make the best decision here. Even though MessagePack doesn't have a dedicated format for fixed size array, it has the "bin" format, which is the right format to use in this case.
If Serde had a `serialize_fixed_bytes()` function, MessagePack could do the right thing, which in this case is doing the same thing that `serialize_bytes()` does, because the MessagePack format can't optimize this type any further.

---

If you meant with length specifier just the length of the byte string ... then AFAIK MessagePack can't do that.
But it should still pick the best format for the task, which is still "bin" and not "array".

fjarri:
> I assume with "length specifier" you mean varint encoding? That would be the "bin 8" format.

No, the information about the length of the following array. Like, for a Messagepack `"DC0010000102030405060708090A0B0C0D0ECCFF"`, `0x10000` is the length specifier. Bincode does not add that to fixed-size arrays.

daxpedda:
> > If Serde had a serialize_fixed_bytes() function, MessagePack could do the right thing
> 
> Well, it could do an array with a length specifier, [..]

`rmp-serde` does make an array with a length specifier (in MessagePack all arrays use length specifiers), which is the problem, because then every element needs to be encoded separately, which introduces the varint encoding.

>  [..], and `bincode` could do the actual right thing because the format allows it.

Well, Bincode doesn't actually use varint encoding by default, that's why it doesn't hit that problem. Additionally, Bincode's varint encoding doesn't affect `u8`s.

So if you used Bincode with varint encoding, a `[u16; N]` and used `serialize_tuple()`, you would have the exact same problem in Bincode. In this case Serde fails even further, because you can't communicate non-`u8` slices at all, `serialize_bytes()` only takes `&[u8]`.

Generally speaking you want to use varint encoding if you are dealing with individual integers, but not integer strings. Which is exactly what Serde doesn't expose, but both MessagePack and CBOR support (and Bincode would too).

serdect/Cargo.toml
```diff
@@ -30,6 +30,7 @@ serde = { version = "1.0.119", default-features = false, features = ["derive"] }
 serde_json = "1"
 serde-json-core = { version = "0.5", default-features = false, features = ["std"] }
 toml = "0.7"
+rmp-serde = "1"
```
baloo, line 33: Could you relocate this a couple lines above to preserve the ordering?

fjarri: Done

daxpedda:
@fjarri are you still interested in pushing this PR over the finishing line?
According to tarcieri, https://github.com/RustCrypto/formats/pull/1112#issuecomment-1601301905, we have the green light to fix arrays as well now.

fjarri:
I am, but I am currently away on a vacation, and will only be near a
computer next Thursday. Perhaps this can be merged as is, and arrays
treated in a subsequent PR.

tarcieri:
I'd prefer to save this for the next breaking release cycle, so maybe you can finish it up before we merge (no rush)

fjarri:
Made the changes for the arrays. Not sure if `LengthCheck` makes sense, or I should have just copy-pasted the implementation for simplicity. Also note that `deserialize` for slices does not return the reference to the buffer anymore (since it's mutated in place anyway).

serdect/src/array.rs
```diff
@@ -1,11 +1,20 @@
 //! Serialization primitives for arrays.
 
-use core::fmt;
+// Unfortunately, we currently cannot assert generically that we are serializing
+// a fixed-size byte array.
```
daxpedda, line 4: I'm not sure what this sentence says, do you mean to say that we can't know if the tuple type of a serialization format fits our needs?

serdect/tests/cbor.rs
```diff
@@ -9,12 +9,16 @@ use serde::Serialize;
...
+const EXAMPLE_BYTES: [u8; 16] = hex!("000102030405060708090A0B0C0D0EFF");
 
 /// CBOR serialization of [`EXAMPLE_BYTES`] as a slice.
-const CBOR_SLICE: [u8; 17] = hex!("90000102030405060708090A0B0C0D0E0F");
+/// (first three bits, `0b010` denote a byte string, and the last five, `0b10000` denote the length)
+const CBOR_SLICE: [u8; 17] = hex!("50000102030405060708090A0B0C0D0EFF");
 
 /// CBOR serialization of [`EXAMPLE_BYTES`] as an array.
+/// (first three bits, `0b100` denote an array, and the last five, `0b10000` denote the length)
+/// Note the 0x18 marker before 0xFF, denoting that the integers are dynamically sized.
```
daxpedda, line 21: I think this whole comment has to be adjusted now.

serdect/tests/messagepack.rs
```diff
@@ -0,0 +1,61 @@
...
+
+/// Example input to be serialized.
+/// Last byte is `0xFF` to test that no packing is performed for values under 128.
+const EXAMPLE_BYTES: [u8; 16] = hex!("000102030405060708090A0B0C0D0EFF");
+
+/// messagepack serialization of [`EXAMPLE_BYTES`] as a slice.
+const MESSAGEPACK_SLICE: [u8; 18] = hex!("C410000102030405060708090A0B0C0D0EFF");
+
+/// messagepack serialization of [`EXAMPLE_BYTES`] as an array.
+/// Note the 0xCC marker before 0xFF, denoting that the integers are dynamically sized.
```
daxpedda, line 17: Same as before.

(I think we don't need this anymore, because now bytes and slice have the exact same output.)

daxpedda (changes requested):
Sorry for the delay!

Just a couple of documentation and test nits.
Would also like to see an `assert!` in the proptests to make sure the length actually stays the same. Basically testing what this PR is fixing.

serdect/src/array.rs
fjarri: Perhaps it is not very well phrased. By "generically" I meant generically over formats supported by `serde`, not generically in the sense of Rust type system generics.

serdect/tests/cbor.rs
fjarri: Fixed

serdect/tests/messagepack.rs
fjarri: Fixed

daxpedda (approved):

daxpedda (changes requested):
Actually I forgot to mention one last change I would like to see: move the `Visitor` implementation into it's own file. Now that it's equally shared between slice and array implementation, it doesn't make sense that it's living in the `slice` module.

Would also be interesting if we could move `ExactLength` and `UpperBound` into their respective modules.

serdect/src/array.rs
fjarri: Amended the phrasing.

daxpedda (approved):
Thank you!

serdect/src/slice.rs
```diff
@@ -32,91 +36,39 @@ where
...
+            data_type, data_length
+        )
     }
 }
 
 /// Deserialize from hex when using human-readable formats or binary if the
 /// format is binary. Fails if the `buffer` is smaller then the resulting
 /// slice.
-pub fn deserialize_hex_or_bin<'de, D>(buffer: &mut [u8], deserializer: D) -> Result<&[u8], D::Error>
+pub fn deserialize_hex_or_bin<'de, D>(buffer: &mut [u8], deserializer: D) -> Result<(), D::Error>
```
tarcieri, line 64: @fjarri running into problems with this change when upgrading this crate.

This API in particular is for deserializing variable-length data, however removing the `&[u8]` return value means the amount of data that was actually deserialized was lost.

`buffer` is intended to hold a backing buffer which is as large as the deserialized data can possibly be, but it may be smaller, and the return value is there to return the actual amount of data deserialized.

tarcieri: Made a tracking issue: #1322

PR #1115, opened 2023-06-23
CMS `EnvelopedData`

bkstein:
This PR brings a builder for CMS `EnvelopedData` and a `RecipientInfo` trait. Only `KeyTransRecipientInfo` with RSA was implemented for key encryption, but the code is prepared for beeing extended to all other `RecipientInfo`s and arbitrary encryption algorithms. Volunteers are welcome 🙂.

Open issues:
- Encryption is limited to what is supported by the builder(s). This should be more flexible, e.g. by requiring the user to implement the `encrypt()` function.


tarcieri:
@bkstein being able to pass in an RNG is nice for `no_std` targets (or more specifically targets where `getrandom` doesn't build)

One option would be to add a `fn build_with_rng` and gate a provided `fn build` on a `getrandom` feature.

bkstein:
@tarcieri An `fn build_with_rng` should be possible. We already have a corresponding function for the x509 builder, so that would fit nicely.

bkstein:
@tarcireri As for the failing test: I use [`impl TryFrom<String> for PrintableString`](https://github.com/bkstein/RustCrypto-formats/blame/a602f0441305c168a59894d56899e9daadefbc6d/der/src/asn1/printable_string.rs#L224) which was implemented by turbocool3er last month, but is not released, yet. How do I handle this gracefully?

carl-wallace:
This looks like a good start. Two thoughts as it moves forward. As with my suggestion (earlier today) re: signer identifiers, it may be nice to interrogate the key encryption info for the recipient identifier. Similarly, it may be good to define a few traits so algorithm and padding details don't live in the builder object. This will likely matter more as key agreement and KEM support is added but at present OAEP would require a new builder implementation.

bkstein:
> I can cut a release, then you can bump the minor version requirement

@tarcieri I'm not sure, if you saw my 👍? A new release were nice. However, I'm working on the rng feature and I hope, I can push it by the end of this week.

bkstein:
> ... As with my suggestion (earlier today) re: signer identifiers, it may be nice to interrogate the key encryption info for the recipient identifier. Similarly, it may be good to define a few traits so algorithm and padding details don't live in the builder object. This will likely matter more as key agreement and KEM support is added but at present OAEP would require a new builder implementation.

@carl-wallace you address a point, that I would like to improve, too. I also don't like my hardcoded implementation of the encryption. I searched for a way how we could leave the choice for encryption algorithms to the user of this crate. My first idea was to pass an encryptor, but there is no common trait for that. What, if the builder accepted an
```rust
trait Encryptor {
    fn encrypt(data: &[u8]) -> Result<Vec<u8>>;
}
```
which must be implemented by the user? This would keep all encryption details outside the builder.

carl-wallace:
That could work. I wonder if a companion to the signature crate is really what's needed though (maybe as a longer term goal).

tarcieri:
We do have traits like this in the `rsa` crate:

- https://docs.rs/rsa/latest/rsa/traits/trait.RandomizedEncryptor.html
- https://docs.rs/rsa/latest/rsa/traits/trait.RandomizedDecryptor.html
- https://docs.rs/rsa/latest/rsa/traits/trait.Decryptor.html

bkstein:
I made rng an input parameter for key and content encryption. This will probably become obsolete anyway, when and if encryption details are moved out of the builders.
I also finished the `build_pkcs7_scep_pkcsreq` test, which is quite long now for a test. But it demonstrates, what has to be done to generate and read a SCEP pkiMessage.

tarcieri:
@bkstein if you think the PR is ready for review, can you remove draft?

bkstein:
@tarcieri I'd like to test the implementation in our project. Just to make sure, I didn't miss something. This shouldn't take too long. I'll remove the draft then.

carl-wallace:
FWIW, I should be able to give the EnvelopedData some exercise on a project over the next week or two (as part of SCEP execution). 

@bkstein I am curious if you have tried SignedData with a SignerInfo that used SKID for identifying the signer (the openssl test cases look to use issuer/serial). In some interop work today I ran into an issue and it looks like 
```
    #[asn1(context_specific = "0", tag_mode = "EXPLICIT")]
    SubjectKeyIdentifier(SubjectKeyIdentifier),
```
should have been
```
    SubjectKeyIdentifier(SubjectKeyIdentifier),
```

carl-wallace:
Re: SCEP, I had success this morning but had to make one change relative to the test code. The message type attribute should be a PrintableString (see table 1 in section 3.2.1 in RFC8894).  The code that worked for me was:
```
    let mut message_type_value: SetOfVec<AttributeValue> = Default::default();
    let id = PrintableString::try_from(String::from("19")).unwrap();
    message_type_value
        .insert(Any::from(&id))
        .unwrap();
    let message_type = Attribute {
        oid: RFC8894_ID_MESSAGE_TYPE,
        values: message_type_value,
    };
```
This change was in addition to the change noted earlier relative to key identifier variant of SignerIdentifier. Oddly, RecipientIdentifier was fine.

This was part of an implementation of a client for Apple's Over-the-Air profile delivery and configuration protocol to provision a yubikey (with a custom attribute included in the SCEP request to convey an attestation to the CA). I exercised both the SignedData and EnvelopedData builders (as well as some of the untested code in the yubikey crate along with one or two additions there). Those worked well. I have some cleanup to do before contemplating PRs.

bkstein:
> Re: SCEP, I had success this morning but had to make one change relative to the test code. The message type attribute should be a PrintableString (see table 1 in section 3.2.1 in RFC8894). 

I also found that yesterday. I can fix that before removing the draft status.

Great to hear, that you had success. I hope to follow this week with my application.

bkstein:
@baloo When I followed your x509 builders to implement `SignerInfoBuilder`, I also copied the `S::VerifyingKey: EncodePublicKey` trait bound

```rust
impl<'s, S> Builder for CertificateBuilder<'s, S>
where
    S: Keypair + DynSignatureAlgorithmIdentifier,
    S::VerifyingKey: EncodePublicKey,
    ...
```

I removed it now from my `SignerInfoBuilder`, because I don't a public/verifying key.
Was deriving a public key from the private key the only reason for using the `S::VerifyingKey: EncodePublicKey`?

baloo:
We didn't really need to "derive" it (in the sense that it could be held behind HSM or secure device), but that was convenient to encode the Authority Key Identifier (sha1 of the signing public key used to build the trust graph).

baloo:
Just a heads up, I might want to publish a release of CMS by the end of this week (to pull the `SignedData` builder more easily (thank you for doing that!!)).
Let me know if can do anything to help with the review of this one and get it in the same train too :)

I guess you'd need a `const-oid 0.9.4` release with backports of the `RFC8894_ID_*` objects, correct?

carl-wallace:
Re: CMS release, please make sure to include the tagging change for SignerIdentifier that's in this PR.

bkstein:
Re: CMS, we probably should remove setting the signing time attribute first. Or move creation of this attribute into an own method for convenience, if someone should need it. I'm still busy with replacing our rust-openssl implementation with RustCrypto cms. Thus not sure, if I will be able to do it until Friday. But I can try it.

bkstein:
> I guess you'd need a `const-oid 0.9.4` release with backports of the `RFC8894_ID_*` objects, correct?

@baloo Correct. I currently have to use my own defnitions, until the `RFC8894_ID_*` are available.

baloo:
> Re: CMS release, please make sure to include the tagging change for SignerIdentifier that's in this PR.

Good point, I completely missed that. @bkstein do you want to send a separate PR for https://github.com/rustcrypto/formats/commit/29ac6373068510ad679b3ed1fef1e8bc48eaffcf (I can do it, but this will strip authorship in the final commit, up to you).

baloo:
@bkstein https://crates.io/crates/const-oid/0.9.4

cms/Cargo.toml
```diff
@@ -14,29 +14,36 @@ edition = "2021"
 rust-version = "1.65"
 
 [dependencies]
-der = { version = "0.7.6", features = ["alloc", "derive", "oid", "pem"] }
+der = { version = "0.7.7", features = ["alloc", "derive", "oid", "pem"] }
 spki = { version = "0.7" }
 x509-cert = { version = "0.2.3", default-features = false, features = ["pem"] }
 const-oid = { version = "0.9", features = ["db"] } # TODO: path = "../const-oid"
```
baloo, line 20:
```suggestion
const-oid = { version = "0.9.4", features = ["db"] }
```

This would pull any `const-oid <0.10 >= 0.9.4`

baloo: I'm surprised minimal-version CI job  doesn't complain here though.

tarcieri: It's possible a (transitive) dependency already has this locked

baloo: I don't believe this is the case though.

As far as I can tell it pulls `0.9.2`:
```
[[package]]
name = "const-oid"
version = "0.9.2"
```

baloo: well because we sill rely on the duplicated definition ^^ I should learn how to read a diff sometimes.

bkstein:
@baloo @npmccallum During my refactoring I stumbled over [this pathlen type](https://github.com/RustCrypto/formats/blob/98085f45347d099ab822a18f22391607ba6e83f5/x509-cert/src/ext/pkix/constraints/basic.rs#L19):
```rust
/// BasicConstraints as defined in [RFC 5280 Section 4.2.1.9].
///
/// ```text
/// BasicConstraints ::= SEQUENCE {
///     cA                      BOOLEAN DEFAULT FALSE,
///     pathLenConstraint       INTEGER (0..MAX) OPTIONAL
/// }
/// ```
///
/// [RFC 5280 Section 4.2.1.9]: https://datatracker.ietf.org/doc/html/rfc5280#section-4.2.1.9
#[allow(missing_docs)]
pub struct BasicConstraints {
    pub ca: bool,
    pub path_len_constraint: Option<u8>,
}

```

According to the RFC, shouldn't this be a `u32` ? That's also how it is in [rust-openssl](https://github.com/sfackler/rust-openssl/blob/50787ed35bf9efa9dd3cbb1993a2564014b67489/openssl/src/x509/extension.rs#L64)

baloo:
Curious, which RFC defines that as u32? all I could find was [RFC5280](https://datatracker.ietf.org/doc/html/rfc5280#appendix-B):
```
MAX indicates that the upper bound is unspecified.
   Implementations are free to choose an upper bound that suits their
   environment.
```

Is there someone with more than 255 certificates in a chain?!?

bkstein:
@baloo I'm sorry for the confusion. Yes, I deleted the comment. I think, you are right. I read about the meaning of `MAX` after I posted the comment and hoped to get away with it unnoticed. Probably not good style to delete comments, especially, when I mentioned someone. I was mislead by rust-openssl, which defined the `pathlen` as `u32`. I'm currently replacing our rust-openssl usage by RustCrypto. That's where I met this type mismatch. But I would also guess, that a `u8` will be sufficient.

bkstein:
Empty `EncapsulatedContentInfo`s are now allowed. This is required for degenerate certificate-only CMS messages.

cms/src/builder.rs
```diff
@@ -99,7 +109,6 @@ pub struct SignerInfoBuilder<'s, S> {
 impl<'s, S> SignerInfoBuilder<'s, S>
```
tarcieri, line 109: Perhaps this should follow the revised approach in #1161? Though that would require the corresponding changes to the `Builder` trait I suppose.

bkstein: I also think, we should do follow the #1161 `Builder`. But this PR wouldn't work until #1161 is merged. So, it's just a timing problem. It's also possible to merge this PR first and update the CMS crate in #1161, before that is merged.

tarcieri: If you're okay with updating `cms` after #1161 is merged, we can go ahead and merge this as-is.

bkstein: I'm okay with updating `cms` after #1161 was merged. But end of this week I plan to be on holidays for about 4 weeks, so this should be taken into account.

baloo: I'm out for parental leave until the first week of september too, I have some spare time, but kid has some amount of priority here ^^. All that to say, no idea if #1161 will move before you come back.
On top of that, we also need to plan/sequence the next set of breaking changes (If I could throw #1117 in the next train too, that'd be cool).

tarcieri (approved):

PR #1125, opened 2023-06-27
tls_codec: fix deserialization of `Option<T>`

kkohbrok:
The implementation of `Option<T>` had a small bug, where the remainder was not returned correctly. This PR fixes that bug and adds a small test to test for its presence.


franziskuskiefer (approved):
Thanks!

Issue #1109, opened 2023-06-17
Understanding `tls_codec` and it's traits


imor:
I have been working on adding support for the the [Signed Certificate Timestamp extension](https://letsencrypt.org/2018/04/04/sct-encoding.html) in x509-cert crate. Since some data in this extension is TLS encoded I an using the `tls_codec` crate. But I'm a bit confused about the structure of some of the types and traits in `tls_codec`. Specifially:

* What's the difference between `TlsVecU8<T>` and `TlsByteVecU8`. To me it looks like `TlsByteVecU8` can be just a type alias: `type TlsByteVecU8 = TlsVecU8<u8>`. Is this true? If yes, why do we have two entirely separate types?
* What is the difference between `Serialize` and `SerializeBytes` traits (same for `Deserialize` and `DeserializeBytes`)? Are `Serialize` and `Deserialize` meant to be used only in an `std` environment (because they use std::io::{Read, Write})? If yes, why can't `SerializeBytes` and `DeserializeBytes` work for both `std` and `no_std`?
* Why is `DeserializeBytes` impl'd for `TlsVecU8<T>` but `SerializeBytes` is not?

Answers to these would be greatly appreciated. Especially the last one because currently my types implement `Serialize` and `DeserializeBytes` which looks a bit asymmetric to me.

cc @tarcieri @franziskuskiefer

franziskuskiefer:
> * What's the difference between `TlsVecU8<T>` and `TlsByteVecU8`. To me it looks like `TlsByteVecU8` can be just a type alias: `type TlsByteVecU8 = TlsVecU8<u8>`. Is this true? If yes, why do we have two entirely separate types?

It's not a type alias. It's a more efficient version of the same.
`TlsVecU8<T>` has to call `T::tls_(de)serialize` for each element while `TlsByteVecU8` knows `T==u8` and can thus skip the extra function call.
Unfortunately Rust doesn't allow us to overload functions and this is the quickest way around (even if not the most elegant one).

> * What is the difference between `Serialize` and `SerializeBytes` traits (same for `Deserialize` and `DeserializeBytes`)? Are `Serialize` and `Deserialize` meant to be used only in an `std` environment (because they use std::io::{Read, Write})? If yes, why can't `SerializeBytes` and `DeserializeBytes` work for both `std` and `no_std`?

Yes, the way the `Serialize` trait is defined it uses `Read`/`Write`, which requires `std`.
[...]

PR #1132, opened 2023-07-02
Fix TlsByteVecUX deserialization

imor:
This PR fixes the bug found in #1109. It adds a `DeserializeBytes::tls_deserialize` test for each of the `TlsByteVecUX` types and fixes the bug.


franziskuskiefer (approved):
Thanks!

PR #1135, opened 2023-07-02
[breaking change] Add `no_std` support for `tls_codec_derive`

### LINKED ISSUE #1106: Make `tls_codec_derive` compatible with `#![no_std]`
The `tls_codec_derive` crate currently unconditionally refers to items from stdlib. For example [this line uses `std::mem::size_of`](https://github.com/RustCrypto/formats/blob/8bea5699c37a842540fd0efa56227ad89d0df99a/tls_codec/derive/src/lib.rs#L686). Similarly [this line refers to `std::io::Write`](https://github.com/RustCrypto/formats/blob/8bea5699c37a842540fd0efa56227ad89d0df99a/tls_codec/derive/src/lib.rs#L722). These direct unconditional references makes it impossible to use `tls_codec_derive` from a `#![no_std]` crate. The code needs to be updated so that the crate can be used in a `#![no_std]` environment.

To attack this problem two broad techniques will be useful:

1. Wherever possible replace a `std` reference with either a `core` reference or an `alloc` reference. For example, `std::mem::size_of` can be replaced with `core::mem::size_of`. Presence of liballoc can be assumed, see [this comment](https://github.com/RustCrypto/formats/pull/1103#issuecomment-1588094479).
2. For the references which are not reexported from libcore or liballoc, feature gate them on a new `std` feature. For example `std::io::Write` is such a reference. This `std` feature should be enabled if the corresponding `std` feature in the `tls_codec` crate is enabled.

Ideally a failing test case should be written as well but I'm not sure if CI builds the crates in a `#![no_std]` environment or not. Maybe @tarcieri  or @franziskuskiefer know about it.

imor:
This PR adds support for `no_std` for the `tls_codec_derive` crate. A new features named `std` is added to this crate. It will be automatically enabled whenever the `std` feature of the `tls_codec` crate is enabled.

Note that this is a breaking change because the `derive` feature no longer enables the `std` feature.

Fixes #1106


franziskuskiefer (changes requested):
The derive crate is currently not tested on CI for `no_std`.
Can you add this to make sure we CI checks your changes?

imor:
Bump, @franziskuskiefer could you please review the changes.

imor:
@franziskuskiefer the changes you requested have been done, could you please review this PR.

franziskuskiefer (changes requested):
This doesn't build for a `no_std` target like `thumbv7em-none-eabi` right now.

```
error[E0463]: can't find crate for `std`
  --> tls_codec/src/lib.rs:31:1
   |
31 | extern crate std;
   | ^^^^^^^^^^^^^^^^^ can't find crate
   |
```

.github/workflows/tls_codec.yml
```diff
@@ -70,4 +70,4 @@ jobs:
       - run: ${{ matrix.deps }}
       - uses: RustCrypto/actions/cargo-hack-install@master
       - run: cargo hack test --feature-powerset
-      - run: cargo hack test -p tls_codec_derive --test encode\* --test decode\*
+      - run: cargo hack test -p tls_codec_derive --feature-powerset --test encode\* --test decode\*
```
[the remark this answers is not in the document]
imor: Included the `derive` feature when building `no_std` targets.

imor:
> This doesn't build for a `no_std` target like `thumbv7em-none-eabi` right now.
> 
> ```
> error[E0463]: can't find crate for `std`
>   --> tls_codec/src/lib.rs:31:1
>    |
> 31 | extern crate std;
>    | ^^^^^^^^^^^^^^^^^ can't find crate
>    |
> ```

Did you build with the command line `cargo build --target thumbv7em-none-eabi`? This will fail because this target is `no_std` and `std` feature is enabled by default. I tried with `cargo build --target thumbv7em-none-eabi --no-default-features` and it builds.

.github/workflows/tls_codec.yml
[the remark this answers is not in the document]
franziskuskiefer: Thanks

.github/workflows/tls_codec.yml
```diff
@@ -70,4 +70,4 @@ jobs:
       - run: ${{ matrix.deps }}
```
franziskuskiefer, line 70: Looks like there's a typo in the matrix "targets" -> "target".

franziskuskiefer (approved):
Thanks!
Sorry for the delay. Let's get this in.

.github/workflows/tls_codec.yml
imor: Since this PR is merged, I'll handle the typo as part of another PR.

franziskuskiefer: Yes, for some reason this works even with this typo. I'm not entirely sure why. But the linter complained.

PR #1148, opened 2023-07-11
cms: Don't insert signing time attribute by default.

bkstein:
The current implementation of `SignedDataBuilder` inserted a signing time attribute by default. This was removed, as it is not required ans might even be unwanted.

### DIFF
--- a/cms/src/builder.rs
+++ b/cms/src/builder.rs
@@ -204,23 +204,9 @@ where
             },
         };
 
-        // We set the signing time attribute. In this case, signed attributes are used and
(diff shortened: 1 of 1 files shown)


tarcieri (approved):

baloo (approved):
I don't see a requirement to have it included by default in RFC 5652

carl-wallace (approved):
Looks good to me (and worked in context of the SCEP usage I mentioned before). I had a question re: whether always using signed attributes was OK but nothing in 5652 compels one to not use content type and message digest attributes, so this seems fine.

PR #1154, opened 2023-07-15
pkcs12: KDF support

xemwebe:
Implementation of the KDF in RFC7292 Appendix B


pkcs12/Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "pkcs12"
-version = "0.0.0"
+version = "0.1.0"
```
tarcieri, line 3: Please don't bump the version. We do separate release PRs for this.

pkcs12/src/kdf.rs
```diff
@@ -0,0 +1,110 @@
+//! Implementation of the key derivation function
+//! [RFC 7292 Appendix B](https://datatracker.ietf.org/doc/html/rfc7292#appendix-B)
+
+use alloc::vec::Vec;
+use digest::{core_api::BlockSizeUser, Digest, FixedOutputReset, OutputSizeUser, Update};
+use zeroize::Zeroize;
+
+/// Transform a utf-8 string in a unicode (utf16) string as binary array.
+/// The Utf16 code points are stored in big endian format with two trailing zero bytes.
+pub fn str_to_unicode(utf8_str: &str) -> Vec<u8> {
```
tarcieri, line 10: I don't think this needs to be `pub`?

```suggestion
fn str_to_unicode(utf8_str: &str) -> Vec<u8> {
```

pkcs12/src/kdf.rs
```diff
@@ -0,0 +1,110 @@
...
+    let mut utf16_bytes = Vec::new();
+    // reserve max number of required bytes to avoid re-allocation
+    utf16_bytes.reserve(utf8_str.len() * 2 + 2);
+    for code_point in utf8_str.encode_utf16() {
```
tarcieri, line 14: Another way to do this:

```suggestion
    for code_point in utf8_str.encode_utf16().chain(Some(0)) {
```

...which should eliminate the need to `push(0)` twice at the end.

pkcs12/tests/kdf.rs
```diff
@@ -0,0 +1,130 @@
+use cfg_if::cfg_if;
```
tarcieri, line 1:
```suggestion
#![cfg(feature = "alloc")]
```

pkcs12/tests/kdf.rs
```diff
@@ -0,0 +1,130 @@
...
+
+cfg_if! {
+    if #[cfg(feature="alloc")] {
+
+        #[test]
+        fn pkcs12_key_derive_sha256() {
```
tarcieri, line 7: Can you provide some provenance information for these test vectors?

pkcs12/Cargo.toml
xemwebe: done

pkcs12/src/kdf.rs
xemwebe: done

pkcs12/tests/kdf.rs
xemwebe: done

pkcs12/tests/kdf.rs
xemwebe: I have basically written a small programming calling the openssl API to get the reference results. I will see if I find some more transparent way to document the validity of the tests.

tarcieri: Can you check in the code you used to generate it? (provided it's Rust)

Otherwise it might be possible to generate such vectors using the `openssl` CLI, possibly using a shell script to generate them, which would be nice.

`openssl asn1parse` is quite useful for inspecting the contents of ASN.1-encoded documents.

xemwebe: You can find the tool I used to generate the test cases here: https://github.com/xemwebe/test_pkcs12_kdf/tree/main/src .
However, recent versions of `openssl` does not seem to support the `whirlpool` algorithm by default any longer.
Unfortunately, I haven't found a way yet to generate these vectors with the `openssl` CLI.

pkcs12/Cargo.toml
```diff
@@ -14,6 +14,20 @@ readme = "README.md"
 edition = "2021"
 rust-version = "1.65"
 
+[features]
+alloc = []
```
tarcieri, line 18: No need to feature-gate `alloc`. This crate will definitely have a hard dependency on `alloc` by way of `cms`.

tarcieri: That said, feature-gating `kdf` might make sense. You could make the `digest` dependency optional that way.

pkcs12/src/lib.rs
```diff
@@ -14,4 +14,16 @@
...
+
+cfg_if! {
+    if #[cfg(feature = "alloc")] {
+
+        #[macro_use]
+        extern crate alloc;
+
+        pub mod kdf;
+    }
+}
```
tarcieri, line 29: Per my comment about a hard dependency on `alloc` above:

```suggestion
extern crate alloc;

pub mod kdf;
```

pkcs12/tests/kdf.rs
tarcieri: Okay, can you at least add a comment with a link to your tool?

pkcs12/src/kdf.rs
```diff
@@ -0,0 +1,108 @@
+//! Implementation of the key derivation function
+//! [RFC 7292 Appendix B](https://datatracker.ietf.org/doc/html/rfc7292#appendix-B)
+
+use alloc::vec::Vec;
+use digest::{core_api::BlockSizeUser, Digest, FixedOutputReset, OutputSizeUser, Update};
+use zeroize::Zeroize;
+
+/// Transform a utf-8 string in a unicode (utf16) string as binary array.
+/// The Utf16 code points are stored in big endian format with two trailing zero bytes.
+fn str_to_unicode(utf8_str: &str) -> Vec<u8> {
```
baloo, line 10: The RFC talks about `BMPString` which happens to be UTF16, but renaming that `str_to_bmpstring` or something would be great I think.

tarcieri: Hmm, I guess we should really add `BMPString` support to `der`

baloo: I'm a bit uncomfortable with the use of `str` for inputs here. This will backfire with any utf8mb4 character. https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=c4edb02a6d7bf8fb45147475e43f1930

I don't know what the best course of action is here though, the best I can think of would be to add a `der::asn1::BmpString` but it is meant to only be subtyped from what I can read (https://www.oss.com/asn1/resources/asn1-made-simple/asn1-quick-reference/bmpstring.html)

pkcs12/src/kdf.rs
```diff
@@ -0,0 +1,108 @@
...
+    }
+    for i in 0..plen {
+        init_key[slen + i] = pass_utf16[i % pass_utf16.len()];
+    }
+    pass_utf16.zeroize();
+
+    let id_block = match id {
+        Pkcs12KeyType::EncryptionKey => vec![1u8; block_size],
+        Pkcs12KeyType::Iv => vec![2u8; block_size],
+        Pkcs12KeyType::Mac => vec![3u8; block_size],
```
baloo, line 67: RFC calls for:
```
   1.  Construct a string, D (the "diversifier"), by concatenating v/8
       copies of ID.
```

I think that works here because of the block size of sha256 lines up, but that doesn't look correct.

pkcs12/src/kdf.rs
```diff
@@ -0,0 +1,108 @@
...
+    let plen = block_size * ((pass_utf16.len() + block_size - 1) / block_size);
+    let ilen = slen + plen;
+    let mut init_key = vec![0u8; ilen];
+    for i in 0..slen {
+        init_key[i] = salt[i % salt.len()];
+    }
+    for i in 0..plen {
+        init_key[slen + i] = pass_utf16[i % pass_utf16.len()];
+    }
+    pass_utf16.zeroize();
```
baloo, line 62: This looks like step2 & 3:
```


   2.  Concatenate copies of the salt together to create a string S of
       length v(ceiling(s/v)) bits (the final copy of the salt may be
       truncated to create S).  Note that if the salt is the empty
       string, then so is S.

   3.  Concatenate copies of the password together to create a string P
       of length v(ceiling(p/v)) bits (the final copy of the password
       may be truncated to create P).  Note that if the password is the
       empty string, then so is P.
 ```
 
 Could you move code around and copy the algorithm from the RFC as comments?

File: pkcs12/src/kdf.rs [hunk 11]
```diff
@@ -0,0 +1,108 @@
...
+    let id_block = match id {
+        Pkcs12KeyType::EncryptionKey => vec![1u8; block_size],
+        Pkcs12KeyType::Iv => vec![2u8; block_size],
+        Pkcs12KeyType::Mac => vec![3u8; block_size],
+    };
+
+    let mut m = key_len;
+    let mut n = 0;
+    let mut out = vec![0u8; key_len];
+    loop {
```
baloo, line 73: RFC calls for a `c = ceiling(n/u)` and for the loop to be:
```
for _i in 1..=c {
}
```

pkcs12/src/kdf.rs
baloo: (sorry last comment was sent before I refreshed the page, well, go for a `BmpString` then)

pkcs12/tests/kdf.rs
xemwebe: I have added a link to reference the test program.

pkcs12/src/kdf.rs
xemwebe: The format of an ASN1 `BMPString` differs slightly from what is here required: It has a leading id byte `0x1e`, one or bytes indicating the length in bytes, and no terminating zeros. I.e. using `BMPString` would also requires some pre-processing. Multi-Byte UTF8 characters should work fine here, at least the example you provided gives the exact same result als openssl's function `PKCS12_gen_key_utf8` (see the added test case and the respective test program).

Openssl provides similar functions for `PKCS12_gen_key_uni` and `PCKS12_gen_asc` for unicode and ASCII string passwords (see here: [https://www.openssl.org/docs/man3.0/man3/PKCS12_key_gen_utf8.html](https://www.openssl.org/docs/man3.0/man3/PKCS12_key_gen_utf8.html)) Maybe a similar approach could be taken here.

However, I am not sure how to interpret the remark "There are no Unicode byte order marks." in the RFC.

baloo: [last commit](https://github.com/RustCrypto/formats/pull/1154/commits/282baa59bac74845c280308877f291a57559933f) does not really fix the issue with unicode characters that do not encode on two bytes.
[RFC 7292](https://datatracker.ietf.org/doc/html/rfc7292#appendix-B.1) does not really specify the behavior with those.
And I would be in favor of outright rejecting them (or maybe hide a `From<Vec<u8>>` under `hazmat` feature flag for a `der::asn1::BmpString`).

tarcieri: A BOM would precede the string, and if it had one in this case would be `FEFF` to indicate big endian.

`BmpString` (if it existed) would still be a nice place to handle all of the string conversions. The header doesn't matter: we can still use [`EncodeValue`](https://docs.rs/der/latest/der/trait.EncodeValue.html) to avoid it, and two trailing digest bytes are easily added to the `Digest` prior to finalization.

We can worry about that all later though. It isn't needed for an initial PR.

pkcs12/src/kdf.rs
xemwebe: `v` is the block size in bits, i.e. the lengths of this vector should be the block size in bytes, which is exactly `block_size`. The RFC also assumes `v` is a multiple of 8.  So I would think this is correct. Or do I miss something?

pkcs12/src/kdf.rs
xemwebe: I have added comments to the code describing the single steps of the implementation, as far as possible since I derive at some point slightly from the algorithm.

pkcs12/src/kdf.rs
xemwebe: Using a `loop` instead has the advantage that the last step 6. could be skipped.

pkcs12/src/kdf.rs
```diff
@@ -0,0 +1,145 @@
...
+use alloc::{vec, vec::Vec};
+use digest::{core_api::BlockSizeUser, Digest, FixedOutputReset, OutputSizeUser, Update};
+use zeroize::Zeroize;
+
+/// Transform a utf-8 string in a unicode (utf16) string as binary array.
+/// The Utf16 code points are stored in big endian format with two trailing zero bytes.
+fn str_to_unicode(utf8_str: &str) -> Vec<u8> {
+    let mut utf16_bytes = Vec::new();
+    // reserve max number of required bytes to avoid re-allocation
+    utf16_bytes.reserve(utf8_str.len() * 2 + 2);
```
tarcieri, line 13:
```suggestion
    let mut utf16_bytes = Vec::with_capacity(utf8_str.len() * 2 + 2);
```

pkcs12/Cargo.toml
xemwebe: I have added a new feature flag `kdf` for the kdf implementation and `digest` depends on that.
However, I had to put the `extern crate alloc` behind the same feature flag to make to linter happy. It could be removed once other functionality have been added to the crate.

xemwebe:
> One minor comment on this PR. Using a string password is out of sync with how decrypt_in_place is implemented in the pkcs5 crate.

I have factored out the conversion from utf8 to byte array and added an additional function `derive_key_utf8` that accepts a string.

tarcieri:
> Using a string password is out of sync with how decrypt_in_place is implemented in the pkcs5 crate.

PKCS#5 doesn't stipulate a string encoding for the password, but the PKCS#12 KDF does (which is arguably a design flaw in the PKCS#12 KDF, but it would only be one of many).

I think we might consider making the password argument a `der::BmpString` whenever that actually exists, then we can place all of the APIs for transcoding from `&str` and so forth there. But as I mentioned earlier, I don't think any of that is a blocker for this PR.

carl-wallace:
Thanks. I had missed that detail, but ran into the difference relative to pkcs5. It's not really an issue either way.

tarcieri:
For a common API you'd probably want to accept a `&str`, treating it as an `&[u8]` for PKCS#5 and transcoding to `BmpString` (prospectively) for the PKCS#12 KDF.

pkcs12/Cargo.toml
```diff
@@ -14,6 +14,20 @@ readme = "README.md"
 edition = "2021"
 rust-version = "1.65"
 
+[dependencies]
+cfg-if = "1.0.0"
```
tarcieri, line 18: It seems like you're not using this anymore.

tarcieri (approved):
Will go ahead and merge this and do a few local cleanups

PR #1163, opened 2023-07-23
pkcs8: eagerly decode PEM labels

tarcieri:
This should give better errors for invalid PEM type labels even if there are subsequent Base64 processing errors, which might occur if e.g. Base64 is wrapped at a nonstandard width.

Notably such errors will include the expected PEM type label.


tarcieri:
Perhaps `SecretDocument` could have a better API here that passes in the expected type label when decoding.

PR #1164, opened 2023-07-24
der: add `BmpString`

tarcieri:
Adds initial support for a Basic Multilingual Plane (a.k.a. UCS-2) string type.

### DIFF
--- a/der/src/asn1.rs
+++ b/der/src/asn1.rs
@@ -6,6 +6,8 @@ mod internal_macros;
 
 mod any;
 mod bit_string;
+#[cfg(feature = "alloc")]
+mod bmp_string;
 mod boolean;
 mod choice;
 mod context_specific;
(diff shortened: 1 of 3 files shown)


baloo:
It looks good, the one change I'd like to introduce would be to avoid silently accepting chars that encode over more than 2 bytes (which this currently does).

I'd prefer that to be a legitimate choice by the consumer of the API if possible (hence the `hazmat` flag and the `new_unchecked` naming), something like: https://github.com/RustCrypto/formats/commit/8b6f5f5701b5186e754e778a3d4630230bc912bb

tarcieri:
@baloo I'm not terribly well versed in UCS-2/UTF-16. Is the issue there unpaired surrogates or something different?

Would composing `from_utf8` in terms of `from_ucs2` catch it, or do we need to check the value of each decoded character (in `from_ucs2`) to make sure it doesn't exceed `u16::MAX`?

carl-wallace:
Looks good to me (no opinion re: encoding to more than two bytes). I dropped it into the work-in-progress P12 code and it worked fine for the (simple) friendly name attributes in that set.

baloo:
@tarcieri I thought there were issues with the back and forth between `str `and `BmpString` but there isn't any. Disregard my previous comment.

baloo (approved):

PR #1180, opened 2023-08-01
Remove proc-macro-error dependency

### LINKED ISSUE #1173: der_derive pulls in both syn 1.x and 2.x
It uses syn 2.x directly, but syn 1.x through `proc-macro-error`. `proc-macro-error` repo is [here](https://gitlab.com/CreepySkeleton/proc-macro-error) if you want to look at the state of `proc-macro-error` switching to syn 2.x. I don't think it's looking good, and it seems like you don't use it a ton, so using syn's builtin error handling functionality instead is what I would do if I had some time to work on this.

jplatte:
Closes #1173.

I chose to minimize the diff by introducing a minimal `abort!` macro that early-returns like `anyhow::bail!` or similar macros, which reduces the amount of code that has to change a lot.

Further refactorings are possible (in particular, I personally prefer to collect all attribute errors and emitting them individually where possible, instead of short-circuiting on the first error), but I didn't do any of that in this PR for the same reason. Let me know if you're interested in anything like that.

### DIFF
--- a/der/derive/src/lib.rs
+++ b/der/derive/src/lib.rs
@@ -121,6 +121,12 @@
     unused_qualifications
 )]
 
+macro_rules! abort {
+    ( $tokens:expr, $message:expr $(,)? ) => {
+        return Err(syn::Error::new_spanned($tokens, $message))
+    };
+}
+
 mod asn1_type;
 mod attributes;
 mod choice;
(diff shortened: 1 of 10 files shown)


der/derive/src/lib.rs
```diff
@@ -121,6 +121,12 @@
     unused_qualifications
 )]
 
+macro_rules! abort {
```
baloo, line 124: This got me confused for a minute. I was wondering where `abort` was coming from since you were deleting `proc-macro-error`.

baloo (approved):

tarcieri (approved):

PR #1181, opened 2023-08-01
base64ct: pin rustix to 0.37 to preserve MSRV 1.60

tarcieri:
Otherwise we get test failures on newer versions which are MSRV 1.63


baloo:
I'd rather do https://github.com/RustCrypto/formats/pull/1179/commits/df0d7f4ff0f06d54e6d0a938819a237c8f15a997

PR #1182, opened 2023-08-02
const-oid/v0.9.5 backporting rfc8410 OIDs

flihp:
backport rfc8410 OIDs into new release from const-oid/v0.9 branch following form from #1139

### DIFF
--- a/.github/workflows/workspace.yml
+++ b/.github/workflows/workspace.yml
@@ -25,7 +25,7 @@ jobs:
           components: clippy
           override: true
           profile: minimal
-      - run: cargo clippy --all-features
+      - run: cargo clippy
 
   doc:
     runs-on: ubuntu-latest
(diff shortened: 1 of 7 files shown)


flihp:
The workspace clippy check is failing here like it did in #1139: https://github.com/RustCrypto/formats/pull/1139#issuecomment-1629602079. I'm happy to help fix that up if necessary.

tarcieri:
Looks like an MSRV-related failure. You can probably avoid it by running clippy without the `arbitrary` feature.

baloo:
I also ignored the clippy error in #1139. Speaking personally, I'm fine with letting it fail.
If you're alright with that @tarcieri I'll merge that (bumping the date) and publish the release.

flihp:
This PR has been updated with the workaround suggested by @tarcieri. I've also created a new issue to describe the root cause and make the info a bit more visible.

tarcieri (approved):

PR #1184, opened 2023-08-05
der_derive: fix doc typo and use a valid tag number

eric-seppanen:
Add the missing close-parenthesis.

Also change the number in the example, because 42 isn't a valid tag, and will result in a vague compile-time error if the reader pastes this example into their code.


tarcieri:
Thanks!

PR #1195, opened 2023-08-09
pkcs5: remove lifetimes

tarcieri:
Removes lifetimes from all types in the `pkcs5` crate, making them own their data.

cc @baloo

### DIFF
--- a/pkcs5/src/pbes2/kdf/salt.rs
+++ b/pkcs5/src/pbes2/kdf/salt.rs
@@ -1,0 +1,102 @@
+//! Salt storage buffer which works on heapless `no_std` targets.
+// TODO(tarcieri): use `ArrayVec` when it's available in `core`.
+
+use core::fmt;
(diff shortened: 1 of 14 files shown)


pkcs5/src/pbes2/kdf/salt.rs
```diff
@@ -0,0 +1,104 @@
+//! Salt storage buffer which works on heapless `no_std` targets.
+// TODO(tarcieri): use `ArrayVec` when it's available in `core`.
```
tarcieri, line 2: It might be worth considering making a crate like `arrayvec` or `tinyvec` an optional dependency of `der` so it can implement much of the boilerplate below.

There are a few places we use fixed-sized buffers like this where it would be nice, I believe.

baloo:
Ha, well. That makes it easy, but we lose the `*Ref` and `*Owned` aliases.

baloo (approved):

PR #1201, opened 2023-08-19
Pin upper version of serde to <1.0.172

newpavlov:
`serde v1.0.172` and later include pre-compiled binaries which is a security hazard. So until the decision gets reverted, I believe it's worth to pin upper version of `serde`. This approach may cause issues if a different crate in someone's dependency tree will depend on a post-1.0.172 version of `serde`, but I think this issue is small enough when compared to the security concerns. Also, a number of other crates in the ecosystem follow this approach, so we are not alone.

`tai64` depends on `serde`, but does not use the `derive` feature, so pinning is not needed there. `serdect` uses `derive` only for dev builds, but I've pinned the version just in case.

More information and discussion about the `serde` change can be found in https://github.com/serde-rs/serde/issues/2538.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -15,9 +15,9 @@ dependencies = [
 
 [[package]]
 name = "aho-corasick"
-version = "1.0.2"
+version = "1.0.4"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "43f6cb1bf222025340178f382c426f13757b2960e89779dfcb319c32542a5a41"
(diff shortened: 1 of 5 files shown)


tarcieri (approved):

tarcieri:
@newpavlov have you confirmed this actually works when the derive feature is disabled?

newpavlov:
I am not sure I understand your concern. The `derive` feature gets unconditionally enabled, if `serde` feature is on. `serdect` does not enable `derive`, but it's widely used by downstream crates and I am not sure if they do not enable it.

newpavlov:
@tarcieri 
I can not publish `serdect` since you are the only owner, so I will leave it to you.

PR #1203, opened 2023-08-21
Update clippy version in CI and fix new lints

### LINKED ISSUE #1187: Update Clippy and fix lints
I get a whole bunch of warnings from Clippy on the current stable toolchain. It would be nice to fix them.

newpavlov:
Closes #1187 

TODO:
-[ ] Fix remaining lints in `der`
-[ ] Update CI config

### DIFF
--- a/base16ct/src/error.rs
+++ b/base16ct/src/error.rs
@@ -27,6 +27,6 @@ impl std::error::Error for Error {}
 
 impl From<Error> for core::fmt::Error {
     fn from(_: Error) -> core::fmt::Error {
-        core::fmt::Error::default()
+        core::fmt::Error
     }
 }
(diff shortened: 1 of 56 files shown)


tarcieri:
The `der` related problems are actually a regression/missing feature in clippy, now that `arithmetic_side_effects` has subsumed the `integer_arithmetic` lint: https://github.com/rust-lang/rust-clippy/issues/11220

It assumes any math with `core::ops` traits is unsafe, even on user-defined types which prevent those problems (in the case of `der`, with fallible results to arithmetic operations).

There's a suggestion to add a feature to `clippy` to hint to it that a given type's arithmetic is safe, but right now there's no solution other than disabling the lint.

newpavlov:
@tarcieri 
So can I merge it or would you like to make some changes?

tarcieri:
Ideally I'd like to see `arithmetic_side_effects` fixed so it provides the same functionality `integer_arithmetic` had before it was replaced. Otherwise merging will be a regression.

But if you want to merge it, I guess that's fine. It's just an annoying regression.

tarcieri:
It just needs a new annotation for arithmetic-safe newtypes as described in https://github.com/rust-lang/rust-clippy/issues/11220#issuecomment-1649638177

PR #1209, opened 2023-08-29
pkcs12: use `der::BmpString` for KDF

tarcieri:
Uses the `BmpString` type to implement transcoding from UTF-8 to BMP/UCS-2.

This type notably restricts the use of UTF-16 surrogate pairs, something the previous implementation was not doing correctly. This entailed making the methods which transcode UTF-8 fallible, since we need to handle the error case that they encode characters outside BMP/UCS-2.

cc @xemwebe

### DIFF
--- a/pkcs12/tests/kdf.rs
+++ b/pkcs12/tests/kdf.rs
@@ -18,17 +18,19 @@ fn pkcs12_key_derive_sha256() {
             Pkcs12KeyType::EncryptionKey,
             100,
             32
-        ),
+        )
+        .unwrap(),
         hex!("fae4d4957a3cc781e1180b9d4fb79c1e0c8579b746a3177e5b0768a3118bf863")
     );
 
     assert_eq!(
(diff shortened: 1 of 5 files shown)


pkcs12/tests/kdf.rs
```diff
@@ -150,14 +162,12 @@ fn pkcs12_key_derive_special_chars() {
...
-        hex!("d01e72a940b4b1a7a5707fc8264a60cb7606ff9051dedff90930687d2513c006")
-    );
+    assert!(derive_key_utf8::<sha2::Sha256>(
+        PASS_SHORT,
+        &SALT_INC,
+        Pkcs12KeyType::EncryptionKey,
+        100,
+        32
+    )
+    .is_err()); // Emoji is not in the Basic Multilingual Plane
```
tarcieri, line 172: Note: now errors on the emoji test. I believe this is the correct behavior since the emoji is not valid BMP/UCS-2

PR #1211, opened 2023-08-30
fix(#1210): fix impl of SubjectDirectoryAttributes

### LINKED ISSUE #1210: Implementation of `x509_cert::ext::pkix::SubjectDirectoryAttributes` is wrong/inaccurate
At recent master branch at cadf42a1f43bcbd19ec5a34e60a21596a737ba74

According to https://datatracker.ietf.org/doc/html/rfc5280#section-4.2.1.8:
```
SubjectDirectoryAttributes ::= SEQUENCE SIZE (1..MAX) OF Attribute
```
And https://datatracker.ietf.org/doc/html/rfc5280#appendix-A.2
```
-- attribute data types

Attribute               ::= SEQUENCE {
      type             AttributeType,
      values    SET OF AttributeValue }
            -- at least one value is required
```

**But** in code:
https://github.com/RustCrypto/formats/blob/cadf42a1f43bcbd19ec5a34e60a21596a737ba74/x509-cert/src/ext/pkix.rs#L107-L115
It's implemented by using `Vec<AttributeTypeAndValue>` , where `AttributeTypeAndValue` is:
```
AttributeTypeAndValue ::= SEQUENCE {
  type     AttributeType,
  value    AttributeValue
}
```

Although `AttributeTypeAndValue` is super set of `Attribute`, I still suggest to fully follow RFC and make implementation more precise so it avoid user from using it in a wrong way.

According to RFC, it should be `Vec<Attribute>`

Taowyoo:
This PR fixes #1210

### DIFF
--- a/x509-cert/src/ext/pkix.rs
+++ b/x509-cert/src/ext/pkix.rs
@@ -10,7 +10,7 @@ mod authkeyid;
 mod keyusage;
 mod policymap;
 
-use crate::attr::AttributeTypeAndValue;
+use crate::attr::Attribute;
 
 pub use access::{AccessDescription, AuthorityInfoAccessSyntax, SubjectInfoAccessSyntax};
 pub use authkeyid::AuthorityKeyIdentifier;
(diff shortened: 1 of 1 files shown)


tarcieri:
Though this does appear to be a bug, it's also a breaking change to incorporate the fix.

We can merge it as part of the next development cycle when we're making breaking changes.

Taowyoo:
sounds good

tarcieri:
@Taowyoo We’re still interested in this, FYI

Taowyoo:
Hi @tarcieri, thank you for reminding me. I accidentally close this PR when cleaning repositories. I will create another one with same commit soon.

Taowyoo:
^^^
https://github.com/RustCrypto/formats/pull/1286

PR #1212, opened 2023-09-03
const-oid: add `ObjectIdentifierRef`; use for `db`

tarcieri:
Adds a type alias for `ObjectIdentifier<&[u8]>` for OIDs whose backing storage is a byte slice.

Changes the `db` to use OIDs backed by `'static` byte slices, which makes them smaller than if they use a fixed-size buffer.

Also makes more methods generic so e.g. `Eq`/`PartialEq` work between `ObjectIdentifier` and `ObjectIdentifierRef`.


tarcieri:
It might make sense to change the database from `const` to `static` so the memory is reused

tarcieri:
Hmm, there is one big disadvantage to [d16bd0b](https://github.com/RustCrypto/formats/pull/1212/commits/d16bd0b55608f084b0d1612ebc0794e7aaafca8c): statics can't be used in constants, so you couldn't reference OIDs from the database in other constants.

Going to revert that for now and we can consider it separately.

PR #1240, opened 2023-10-19
Draft: PKCS#10 support

bgrieder:
This is an initial commit to support PKCS#10: Certificate Signing Requests.
 Tests need to be finalized and augmented.

Should we have helper methods to facilitate CSR creations (much like the openssl prompts) ? And if yes, where should they be located?


pkcs10/src/attribute.rs
```diff
@@ -0,0 +1,310 @@
+//! Attribute-related definitions as defined in X.501 (and updated by RFC 5280).
```
tarcieri, line 1: This module seems to overlap considerably with https://docs.rs/x509-cert/latest/x509_cert/attr/index.html

tarcieri:
I think this is already implemented as https://github.com/RustCrypto/formats/blob/52656ae/x509-cert/src/request.rs#L63

bgrieder:
Yes. It is already implemented in x509 ! 
I did not find it when looking for it because a placeholder link in the `formats` doc for PKCS#10 points to nowhere. Maybe there should be a link at the top of the `formats` repo or some other mention in the doc to look for PKCS#10 inside x509

baloo:
as for the builder: https://docs.rs/x509-cert/latest/x509_cert/builder/struct.RequestBuilder.html (feature flag `builder`)

PR #1241, opened 2023-10-25
x509-ocsp/remove lifetimes and implement AsExtension

bhesh:
This version reorganizes x509-ocsp into separate files for readability and makes use of alloc to remove lifetimes. It also implements an OCSP version of AsExtension for all possible Extensions outlined in RFC-6960 (in a similar fashion to x509-cert) to be used in a future builder. Optional rand_core feature for generating nonces.


tarcieri:
This looks great at first glance! Can do a more in-depth review later

x509-ocsp/src/ext.rs
```diff
@@ -0,0 +1,221 @@
...
+    /// Creates a Nonce object given a random generator and a length
+    #[cfg(feature = "rand_core")]
+    pub fn generate<R: CryptoRngCore>(rng: &mut R, length: usize) -> Result<Self, der::Error> {
+        let mut bytes = Vec::with_capacity(length);
+        let mut random = [0u8; 32];
+        while bytes.len() < length {
+            rng.fill_bytes(&mut random);
+            bytes.extend_from_slice(&random);
+        }
+        Ok(Self(OctetString::new(bytes)?))
```
baloo, line 71:
```suggestion
        Self::new(bytes)
```

x509-ocsp/src/ext.rs
```diff
@@ -0,0 +1,221 @@
...
+    Sequence, ValueOrd,
+};
+use spki::AlgorithmIdentifierOwned;
+use x509_cert::{
+    ext::{pkix::AuthorityInfoAccessSyntax, Extension},
+    impl_newtype,
+    name::Name,
+};
+
+#[cfg(feature = "rand_core")]
```
baloo, line 22: wait, we don't need to explicitly declare the feature in the `Cargo.toml`? :exploding_head:

x509-ocsp/src/ext.rs
```diff
@@ -0,0 +1,221 @@
...
+use rand_core::CryptoRngCore;
+
+/// Trait to be implemented by extensions to allow them to be formated for x509 OCSP
+/// requests/responses.
+pub trait AsExtension: AssociatedOid + der::Encode {
```
baloo, line 27: Is there a reason we can't pull [`x509_cert::ext::AsExtension`](https://docs.rs/x509-cert/latest/x509_cert/ext/trait.AsExtension.html) ?

Besides the `const` that looks very similar.

bhesh: It can be used. The extra subject/extensions arguments for judging criticality didn't seem relevant at all to OCSP though.

x509-ocsp/src/basic.rs
```diff
@@ -0,0 +1,268 @@
...
+#[derive(Clone, Debug, Copy, PartialEq, Eq, Enumerated)]
+#[asn1(type = "INTEGER")]
+#[repr(u8)]
+pub enum Version {
+    /// Version 1 (default)
+    V1 = 0,
+}
+
+impl Default for Version {
+    fn default() -> Self {
```
baloo, line 36: I think this can be replaced like:
```
#[derive(Default)]
pub enum Version {
    #[default]
     V1 = 0
}
```

https://doc.rust-lang.org/std/default/trait.Default.html#enums

This came around in https://blog.rust-lang.org/2022/06/30/Rust-1.62.0.html but there should have an MSRV of 1.65 because of `x509-cert`.

baloo (approved):

x509-ocsp/src/ext.rs
tarcieri: Yeah, features and optional dependencies are somewhat confusingly merged into the same namespace, which is annoying because every optional dependency becomes a feature whether you want it to be or want to hide it as an implementation detail.

The `dep:` syntax somewhat fixes that.

x509-ocsp/src/ext.rs
bhesh: If we change x509_cert::ext::AsExtension to accepts subject as Option<&Name>, it'd be easier. Otherwise, I think I'd have to build an empty Name just to ignore it later.

baloo: or a `Name::default()` :D

PR #1258, opened 2023-11-16
RFC 5544 Implementation

enri1196:
I've included all the structures for the RFC, there are still some problems with mixed encodings. 

For reference: https://www.rfc-editor.org/rfc/rfc5544


leotaku:
Thanks for pinging me! Could you maybe describe to me/give a code example of the "problems with mixed encoding"? Without the full context, it is quite hard for me to figure out the issue.

Also, unrelated, but I assume that if this format gets accepted it should be a new crate rather than part of `cms`.

baloo:
I don't think I've seen it used expect for CMS. I think it's fine there.

cms/src/timestamped_data.rs
```diff
@@ -0,0 +1,199 @@
...
+
+/// ```text
+/// TimeStampAndCRL ::= SEQUENCE {
+///     timeStamp   TimeStampToken,          -- according to RFC 3161
+///     crl         CertificateList OPTIONAL -- according to RFC 5280
+/// }
+/// ```
+#[derive(Clone, Debug, Eq, PartialEq, Sequence)]
+#[allow(missing_docs)]
+pub struct TimeStampAndCRL {
```
baloo, line 102:
```suggestion
pub struct TimeStampAndCrl {
```

https://rust-lang.github.io/api-guidelines/naming.html#casing-conforms-to-rfc-430-c-case

cms/src/timestamped_data.rs
```diff
@@ -0,0 +1,199 @@
...
+use x509_cert::{
+    attr::{Attribute, Attributes},
+    crl::CertificateList,
+};
+
+#[derive(Clone, Copy, Debug, Eq, PartialEq, PartialOrd, Ord, Enumerated)]
+#[asn1(type = "INTEGER")]
+#[repr(u8)]
+#[allow(missing_docs)]
+pub enum TsdVersion {
```
baloo, line 21:
```suggestion
/// The `TimeStampedData` version is defined in [RFC 5544 Section 2].
pub enum Version {
```

cms/src/timestamped_data.rs
```diff
@@ -0,0 +1,199 @@
...
+    V1 = 1,
```
baloo, line 22:
```suggestion
#[derive(Clone, Copy, Debug, Default, Eq, PartialEq, PartialOrd, Ord, Enumerated)]
#[asn1(type = "INTEGER")]
#[repr(u8)]
/// The `TimeStampedData` version is defined in [RFC 5544 Section 2].
pub enum Version {
   #[default]
    V1 = 1,
```

baloo:
@enri1196 you wouldn't happen to be working on PE signatures by any chance? (if so, nixos folks are bringing up their own tooling to do it and this should be one of the pieces we were missing)

baloo (approved):

enri1196:
@baloo No, I did not know about the project from nixos folks. I’m happy to know that this PR can help them, can you show me something about it?

@leotaku  I’m still investigating this problem ‘cause I’m still not sure if I’m the one doing things wrong or there’s a problem with the way some of the documents at my current job are generated. Let me know if for your use case everything is alright. Thanks in advance.

baloo:
Most of it happens on matrix: https://matrix.to/#/#secure-boot:nixos.org
but otherwise:
  - https://github.com/m4b/goblin - the library used for parsing and writing of PE
  - https://github.com/RaitoBezarius/goblin-signing - where most of the logic for signing is intended to go.

baloo:
This looks good to me, but I'd like to run this against goblin before merging that.

enri1196:
any news?

baloo:
I didn't have time to integrate that with a builder quite yet. But I don't see much of an issue with merging this.
I will want to revisit at a later date to add the builder side though.

baloo (approved):

PR #1259, opened 2023-11-17
x509-ocsp: implement builder

bhesh:
Fixed a bug with `CrlReferences`/`CrlId` extensions, added a ton of tests, and implemented a basic builder. Optional `rand`, `builder`, and `std` features.


bhesh:
To keep all changes in `x509-ocsp`, I implemented a `GeneralizedTime` wrapper called `OcspGeneralizedTime`. But I think a `UtcTime` -> `GeneralizedTime` conversion makes sense in `der`. And possibly a `x509_cert::time::Time` -> `GeneralizedTime`. OCSP just doesn't support the legacy `UtcTime` and the conversion code was repeated too often in the builders.

bhesh:
I merged the latest master with the workflow changes but the issues still persist. I'm not sure what the problem is.

x509-ocsp/src/ext.rs
```diff
@@ -138,9 +145,8 @@ impl AssociatedOid for AcceptableResponses {
 /// ```
 ///
 /// [RFC 6960 Section 4.4.4]: https://datatracker.ietf.org/doc/html/rfc6960#section-4.4.4
-pub struct ArchiveCutoff(GeneralizedTime);
+pub type ArchiveCutoff = OcspGeneralizedTime;
```
baloo, line 148: Since this is a type alias, doesn't in mean the `OcspGeneralizedTime` now has a `AssociatedOid`?
```suggestion
pub struct ArchiveCutoff(pub OcspGeneralizedTime);
```

bhesh: That makes sense. We don't really want the ArchiveCutoff OID to be associated to `OcspGeneralizedTime` itself.

baloo (approved):

tarcieri:
@bhesh I'd like to get this merged, but it looks like it needs a rebase

bhesh:
Rebasing was causing me a lot of issues. I just merged with the latest master

tarcieri:
Thank you!

PR #1286, opened 2023-12-21
fix(#1210): fix impl of SubjectDirectoryAttributes

Taowyoo:
This PR fixes https://github.com/RustCrypto/formats/issues/1210

Though this does appear to be a bug, it's also a breaking change to incorporate the fix.

It can be merged as part of the next development cycle when we're making breaking changes.


tarcieri:
Thank you! We can hopefully merge this soon.

Issue #1293, opened 2024-01-04
const-oid: an unfortunate tradeoff in `PartialEq`/`Eq` definitions


tarcieri:
#1212 switched from derived `PartialEq`/`Eq` impls on `ObjectIdentifier` to handwritten ones which are generic and allow two `ObjectIdentifier`s with different backing storage to be compared, which makes it possible to used slice-backed storage for the database.

However, in trying to upgrade the other crates in this repo I ran into the following limitation: only constants with a derived `PartialEq`/`Eq` implementation can be used in `match` statements:

```
error: to use a constant of type `spki::ObjectIdentifier<Buffer<39>>` in a pattern, `spki::ObjectIdentifier<Buffer<39>>` must be annotated with `#[derive(PartialEq, Eq)]`
   --> pkcs5/src/pbes1.rs:192:13
    |
192 |             PBE_WITH_MD2_AND_DES_CBC_OID => Ok(Self::PbeWithMd2AndDesCbc),
    |             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    |
    = note: the traits must be derived, manual `impl`s are not sufficient
    = note: see https://doc.rust-lang.org/stable/std/marker/trait.StructuralEq.html for details
```

To go this route, all of these match statements would need to be rewritten with `if` statements, which is doable but slightly less than ideal.

Perhaps instead of allowing the storage to be fully generic, it could just be const generic around the size of the backing buffer, which would still allow OIDs in `match` statements so long as their backing buffer size were the same.

However, the database requires OIDs to all be the same type, so that would preclude the optimization that the slice-backed storage was going for: making sure each OID only uses as much memory as is required, rather than having additional unused overhead.

tarcieri:
For now I plan to revert #1212 which will get us the derived `PartialEq`/`Eq` back.

Going forward, I think it may make sense to make `ObjectIdentifierRef` its own type, and make `ObjectIdentifier` simply const generic (with a default) around its buffer size, rather than the current approach of `ObjectIdentifierRef` being a type alias for an `ObjectIdentifier` which is generic around backing storage.

That would allow a derived `PartialEq`/`Eq` impl, but also overlapping impls to compare `ObjectIdentifier` and `ObjectIdentifierRef` (and vice versa).

tarcieri:
#1299 backs out the usages of `ObjectIdentifierRef` in the database until such a time as the concerns in this issue have been addressed

tarcieri:
[...]

PR #1296, opened 2024-01-04
Bump versions to prereleases for next release cycle

tarcieri:
This bumps the following crates to prerelease versions (not intended for release, just to signal we're making breaking changes):

- `der`(_derive) v0.8.0-pre
- `pem-rfc7468` v1.0.0-pre
- `pkcs1` v0.8.0-pre
- `pkcs5` v0.8.0-pre
- `pkcs8` v0.11.0-pre
- `sec1` v0.8.0-pre
- `spki` v0.8.0-pre

This notably doesn't bump `x509-cert` or any crates which are dependent upon it due to a circular dependency problem: the certificate builder tests rely on updated public key crypto crates like `rsa`, so we'll need to do this update in two phases: this initial set of format-only crates, and then updating the cryptographic dependencies.

`pem-rfc7468` has been bumped to v1.0.0-pre based on the observation that there have been no significant changes since the last release, and that the baseline set of functionality it provides is stable.

### DIFF
--- a/der/Cargo.toml
+++ b/der/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "der"
-version = "0.7.8"
+version = "0.8.0-pre"
 description = """
 Pure Rust embedded-friendly implementation of the Distinguished Encoding Rules
 (DER) for Abstract Syntax Notation One (ASN.1) as described in ITU X.690 with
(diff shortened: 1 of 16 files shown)


der/Cargo.toml
```diff
@@ -19,9 +19,9 @@ rust-version = "1.65"
 arbitrary = { version = "1.3", features = ["derive"], optional = true }
 bytes = { version = "1", optional = true, default-features = false }
 const-oid = { version = "0.9.2", optional = true }
```
baloo, line 21: Could that be an opportunity to bump this dependency as well?
(I believe you published const-oid yesterday?)

tarcieri: Saving it for a followup due to #1293.

I'm going to make some API changes to it first before upgrading it.

tarcieri: Also upgrading needs prereleases of a bunch of the hash and cipher crates, due to their use of OIDs.

spki/Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "spki"
-version = "0.7.3"
+version = "0.8.0-pre"
```
baloo, line 3: If you don't mind I'd love to include https://github.com/RustCrypto/formats/pull/1290 in there as well.

tarcieri: You can just rebase/merge that after I land this, which is only bumping versions.

tarcieri: Not yet (though I guess I usually note there's not a release associated with these PRs and forgot in this case).

This is just bumping versions to signal that breaking changes are coming.

PR #1304, opened 2024-01-05
const-oid: fix `Encoder` size check and off-by-one error

tarcieri:
The `Encoder` previously returned `Error::Length` even when there was one remaining byte in the buffer.

This fixes that error and adds an edge case test.


baloo:
Missing the test?

tarcieri:
The test is the existing one in `encoder.rs`, however it passed before because it was using the wrong length

PR #1305, opened 2024-01-05
const-oid: add `ObjectIdentifierRef`

tarcieri:
Adds a `repr(transparent)` newtype for a `[u8]` which is guaranteed to contain a valid BER serialization of an OID. This is a similar approach to how `Path`/`PathBuf` or `OsStr`/`OsString` work (except with `ObjectIdentifier` being stack-allocated instead of heap allocated).

An unsafe pointer cast is required to go from `&[u8]` to `&ObjectIdentifierRef`, so unfortunately this means the crate is no longer `forbid(unsafe_code)`, however it's been lowered to `deny(unsafe_code)` to ensure contributors think twice before adding more.

`Borrow` and `Deref` impls have been added to the owned `ObjectIdentifier` type, allowing common functionality to be moved to `ObjectIdentifierRef`, allowing both types to exist while eliminating code duplication.

A `PartialEq` impl allows them to be compared.

The `db` module continues to use `ObjectIdentifier` for now, however hopefully this approach would allow #1212 to be reinstated and for `ObjectIdentifierRef`s to be used for the database eventually (i.e. revert the revert in #1299)

NOTE: this PR also relaxes the previous requirement that an OID have at least three arcs. It is now allowed to only have two. It also removes the `Error::NotEnoughArcs` variant that covered that particular case.

### DIFF
--- a/const-oid/src/arcs.rs
+++ b/const-oid/src/arcs.rs
@@ -58,7 +58,8 @@ impl<'a> Arcs<'a> {
         match self.cursor {
             // Indicates we're on the root arc
             None => {
-                let root = RootArcs::try_from(self.bytes[0])?;
+                let root_byte = *self.bytes.first().ok_or(Error::Empty)?;
+                let root = RootArcs::try_from(root_byte)?;
                 self.cursor = Some(0);
                 Ok(Some(root.first_arc()))
             }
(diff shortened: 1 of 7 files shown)


tarcieri:
> An unsafe pointer cast is required to go from &[u8] to &ObjectIdentifierRef, so unfortunately this means the crate is no longer forbid(unsafe_code)

Sidebar: really wish these could eventually be "safe transmutes"

PR #1330, opened 2024-01-22
tls_codec: Introduce helper macro for conditional deserialization

kkohbrok:
This PR introduces the `#[tls_codec(cd_field)]` macro. It can be used alongside the `#[conditionally_deserializable]` macro to mark fields that are also conditionally deserializable and that internally need to have the const generic added.

### DIFF
--- a/tls_codec/derive/tests/decode.rs
+++ b/tls_codec/derive/tests/decode.rs
@@ -530,7 +530,7 @@ fn type_with_unknowns() {
 
 #[cfg(feature = "conditional_deserialization")]
 mod conditional_deserialization {
-    use tls_codec::{Deserialize, Serialize};
+    use tls_codec::{Deserialize, DeserializeBytes, Serialize, Size};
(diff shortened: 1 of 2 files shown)


tls_codec/derive/tests/decode.rs
```diff
@@ -549,6 +549,13 @@ mod conditional_deserialization {
         assert_eq!(deserializable_struct.a, undeserializable_struct.a);
         assert_eq!(deserializable_struct.b, undeserializable_struct.b);
 
+        #[conditionally_deserializable]
+        #[derive(TlsSize, TlsSerialize, PartialEq, Debug)]
+        struct NestedExampleStruct {
```
franziskuskiefer, line 554: This seems to fail when adding a generic member. Is this supposed to work?
It's ok when not, but should be documented then and throw a useful error.

franziskuskiefer (approved):
Thanks, lgtm

PR #1334, opened 2024-01-25
tls_codec: Add missing docs

kkohbrok:
Just noticed that I forgot to add documentation for the `DeserializeBytes` trait.


tls_codec/src/lib.rs
```diff
@@ -198,6 +198,9 @@ pub trait Deserialize: Size {
     }
 }
 
+/// The `DeserializeBytes` trait defines functions to deserialize a byte slice
+/// to a struct or enum. In contrast to [`Deserialize`], this trait operates
+/// directly on byte slices and can return any remeining bytes.
```
franziskuskiefer, line 203:
```suggestion
/// directly on byte slices and can return any remaining bytes.
```

franziskuskiefer (approved):
Thanks

PR #1352, opened 2024-02-15
fix CertTemplate (should have used OptionalValidity, not Validity)

carl-wallace:


tarcieri:
Looks like you have an unused import for `Validity` now

carl-wallace:
Removed the Validity import

PR #1373, opened 2024-03-31
Fix `repository` field in Cargo.toml

link2xt:
According to https://doc.rust-lang.org/cargo/reference/manifest.html#the-repository-field
`repository` is a source repository URL.
GitHub URLs with `tree` are not repository URLs
and are not usable as `git clone` arguments.

Path within the worktree
is encoded in `.cargo_vcs_info.json` of the crate,
so encoding it in `Cargo.toml` is not necessary.

I am automating downloading of crate sources
and invalid URLs require workarounds
to turn them into repository URLs that can be cloned.


tarcieri:
> Path within the worktree is encoded in .cargo_vcs_info.json of the crate, so encoding it in Cargo.toml is not necessary.

...but that's not displayed on crates.io.

It would be nice to retain the direct links to the actual subdirectory containing the package somewhere that generates a clickable link on crates.io. Perhaps `homepage`?

der/derive/Cargo.toml
```diff
@@ -6,6 +6,7 @@ authors = ["RustCrypto Developers"]
 license = "Apache-2.0 OR MIT"
 documentation = "https://docs.rs/der"
 repository = "https://github.com/RustCrypto/formats/tree/master/der/derive"
```
tarcieri, line 8:
```suggestion
homepage = "https://github.com/RustCrypto/formats/tree/master/der/derive"
```

tarcieri (approved):

PR #1374, opened 2024-04-01
Make sure `der` crate is compatible with potential language breaking changes

WaffleLapkin:
`?` currently influences inference s.t. for `error()?` rustc infers `T = ()`. However, it is quite confusing -- `?` is not supposed to influence inference, it's just a conicidence because of the curent implementation. There is an idea to change this behavior in future Rust versions, such that this code won't compile without this change.

See https://github.com/rust-lang/rust/pull/122412 for more information about the change.
(`der` is used by cargo, so this partially blocks the PR, please ping me once you release a new `der` version so I can update it inside cargo)

### DIFF
--- a/der/src/writer/slice.rs
+++ b/der/src/writer/slice.rs
@@ -31,7 +31,7 @@ impl<'a> SliceWriter<'a> {
     /// Encode a value which impls the [`Encode`] trait.
     pub fn encode<T: Encode>(&mut self, encodable: &T) -> Result<()> {
         if self.is_failed() {
-            self.error(ErrorKind::Failed)?;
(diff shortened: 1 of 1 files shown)


baloo:
@WaffleLapkin this is going to take a little while until we tag a release for der.
We could backport that to `0.7` and release a `0.7.9` if that makes it easier for you.

baloo:
cargo currently uses `0.7.8`
https://github.com/rust-lang/cargo/blob/a59aba136aab5510c16b0750a36cbd9916f91796/Cargo.lock#L799C15-L799C16

PR #1378, opened 2024-04-01
spki: fix `OidUnknown` reporting

tarcieri:
As noted in RustCrypto/RSA#422, we're not returning the actual "unknown OID" encountered in these cases, but the expected one, which makes it hard to debug mismatches.

### DIFF
--- a/spki/src/algorithm.rs
+++ b/spki/src/algorithm.rs
@@ -106,7 +106,7 @@ impl<Params> AlgorithmIdentifier<Params> {
         if self.oid == expected_oid {
             Ok(expected_oid)
         } else {
-            Err(Error::OidUnknown { oid: expected_oid })
+            Err(Error::OidUnknown { oid: self.oid })
         }
     }
 }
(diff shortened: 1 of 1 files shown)


spki/src/algorithm.rs
```diff
@@ -122,7 +122,7 @@ impl<'a> AlgorithmIdentifierRef<'a> {
         if actual_oid == expected_oid {
             Ok(actual_oid)
         } else {
-            Err(Error::OidUnknown { oid: expected_oid })
+            Err(Error::OidUnknown { oid: actual_oid })
```
tarcieri, line 125: These could probably use tests.

Perhaps the it should also carry the expected OID if available.

baloo (approved):

spki/src/algorithm.rs
baloo: adding the expected should be cheap.

tarcieri:
Can make an issue for followup work

PR #1387, opened 2024-04-22
Don't fail with InvalidLength when reading nothing at end of data

mpalmer:
This is a curious corner case that can happen when, for example, you're reading the length-value encoded strings in OpenSSH. If the last string in the data is of zero-length, then decode() will be called with an empty buffer, so we don't need to read any bytes. However, since `is_finished()` was checked unconditionally, the zero-length read (which would be OK) was instead rejected because there was no data to (not) read.

The bug was particularly mischievous because `is_finished()` will return false if there's `=` padding, so 2/3 of empty read calls will work normally.

### DIFF
--- a/base64ct/src/decoder.rs
+++ b/base64ct/src/decoder.rs
@@ -103,8 +103,13 @@ impl<'i, E: Encoding> Decoder<'i, E> {
     ///
     /// # Returns
     /// - `Ok(bytes)` if the expected amount of data was read
-    /// - `Err(Error::InvalidLength)` if the exact amount of data couldn't be read
+    /// - `Err(Error::InvalidLength)` if the exact amount of data couldn't be read, or
+    ///   if the output buffer has a length of 0
     pub fn decode<'o>(&mut self, out: &'o mut [u8]) -> Result<&'o [u8], Error> {
+        if out.is_empty() {
+            return Err(InvalidLength);
+        }
+
         if self.is_finished() {
             return Err(InvalidLength);
         }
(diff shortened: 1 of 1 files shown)


tarcieri:
Is this really necessary and the best solution? Wouldn’t it be better not to perform a zero length read in the first place?

mpalmer:
> Is this really necessary and the best solution? Wouldn’t it be better not to perform a zero length read in the first place?

If "don't perform zero length reads" is the best solution, that's cool, although in that case it would be better, IMO, if this method *always* rejected zero length reads.  The thing that made this bug so hard to track down was the fact that the zero length read only failed in a very narrow set of circumstances.

base64ct/src/decoder.rs
```diff
@@ -105,6 +105,11 @@ impl<'i, E: Encoding> Decoder<'i, E> {
     /// - `Ok(bytes)` if the expected amount of data was read
     /// - `Err(Error::InvalidLength)` if the exact amount of data couldn't be read
     pub fn decode<'o>(&mut self, out: &'o mut [u8]) -> Result<&'o [u8], Error> {
+        if out.is_empty() {
+            // We did not intend to read anything from the reader.
+            return Ok(out)
```
mpalmer, line 110:
```suggestion
            return Ok(out);
```

tarcieri:
I think it's fine to always reject zero length reads

mpalmer:
Righto, I'll rework the PR to go that way.

mpalmer:
OK, base64ct now errors on zero-length decode requests, with `InvalidLength`.

tarcieri (approved):

tarcieri:
Ended up reverting this as it's a breaking change: #1714

PR #1394, opened 2024-04-30
Add STB (Belarus standarts) to const-oid

makavity:
Added OIDS from Republic of Belarus standarts (STB)
https://github.com/bcrypto/bake/blob/master/spec/bake-module-v1.asn (for example, one file)


tarcieri:
Can you leave the `pkcs8` changes out for now? It would probably be good to just put those in the `bign256` crate.

tarcieri (approved):

Issue #1228, opened 2023-09-22
der: Huge generic bloat due to NestedReader


dishmaker:
Hi!

My .exe file using `der` crate for a huge protocol is too large (12 MB for a simple DER encoder/decoder)
I've just used `cargo bloat` and `cargo llvm-lines` on my protocol crate and was surprised that `der` crate makes a lot of generic functions.

`read_nested` is actually the main cause!
```
Lines                  Copies               Function name
-----                  ------               -------------
34633 (2.6%,  2.6%)    187 (1.0%,  1.0%)    der::reader::Reader::read_nested
```

Because it is nesting itself recursively!
```
NestedReader<NestedReader<NestedReader<SliceReader>>>
```

You can easily check by adding field `test: FailAsn1` inside your `#[derive(Sequence)]`
```rust
pub struct FailAsn1 {}

impl<'a> DecodeValue<'a> for FailAsn1 {
    fn decode_value<R: Reader<'a>>(
        reader: &mut R,
        header: Header,
    ) -> Result<Self> {
        // prints: reader type name: NestedReader<NestedReader<NestedReader<NestedReader<NestedReader<SliceReader>>>>>
        panic!("reader type name: {}", tynm::type_name::<R>())
    }
}
```

Not only that Reader gets generic-bloated but also all `Decode` and `DecodeValue` traits.

Really! Just look above, `DecodeValue` is generic over `R: Reader` so all objects implementing `Decode` also get bloated.

But the main cause is that:
- `NestedReader` takes `Self` that can be another `NestedReader`:

https://github.com/RustCrypto/formats/blob/086ecd7d27aa3f6c5c97dc2a33adc7159008086b/der/src/reader.rs#L127-L131

tarcieri:
I'll also note that on Linux/Mac, the binaries produced by the test cases in this repo are around 1MB-1.5MB.

I have no idea what's bloating your particular binary to 12MB.

dishmaker:
> This is an extremely immaturely phrased issue.

Ok, I've rephrased the issue.
However I think you didn't understand the simple problem:
`NestedReader<NestedReader<NestedReader<SliceReader>>>`

should become at most:
`NestedReader<SliceReader>`
and never nest itself deeper. I know it might be hard to solve without `dyn Reader`.


Here is the protocol I'm implementing:
https://github.com/monai/node-passport/blob/master/lib/pkcs15/cryptographic_information_framework.asn1


[...]

PR #1416, opened 2024-05-23
support private tags and tag numbers >30 that are stored in long form

### LINKED ISSUE #1381: der: Encoding higher tag number
Hello,

I have an use case where I need to DER-encode ASN.1 message using greater than 30 for tag number.
I noticed in some comments that "This library deliberately does not support this: tag numbers greater than 30 are disallowed.".

I realize that `TagNumber` are an enumeration in this library, which makes arbitrarily high tag number problematic.
I am curious if you would consider lifting this limitation and willing to accept patches.

I can imagine 2 approaches:
    + Increase max tag number to 127, and support 2-bytes encoding of tags. Since 127 is still reasonable number it sounds like it would acceptable to have tag number up to `TagNumber::N127`.
    + Accept large number by using `u64` or `u128`. This might be a too big impact the library API.

Thank you,

kamulos:
closes #1381 

This PR implements two things:

- Allow the use of tag numbers >30. In the der representation they are stored in the long form as multiple bytes.
- Allow `PRIVATE` tags be annotated like context specific ones in the derive macros

## Current State

Currently this is a draft to get feedback on the chosen approach. This code is in use in my project and works correctly for it's use-case: a lot of private tags in a deep hierarchy of choices and sequences.

For the most part this code should be in a state to be reviewed. There are a few parts that are still lacking:

- The code duplication of the `private` module, that is basically just a copy of the `context_sensitive` module
- The tag parsing and serializing code is too complicated
- The tests have not been adjusted / written

## Overview

`TagNumber` is now represented by an `u16`. This should be enough for any use-case imaginable, but also small enough to be embedded-friendly. It's content is now `pub` instead of `pub(super)`. This was necessary because `der_derive` wants to match against constant `TagNumber`s. In general having this public should not be an issue, because all `2^16` tag numbers are now valid and it is not possible to create an invalid state by having access to the `u16`.

Also the `Private` and `PrivateRef` types exist now. In the derive macros `private` can be annotated in the same way as `context_specific`. Internally this is represented as `class: Option<Class>` instead of `context_sensitive: Option<TagNumber>`.

Breaking changes that I found no way to avoid are:

- `TagNumber::new()` and `TagNumber::value()` use `u16` now
- Conversions from and to `u8` for `Tag` and `TagNumber` do not make sense anymore, this also applies to `Tag::octet()`
- The `Reader` trait now needs a function to peek multiple bytes in advance to be able to peek a tag (`Reader::peek_offset()`)
- `Length::for_tlv()` just assumed, that a tag is always one byte, it now somehow needs to know about the specific tag we are looking at


der/src/reader.rs
```diff
@@ -23,8 +23,10 @@ pub trait Reader<'r>: Sized {
     /// Get the length of the input.
     fn input_len(&self) -> Length;
 
-    /// Peek at the next byte of input without modifying the cursor.
-    fn peek_byte(&self) -> Option<u8>;
+    /// Peek at the byte of the input at position `offset` relative to the current position.
+    ///
+    /// Does not modify the decoder's state.
+    fn peek_offset(&self, offset: usize) -> Option<u8>;
```
tarcieri, line 29: Changes to the reader trait like this will delay acceptance of this PR until we have a solution to #1228, FYI

kamulos: Tried to look around for alternatives. In reality I only need the `peek_tag` function now, peeking a byte at a specific offset is not necessarily needed.

Initially I thought, I could just do the same thing `SliceReader::peek_header()` is doing for the function `Reader::peek_tag()` (meaning clone and mutating read), but `peek_tag` is a provided method of `Reader`, which is not `Clone`.

That leaves two options as far as I can tell:

1. I could misuse `Reader::peek_header()` as a building block of `Reader::peek_tag()`: peeking the header and returning the tag. This is absolutely the wrong way around, but could allow the functionality without a change to the trait. [^1]
2. If it takes a while before this is merged, we will use our forked version until then. That is no issue at all

What do you think?

[^1]: I still think having the `Option` in `peek_tag(&self) -> Result<Option<Tag>, Error>` is pretty useful and could benefit the trait.

tarcieri: #1418 contains part of the changes I want to make to the `Reader` trait. It should be a simplification for your purposes.

Regarding this method, perhaps it could be named `peek_byte_at` or thereabouts for consistency with the other methods which are named after the type they're peeking at?

tarcieri: The methods return different types:

- `peek_tag` returns a `Tag`
- `peek_header` returns a `Header`
- `peek_byte` returns `u8`

While your proposal allows more of these to be provided methods (something that can also be implemented in terms of `peek_offset`, and would be generally nice to have), it doesn't actually provide equivalent functionality.

dishmaker:
This crate is still unusable for the European Tachograph regulation.

There are many 2-byte tags.

https://eur-lex.europa.eu/legal-content/EN/TXT/?uri=celex%3A02016R0799-20200226

| Field | Field ID | Tag | Length (bytes) | ASN.1 data type |
|--|--|--|--|--|
| ECC Certificate | C | `7F 21` | var | custom SEQUENCE |
| ECC Certificate Body | B | `7F 4E` | var | custom SEQUENCE  |
| Certificate Profile Identifier | CPI | `5F 29` | 1 | `INTEGER(0..255)` |
| Certificate Authority Reference | CAR | `42` | 8 | `KeyIdentifier` |
| Certificate Holder Authorisation | CHA | `5F 4C`| 7 | `CertificateHolder Authorisation` |
| Public Key | PK | `7F 49` | var | custom SEQUENCE |
| Domain Parameters | DP | `06` | var | `OBJECT IDENTIFIER` |
| Public Point | PP | `86` | var | `OCTET STRING` |
| Certificate Holder Reference | CHR | `5F 20` | 8 | `KeyIdentifier` |
| Certificate Effective Date | CEfD | `5F 25` | 4 | `TimeReal` |
| Certificate Expiration Date | CExD | `5F 24` | 4 | `TimeReal` |
| ECC Certificate Signature | S | `5F 37` | var | `OCTET STRING` |

baloo:
@dishmaker we're still open for breaking changes in this release cycle. There is a legitimate use-case, PRs are welcome!

kamulos:
Sorry for the slow reaction. In a week I am back and try to get in some more progress here.

tarcieri:
@kamulos I've finished making the other changes I consider blockers for this one, though it's created quite a few conflicts.

If you can get this PR green though, I can make it a priority for the next to merge. I don't expect there will be many other major changes to `der` itself.

kamulos:
@tarcieri Thanks for your work :heart: And sorry for being very slow, the workload at the job was high and I also had vacations.

Currently we are actually releasing this code into production and things will calm down from here. So I will get back to this next week and hopefully get some version ready that is good quality.

tarcieri:
@kamulos no worries, we'd like to do a release "soon", but that's probably in a month or two so you have some time

kamulos:
@tarcieri I rebased and polished the implementation. The unit tests for the `Private` type are still broken, because I did not adjust them and I should definitely write some more unit Tests for the encoding and decoding of the `Tag` type.

But other than that I think it is ready for review. :+1: 

There are some noteworthy points:

1. The `decode_with` function in `Private` and `ContextSpecific` requires peeking for a `Tag` but maybe getting `None` as return. Therefore I added another function `Tag::peek_optional` that is only crate internal. This feels a little bit wring, but I did not want to modify the signature of public functions or traits.
2. `Private` is basically a copy of `ContextSpecific` with one difference: In `decode_with` I removed the skipping of lower tag numbers. This did not work for my use-case and I did not understand it.

dishmaker:
> `Private` is basically a copy of `ContextSpecific`

Well, I also need `Application` 🥺 

I think the best solution is using generics:
```rust
pub type Application<const TAG: u16, T> = CustomClass<T, TAG, CLASS_APPLICATION>;
pub type ContextSpecific<const TAG: u16, T> = CustomClass<T, TAG, CLASS_CONTEXT_SPECIFIC>;
pub type Private<const TAG: u16, T> = CustomClass<T, TAG, CLASS_PRIVATE>;

pub type ApplicationImplicit<const TAG: u16, T> = CustomClassImplicit<T, TAG, CLASS_APPLICATION>;
pub type ContextSpecificImplicit<const TAG: u16, T> = CustomClassImplicit<T, TAG, CLASS_CONTEXT_SPECIFIC>;
pub type PrivateImplicit<const TAG: u16, T> = CustomClassImplicit<T, TAG, CLASS_PRIVATE>;

/// EXPLICIT
#[derive(Copy, Clone, Debug, Eq, PartialEq, PartialOrd, Ord)]
pub struct CustomClass<T: for<'a> Decode<'a>, const TAG: u16, const CLASS: u8> {
    pub value: T,
}

/// IMPLICIT
#[derive(Copy, Clone, Debug, Eq, PartialEq, PartialOrd, Ord)]
pub struct CustomClassImplicit<T: for<'a> DecodeValue<'a>, const TAG: u16, const CLASS: u8> {
    pub value: T,
}

pub const CLASS_APPLICATION: u8 = 0b01;
pub const CLASS_CONTEXT_SPECIFIC: u8 = 0b10;
pub const CLASS_PRIVATE: u8 = 0b11;
```

dishmaker:
I've fixed all dependencies and added these 3 custom tags as generic ones :)

tarcieri:
@kamulos if you are still interested in this, I would prefer a much more minimal PR which only changes the handling of the `Tag` type, and doesn't try to add higher-level APIs beyond that, which are really what's blocking getting this merged

kamulos:
@tarcieri thanks for the message. I looked into it, rebased my commit and removed the stuff in a new commit.

With my last commit I added a three unit test cases, where the decoder is pretty forgiving at the moment. The second test case ("universal tag in long form") might be ok, but the other two accept invalid `der` encoding. Should they be rejected? I saw in the API that there is also a `.from_ber()` function now, and these encodings might be acceptable in the `ber` encoding. (not sure about that)

tarcieri:
@kamulos can you perhaps open a PR to #1651 to add your test cases?

turbocool3r:
> @kamulos can you perhaps open a PR to #1651 to add your test cases?

I'll be unavailable for about 4 days, but I'll be happy to accept a PR or include the necessary changes myself. Also I should add @kamulos to PR authors somehow, is there any convention for this? Are you ok with this @kamulos?

tarcieri:
Ended up going with #1651 instead

PR #1424, opened 2024-05-29
base32ct: add const encoded_len fn

dodomorandi:
This can be handy to avoid an heap allocation, because it is possible to use the result of the const fn to allocate an array of the right size.

EDIT: I changed a bit the initial proposal in order to make it actually easier to use in const context. Previously `encoded_len` took a slice (like in the trait), now it takes an `usize` representing the number of bytes to encode.

### DIFF
--- a/base32ct/src/encoding.rs
+++ b/base32ct/src/encoding.rs
@@ -220,14 +220,9 @@ impl<T: Alphabet> Encoding for T {
         }
     }
 
+    #[inline]
     fn encoded_len(bytes: &[u8]) -> usize {
-        if bytes.is_empty() {
-            0
-        } else if Self::PADDED {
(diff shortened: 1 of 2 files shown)


base32ct/src/encoding.rs
```diff
@@ -259,6 +254,17 @@ fn remove_padding(mut input: &[u8]) -> Result<&[u8]> {
     Ok(input)
 }
 
+/// Get the length of Base32 produced by encoding the given amount of bytes.
+pub const fn encoded_len<T: Encoding>(bytes: usize) -> usize {
```
baloo, line 258: nit:

```suggestion
pub const fn encoded_len<T: Encoding>(length: usize) -> usize {
```

baloo (approved):

tarcieri (approved):

PR #1425, opened 2024-06-03
Correctly format email addresses.

cobward:
I'm also wondering if it's worth documenting that the `Name::from_str` functionality is not guaranteed to generate a valid X.501 Name as per RFC 5280, and that manual construction of a Name from OIDs and ASN.1 is the preferred method?

### DIFF
--- a/x509-cert/src/attr.rs
+++ b/x509-cert/src/attr.rs
@@ -2,7 +2,7 @@
 
 use alloc::vec::Vec;
 use const_oid::db::{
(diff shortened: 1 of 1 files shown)


baloo (approved):
Thanks!

baloo:
I think I would still recommend using `Name::from_str` over manual construction. And I would consider any deviation from the spec a bug (that we're happy to fix).
If there was any conflicts between two specs, then we would document that.

PR #1433, opened 2024-06-12
pkcs5: add support for using AES-GCM in PBES2

### LINKED ISSUE #1432: Support for GCM in PKCS8
The best widely implemented algorithm for PBES2/PKCS8 is AES-256-CBC. However RFC 8018 is open supporting any cipher in an application defined way

> encryptionScheme identifies the underlying encryption scheme. It shall be an algorithm ID with an OID in the set PBES2-Encs, *whose definition is left to the application*.

[RFC 8018 section A.4](https://datatracker.ietf.org/doc/html/rfc8018#appendix-A.4)

Some implementations of PBES2/PKCS8 support GCM as an alternative; I'm aware of https://github.com/randombit/botan and https://github.com/Legrandin/pycryptodome (these implementations are compatible).

Would there be any interest in extending `pkcs8` to also support AES-256-GCM? It extends pretty naturally from the AES-CBC ASN.1 definitions:

1) Use the appropriate OID for AES-GCM
2) Put the GCM nonce where CBC places the IV
3) Profit with AEAD protected keys

randombit:
Fixes #1432

Some notes/comments

* I implemented only AES-128 and AES-256. pycryptodome also supports using AES-192 with GCM. Botan supports any cipher that has a GCM OID defined (for instance SM4). I imagine most users really just want AES-256.
* For pbes2::Parameters I always associated GCM with Scrypt, partially to avoid a combinatoric explosion of constructors and partially because there doesn't seem to be any reason to use PBKDF2 given the existence of Scrypt. Happy to add PBKDF2 if you prefer otherwise.
* The GCM code is quite repetitive and awkward but I could not figure out any good ways of cleaning it up, definitely would appreciate any suggestions here.
* The test data was generated using Botan 2.19.3. I also confirmed latest pycryptodome can import `ed25519-encpriv-aes256-gcm-scrypt.der`
```
>>> ECC.import_key(binascii.unhexlify(k), b'hunter42')
EccKey(curve='Ed25519', point_x=2621690352....
```


pkcs5/src/pbes2.rs
```diff
@@ -321,6 +369,18 @@ pub enum EncryptionScheme {
...
+    /// AES-128 in CBC mode
+    Aes128Gcm {
+        /// GCM nonce
+        iv: [u8; GCM_NONCE_SIZE],
+    },
+
+    /// AES-256 in GCM mode
+    Aes256Gcm {
+        /// GCM nonce
+        iv: [u8; GCM_NONCE_SIZE],
```
tarcieri, line 381: Any reason to call it `iv` instead of just `nonce`?

pkcs5/src/pbes2/encryption.rs
```diff
@@ -48,6 +49,104 @@ fn cbc_decrypt<'a, C: BlockCipherDecrypt + BlockCipher + KeyInit>(
         .map_err(|_| Error::EncryptFailed)
 }
 
+fn aes_128_gcm_encrypt<'a>(
```
tarcieri, line 52: You can probably write this and `aes_*_gcm_decrypt` generically in terms of the `AesGcm` struct by parameterizing it over the `Aes` cipher used, e.g.

```rust
fn aes_gcm_encrypt<'a, Aes: BlockCipher + BlockSizeUser<BlockSize = U16> + BlockEncrypt>
```

(though I would probably recommend using a `where` clause)

If you don't want to make it generic yourself, I can probably do that as a followup.

pkcs5/src/pbes2.rs
randombit: Nope! I'll change this

pkcs5/src/pbes2/encryption.rs
randombit: OK I'll give this a try

pkcs5/src/pbes2.rs
randombit: Done

pkcs5/src/pbes2/encryption.rs
randombit: I made an attempt at this and am hitting trait errors that don't really make sense. I think it might be caused by the existence of two different versions of `cipher` crate in the dependency tree.

baloo: you might need to use a pre-release version of aes and aes-gcm.

```
[dependencies]
aes = { version = "=0.9.0-pre" }
aes-gcm = { version = "=0.11.0-pre" }
```

and in the root `Cargo.toml`:
```
[patch.crates-io]
aes-gcm = { git = "https://github.com/RustCrypto/AEADs.git" }
```

randombit: Using the pre-release crates helped. Pushed c5280f31cb2597a5944a9d2573659751e3aa7765 everything seems to work

pkcs5/src/pbes2/encryption.rs
```diff
@@ -48,6 +49,71 @@ fn cbc_decrypt<'a, C: BlockCipherDecrypt + BlockCipher + KeyInit>(
         .map_err(|_| Error::EncryptFailed)
 }
 
+fn gcm_encrypt<'a, C>(
+    es: EncryptionScheme,
+    key: EncryptionKey,
+    iv: &[u8],
```
baloo, line 55:
```suggestion
    iv: Nonce<C>,
```

from: https://docs.rs/aead/0.5.2/aead/type.Nonce.html

pkcs5/src/pbes2/encryption.rs
```diff
@@ -48,6 +49,71 @@ fn cbc_decrypt<'a, C: BlockCipherDecrypt + BlockCipher + KeyInit>(
...
+        .encrypt_in_place_detached(&nonce, &[], &mut buffer[..pos])
+        .map_err(|_| Error::EncryptFailed)?;
+    buffer[pos..].copy_from_slice(tag.as_ref());
+    Ok(&buffer[0..pos + TAG_SIZE])
+}
+
+fn gcm_decrypt<'a, C>(
+    es: EncryptionScheme,
+    key: EncryptionKey,
+    iv: &[u8],
```
baloo, line 83:
```suggestion
    iv: &Nonce<C>,
```

pkcs5/src/pbes2/encryption.rs
```diff
@@ -48,6 +49,71 @@ fn cbc_decrypt<'a, C: BlockCipherDecrypt + BlockCipher + KeyInit>(
...
+    }
+    let gcm = <aes_gcm::AesGcm<C, U12> as GcmKeyInit>::new_from_slice(key.as_slice())
+        .map_err(|_| es.to_alg_params_invalid())?;
+    let nonce: [u8; NONCE_SIZE] = iv.try_into().map_err(|_| es.to_alg_params_invalid())?;
+    let nonce = Nonce::from(nonce);
+    let tag = gcm
+        .encrypt_in_place_detached(&nonce, &[], &mut buffer[..pos])
+        .map_err(|_| Error::EncryptFailed)?;
+    buffer[pos..].copy_from_slice(tag.as_ref());
+    Ok(&buffer[0..pos + TAG_SIZE])
```
baloo, line 77:
```suggestion
fn gcm_encrypt<'a, C, NonceSize, TagSize>(
    es: EncryptionScheme,
    key: EncryptionKey,
    nonce: Nonce<C>,
    buffer: &'a mut [u8],
    pos: usize,
) -> Result<&'a [u8]>
where
    C: BlockCipher + BlockSizeUser<BlockSize = U16> + GcmKeyInit + BlockCipherEncrypt,
    aes_gcm::AesGcm<C, NonceSize, TagSize>: GcmKeyInit,
    TagSize: TagSize,
{
    const TAG_SIZE: usize = 16;
    const NONCE_SIZE: usize = 12;

    if buffer.len() < TagSize::USIZE + pos {
        return Err(Error::EncryptFailed);
    }
    let gcm = <aes_gcm::AesGcm<C, NonceSize, TagSize> as GcmKeyInit>::new_from_slice(key.as_slice())
        .map_err(|_| es.to_alg_params_invalid())?;
    let tag = gcm
        .encrypt_in_place_detached(&nonce, &[], &mut buffer[..pos])
        .map_err(|_| Error::EncryptFailed)?;
    buffer[pos..].copy_from_slice(tag.as_ref());
    Ok(&buffer[0..pos + TagSize::U8])
```

randombit: That doesn't compile directly but I made effectively this change (along with similar for GCM decrypt) in 
603bbb4 PTAL

pkcs5/src/pbes2/encryption.rs
```diff
@@ -49,61 +49,55 @@ fn cbc_decrypt<'a, C: BlockCipherDecrypt + BlockCipher + KeyInit>(
...
-    let msg_len = if buffer.len() < TAG_SIZE {
+    let msg_len = if buffer.len() < TagSize::USIZE {
         return Err(Error::DecryptFailed);
     } else {
-        buffer.len() - TAG_SIZE
+        buffer.len() - TagSize::USIZE
     };
 
-    let gcm = <aes_gcm::AesGcm<C, U12> as GcmKeyInit>::new_from_slice(key.as_slice())
+    let gcm = <aes_gcm::AesGcm<C, NonceSize> as GcmKeyInit>::new_from_slice(key.as_slice())
```
baloo, line 96:
```suggestion
    let gcm = <aes_gcm::AesGcm<C, NonceSize, TagSize> as GcmKeyInit>::new_from_slice(key.as_slice())
```

This will otherwise take the default TagSize (U16)

pkcs5/src/pbes2/encryption.rs
baloo: I'm not surprised, I didn't run this through a compiler, and made the edits in GitHub's UI.

pkcs5/src/pbes2/encryption.rs
```diff
@@ -48,6 +49,65 @@ fn cbc_decrypt<'a, C: BlockCipherDecrypt + BlockCipher + KeyInit>(
...
+
+fn gcm_decrypt<C, NonceSize, TagSize>(
+    es: EncryptionScheme,
+    key: EncryptionKey,
+    nonce: Nonce<NonceSize>,
+    buffer: &mut [u8],
+) -> Result<&[u8]>
+where
+    C: BlockCipher + BlockSizeUser<BlockSize = U16> + GcmKeyInit + BlockCipherEncrypt,
+    aes_gcm::AesGcm<C, U12>: GcmKeyInit,
```
baloo, line 86:
```suggestion
    aes_gcm::AesGcm<C, NonceSize>: GcmKeyInit,
```

pkcs5/src/pbes2/encryption.rs
```diff
@@ -48,6 +49,65 @@ fn cbc_decrypt<'a, C: BlockCipherDecrypt + BlockCipher + KeyInit>(
...
+    TagSize: aes_gcm::TagSize,
+    NonceSize: aes::cipher::ArraySize,
+{
+    let msg_len = if buffer.len() < TagSize::USIZE {
+        return Err(Error::DecryptFailed);
+    } else {
+        buffer.len() - TagSize::USIZE
+    };
```
baloo, line 94:
```suggestion
    let msg_len = buffer.len().checked_sub(TagSize::USIZE).ok_or{Error::DecryptFailed)?;
```

tarcieri (approved):

PR #1441, opened 2024-07-05
x509-cert: specify concrete types to help the compiler

istankovic:
rustc fails to infer the appropriate types with `Default::default` when building a crate depending on
```
    femme = "2.2.1"
    getrandom = { version = "0.2.15", features = ["js"] }
    rand = "0.8.5
    x509-cert = { version = "0.2.5", features = ["hazmat", "builder"] }
```

This is the primary reason we at [Wire](https://github.com/wireapp) have a fork of `formats` so it'd be nice if we could go back to tracking upstream.

### DIFF
--- a/x509-cert/src/ext.rs
+++ b/x509-cert/src/ext.rs
@@ -29,7 +29,7 @@ pub mod pkix;
 pub struct Extension {
     pub extn_id: ObjectIdentifier,
 
-    #[asn1(default = "Default::default")]
+    #[asn1(default = "bool::default")]
     pub critical: bool,
 
     pub extn_value: OctetString,
--- a/x509-cert/src/ext/pkix/constraints/basic.rs
+++ b/x509-cert/src/ext/pkix/constraints/basic.rs
@@ -14,7 +14,7 @@ use der::Sequence;
 #[derive(Clone, Debug, Eq, PartialEq, Sequence)]
 #[allow(missing_docs)]
 pub struct BasicConstraints {
-    #[asn1(default = "Default::default")]
+    #[asn1(default = "bool::default")]
     pub ca: bool,
     pub path_len_constraint: Option<u8>,
 }
--- a/x509-cert/src/ext/pkix/constraints/name.rs
+++ b/x509-cert/src/ext/pkix/constraints/name.rs
@@ -61,7 +61,7 @@ pub struct GeneralSubtree {
     #[asn1(
         context_specific = "0",
         tag_mode = "IMPLICIT",
-        default = "Default::default"
+        default = "u32::default"
     )]
     pub minimum: u32,
 
--- a/x509-cert/src/ext/pkix/crl/dp.rs
+++ b/x509-cert/src/ext/pkix/crl/dp.rs
@@ -31,14 +31,14 @@ pub struct IssuingDistributionPoint {
     #[asn1(
         context_specific = "1",
         tag_mode = "IMPLICIT",
-        default = "Default::default"
+        default = "bool::default"
     )]
     pub only_contains_user_certs: bool,
 
     #[asn1(
         context_specific = "2",
         tag_mode = "IMPLICIT",
-        default = "Default::default"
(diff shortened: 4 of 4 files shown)


baloo:
I'm not sure I understand the issue you're running into.
I've tried setting up a simple project with the dependencies you're listing, and I can't reproduce. I've also tried with MSRV (rustc 1.65), that also works.

Any chance you could provide a reproducible case?

istankovic:
> Any chance you could provide a reproducible case?

Sure, here it is:

```
$ cargo new zoo && cd zoo
$ cat >> Cargo.toml <<EOF
femme = "2.2.1"                                                    
getrandom = { version = "0.2.15", features = ["js"] }              
rand = "0.8.5"                                                     
x509-cert = { version = "0.2.5", features = ["hazmat", "builder"] }
EOF
$ cargo check --target wasm32-unknown-unknown
```

Note: the `--target` option is important, as the non-wasm target builds fine.
I just noticed I didn't include that bit in the description, sorry about that.

tarcieri (approved):

istankovic:
@tarcieri thanks for merging. Would it be possible to release a 0.2.6 of x509-cert soon so we can move to it?

tarcieri:
Note that the current `master` is v0.3.0-pre. We'd need to do a backport.

baloo:
I think this fix is hiding the real root cause. I believe we have a bug that is latent in the `der_derive` (as in: we don't specify completely the `PartialEq` we expect), and I'd rather fix that instead.

The error output is:
```
   = note: multiple `impl`s satisfying `u32: PartialEq<_>` found in the following crates: `core`, `serde_json`:
           - impl PartialEq<serde_json::value::Value> for u32;
           - impl<host> PartialEq for u32
             where the constant `host` has type `bool`;
   = note: required for `&u32` to implement `PartialEq<&_>`
   = note: this error originates in the derive macro `Sequence` (in Nightly builds, run with -Z macro-backtrace for more info)
```

Which happens because `femme` with wasm target brings in `serde_json` which specifies its `PartialEq` impls

baloo:
```
diff --git a/der/derive/src/sequence/field.rs b/der/derive/src/sequence/field.rs
index 3fb18320..070a67dc 100644
--- a/der/derive/src/sequence/field.rs
+++ b/der/derive/src/sequence/field.rs
@@ -97,7 +97,7 @@ impl SequenceField {
                 !attrs.optional,
                 "`default`, and `optional` are mutually exclusive"
             );
-            lowerer.apply_default(&self.ident, default);
+            lowerer.apply_default(&self.ident, default, &self.field_type);
         }

         lowerer.into_tokens()
@@ -189,11 +189,14 @@ impl LowerFieldEncoder {
     }

     /// Handle default value for a type.
-    fn apply_default(&mut self, ident: &Ident, default: &Path) {
+    fn apply_default(&mut self, ident: &Ident, default: &Path, field_type: &Type) {
         let encoder = &self.encoder;

         self.encoder = quote! {
-            if &self.#ident == &#default() {
+            if &self.#ident == &{
+                let default: #field_type = #default();
+                default
+            } {
                 None
             } else {
                 Some(#encoder)
```

A bit of a ugly hack, but that would seem to fix the output of Default

istankovic:
> Note that the current `master` is v0.3.0-pre. We'd need to do a backport.

Well, perhaps a backport might not be needed, I'd be happy to check if we can easily move on to v0.3, once it is released. Not sure when you plan for v0.3, though.

tarcieri:
@baloo that fix looks fine to me, although you could probably also avoid the let binding by writing it inline as:

```rust
<#field_type as Default>::default()
```

@istankovic it will probably be awhile before we have a final release out

baloo:
> writing it inline
ha right.

I think I will send a followup and revert this, and change the derive instead.

baloo:
I can't use the inline because what I get is the full function (which might not be `Default::default`).

PR #1455, opened 2024-07-26
const-oid v0.10.0-rc.0

tarcieri:
This release is a soft API freeze, and we can start relaxing requirements not to use `=`

cc @baloo


tarcieri:
Note: I'm releasing this individually and not bumping the other crates' dependencies since thanks to the current `=` requirements we need to update everything in lockstep

Notably we need to get `rsa` updated to bump `x509-cert`

PR #1461, opened 2024-07-27
Use latest prereleases

tarcieri:
Updates the following dependencies to their latest prerelease versions:

- `aes-gcm` v0.11.0-pre.1
- `async-signature` v0.6.0-pre.4
- `cbc` v0.2.0-pre.1
- `des` v0.9.0-pre.1
- `digest` v0.11.0-pre.9
- `ecdsa` v0.17.0-pre.7
- `pbkdf2` v0.13.0-pre.1
- `scrypt` v0.12.0-pre.1
- `sha1` v0.11.0-pre.4
- `sha2` v0.11.0-pre.4
- `sha3` v0.11.0-pre.4
- `signature` 2.3.0-pre.4
- `whirlpool` v0.11.0-pre.4

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -19,43 +19,33 @@ checksum = "f26201604c87b1e01bd3d98f8d5d9a8fcbb815e8cedb41ffccbeb4bf593a35fe"
 
 [[package]]
 name = "aead"
-version = "0.6.0-pre.0"
-source = "git+https://github.com/RustCrypto/traits.git#0a3687b58e59d5d2e196f59ca883a2d46eb76abb"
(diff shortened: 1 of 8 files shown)


tarcieri:
Some nasty circular dependencies with `rsa` here. I'm trying to open this so I can update `rsa` separately and then try to pull it back in.

baloo:
We could extract the tests based on RSA or p256 from the workspace altogether. Keep the tests, but put those in a separate crate.

baloo:
(In any case, thanks for taking care of the releases)

tarcieri:
That would definitely help for these sorts of upgrades

Issue #1471, opened 2024-08-11
der_derive: could not compile with syn v2.0.73


yayanyang:
error[E0502]: cannot borrow `generics.params` as mutable because it is also borrowed as immutable

--> sequence.rs:61:29
   |
57 |           let lifetime = generics
   |                          --------
   |                          |
   |  ________________________immutable borrow occurs here
   | |
58 | |             .lifetimes()
   | |________________________- a temporary with access to the immutable borrow is created here ...
...
61 |               .unwrap_or_else(|| {
   |                               ^^ mutable borrow occurs here
62 |                   let lt = default_lifetime();
63 | /                 generics
64 | |                     .params
   | |___________________________- second borrow occurs due to use of `generics.params` in closure
...
67 |               });
   |                 - ... and the immutable borrow might be used here, when that temporary is dropped and runs the destructor for type `impl Iterator<Item = &LifetimeParam>`

foxyseta:
Same here.

yayanyang:
> Same here.

Set syn="=2.0.72"

tarcieri:
What changed in `syn`?

tarcieri:
Opened an upstream issue in `syn` here: https://github.com/dtolnay/syn/issues/1718

PR #1472, opened 2024-08-11
der_derive: add workaround for `syn` v2.0.73

tarcieri:
This workaround gets things compiling on `syn` v2.0.73.

See #1471, dtolnay/syn#1718


tarcieri:
This was fixed in `syn` v2.0.74 so we don't need a workaround, but it would be good to fix the fuzzer to use the in-tree copy of `x509-cert` in the certificate fuzzer

Issue #1475, opened 2024-08-17
der: `Reader::peek_into`


tarcieri:
As I was looking at #1474, the buffering logic for `PemDecoder` seems quite complicated, and somewhat duplicates buffering logic already in `base64ct` and `pem-rfc7468`.

I think what would really help is if we could both push the peeking/buffering as far down the stack as possible (i.e. into `base64ct` where there's already a bunch of existing buffering logic), and also be able to have provided methods for `peek_byte` and `peek_header`.

I think what we could potentially do is expose a method analogous to `read_into`, e.g.:

```rust
fn peek_into(&mut self, buf: &'o mut [u8]) -> Result<(), Error> { ... }
```

This would fill the provided buffer with the next data in the reader, but not update its internal state.

With this, it would be trivial to have provided methods for `peek_byte` and `peek_header`.

Alternative to #1279.

tarcieri:
I think with this I can revert #1418 and have immutable peeking

tarcieri:
Added in #1478

PR #1476, opened 2024-08-17
der: `PemReader` simplifications

tarcieri:
Removes the internal buffer added in #839 in favor of just cloning the `PemReader`'s state.

This is a partial step towards #1475 and would greatly simplify the implementation of #1474.

This also eliminates the need for mutable access in `Reader::peek`, making it possible to revert #1418 as well.


tarcieri:
Note: I looked at implementing #1475 and it will be a lot easier after #1474 since `NestedReader` doesn't impl `Clone`.

PR #1485, opened 2024-08-21
Add support for KEMRecipientInfo as defined in RFC9629

carl-wallace:


cms/src/kemri.rs
```diff
@@ -0,0 +1,65 @@
...
+/// [RFC9629 Section 3]: https://datatracker.ietf.org/doc/html/rfc9629#section-3
+#[derive(Clone, Debug, Eq, PartialEq, Sequence)]
+#[allow(missing_docs)]
+pub struct KemRecipientInfo {
+    pub version: CmsVersion,
+    pub rid: RecipientIdentifier,
+    pub kem: AlgorithmIdentifierOwned,
+    pub kem_ct: OctetString,
+    pub kdf: AlgorithmIdentifierOwned,
+    pub kek_length: u16,
```
baloo, line 43: any opinion about making that a `NonZero<u16>`?

That will require the `DecodeValue`/`EncodeValue` for that (and a bump release of der) but from a type system standpoint this is better?

I don't know if that's worth the effort.

baloo (approved):

cms/src/kemri.rs
carl-wallace: Checking for non-zero would only satisfy part of the requirement (which I'd tend to leave to the decryption component). Here's the snip from the RFC:

```
Implementations MUST confirm that the value provided is consistent with the key-encryption algorithm identified in the wrap field below.
```

baloo: That would look something like: https://github.com/carl-wallace/formats/compare/kemri...baloo:formats:baloo/kemri

PR #1505, opened 2024-09-06
x509-cert: make (Tbs)`CertificateInner` fields private

### LINKED ISSUE #1486: x509-cert: Only allow `Certificate`s to be constructed via either the builder or deserialization
The `Certificate` type is currently completely transparent:

https://github.com/RustCrypto/formats/blob/e0c3e08bd5fb8ee9c30aed7de271e2a25db5d531/x509-cert/src/certificate.rs#L183
https://github.com/RustCrypto/formats/blob/e0c3e08bd5fb8ee9c30aed7de271e2a25db5d531/x509-cert/src/certificate.rs#L199-L203
https://github.com/RustCrypto/formats/blob/e0c3e08bd5fb8ee9c30aed7de271e2a25db5d531/x509-cert/src/certificate.rs#L119-L144

This enables arbitrarily-invalid `Certificate`s to be constructed. In particular, this means that the `P: Profile` generic parameter cannot be trusted; if a downstream API receives an `&CertificateInner<Rfc5280>`, they have _zero_ guarantee that it is actually valid under the RFC 5280 profile.

The `x509-cert` crate's public API should be altered to make `Certificate` (and all types below it) opaque, replacing direct field access with getter methods. This would then enable `Certificate`s to be correct-by-construction, by enforcing the relevant `Profile` in the builder or during deserialization.

tarcieri:
This allows them to maintain invariants.

Adds read-only accessor methods in place of public fields.

Closes #1486.

### DIFF
--- a/x509-cert/src/certificate.rs
+++ b/x509-cert/src/certificate.rs
@@ -228,7 +228,7 @@ impl<P: Profile> TbsCertificateInner<P> {
     /// use x509_cert::{der::DecodePem, ext::pkix::BasicConstraints, Certificate};
     /// let certificate = Certificate::from_pem(CERT_PEM.as_bytes()).expect("parse certificate");
     ///
-    /// let (critical, constraints) = certificate.tbs_certificate.get_extension::<BasicConstraints>()
+    /// let (critical, constraints) = certificate.tbs_certificate().get_extension::<BasicConstraints>()
     ///     .expect("Failed to parse extension")
     ///     .expect("Basic constraints expected");
     /// # let _ = constraints;
@@ -264,7 +264,7 @@ impl<P: Profile> TbsCertificateInner<P> {
     /// use x509_cert::{der::DecodePem, ext::pkix::BasicConstraints, Certificate};
     /// let certificate = Certificate::from_pem(CERT_PEM.as_bytes()).expect("parse certificate");
     ///
-    /// let mut extensions_found = certificate.tbs_certificate.filter_extensions::<BasicConstraints>();
+    /// let mut extensions_found = certificate.tbs_certificate().filter_extensions::<BasicConstraints>();
     /// while let Some(Ok((critical, extension))) = extensions_found.next() {
     ///     println!("Found (critical={critical}): {extension:?}");
     /// }
--- a/x509-cert/src/certificate.rs
+++ b/x509-cert/src/certificate.rs
@@ -210,9 +210,8 @@ impl<P: Profile> TbsCertificateInner<P> {
     }
 
     /// Certificate extensions.
-    pub fn extensions(&self) -> &ext::Extensions {
-        static EMPTY_EXTENSIONS: ext::Extensions = Vec::new();
-        self.extensions.as_ref().unwrap_or(&EMPTY_EXTENSIONS)
+    pub fn extensions(&self) -> Option<&ext::Extensions> {
+        self.extensions.as_ref()
     }
 
     /// Decodes a single extension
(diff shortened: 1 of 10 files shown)


x509-cert/src/certificate.rs
```diff
@@ -143,31 +143,85 @@ pub struct TbsCertificateInner<P: Profile + 'static = Rfc5280> {
...
+    /// Subject unique ID.
+    pub fn subject_unique_id(&self) -> &Option<BitString> {
+        &self.subject_unique_id
+    }
+
+    /// Certificate extensions.
+    pub fn extensions(&self) -> &ext::Extensions {
+        static EMPTY_EXTENSIONS: ext::Extensions = Vec::new();
+        self.extensions.as_ref().unwrap_or(&EMPTY_EXTENSIONS)
+    }
```
tarcieri, line 216: Note: placeholder rustdoc comments. Would definitely like to write some better ones.

tarcieri:
Note: this is just a first pass for `TbsCertificateInner`. I intend to make at least `CertificateInner`'s fields private as well.

x509-cert/src/certificate.rs
tarcieri, line 216: The `EMPTY_EXTENSIONS` static is a bit of a hack but I feel like it makes the API a bit less cumbersome

x509-cert/src/certificate.rs
```diff
@@ -143,31 +143,85 @@ pub struct TbsCertificateInner<P: Profile + 'static = Rfc5280> {
...
+        &self.issuer_unique_id
+    }
+
+    /// Subject unique ID.
+    pub fn subject_unique_id(&self) -> &Option<BitString> {
+        &self.subject_unique_id
+    }
+
+    /// Certificate extensions.
+    pub fn extensions(&self) -> &ext::Extensions {
```
baloo, line 213: I'm curious about the move to non-optional list of extensions here.
I think I would see the value of keeping the option.

tarcieri: There were several test examples which moved the extensions out of the struct, which isn't possible with an immutable borrow.

I first went down the path of cloning/unwrapping, but the ergonomics for accessing the extensions are a lot simpler this way.

What is use case for `Option::None` versus an empty set, particularly within the context of a public accessor method? Is there something semantically meaningful to distinguish there?

baloo: I don't know. There is a semantic in an `Option<Vec<Extension>>` as this represents how this is serialized.

`None` will skip the field altogether, `Some<empty vec>` will put an empty sequence.

Is there value for a consumer? I'm not entirely sure.

tarcieri: In the interest of moving this along and keeping the changes minimal I changed it to return `Option<&ext::Extension>` in [0f9b52b](https://github.com/RustCrypto/formats/pull/1505/commits/0f9b52bc3f13a8eb1dd6fb92187d2745685684a8), however I think it would still be good to circle back on this in a separate PR

x509-cert/src/certificate.rs
tarcieri: ...and as of [0f9b52b](https://github.com/RustCrypto/formats/pull/1505/commits/0f9b52bc3f13a8eb1dd6fb92187d2745685684a8) it no longer exists, resolving

baloo (approved):

x509-cert/src/certificate.rs
tarcieri: Better docs added in [53c16e6](https://github.com/RustCrypto/formats/pull/1505/commits/53c16e6ad13740a9b4f0ee72551c88719288e25f)

tarcieri:
There are definitely more types that should be given this treatment, but this seems like a reasonable first pass

PR #1508, opened 2024-09-08
x509-cert: make `RdnSequence`'s inner field private

tarcieri:
Explicitly delegates the `iter`, `len`, and `push` methods.


tarcieri:
Note: `RelativeDistinguishedName` should probably get a similar treatment

PR #1515, opened 2024-09-10
serdect: return the deserialized buffer during deserialization

### LINKED ISSUE #1322: serdect: `slice::deserialize_hex_or_bin` needs to return the amount of deserialized data
This is tracking a regression from #1112, where `slice::deserialize_hex_or_bin` was changed to return `Result<(), D::Error>` instead of `Result<&[u8], D::Error>`.

Notably this API is intended for use cases where the amount of deserialized data can vary, and may be shorter than the provided `buffer` parameter. The ability to handle messages shorter than `buffer` has been lost.

See https://github.com/RustCrypto/formats/pull/1112#discussion_r1446473939

fjarri:
Fixes #1322

Also fixed a TODO in `StrIntoBufVisitor::visit_str` by mapping `base16ct` errors to `serde` errors.

### DIFF
--- a/serdect/src/common.rs
+++ b/serdect/src/common.rs
@@ -74,7 +74,7 @@ pub(crate) trait LengthCheck {
 pub(crate) struct StrIntoBufVisitor<'b, T: LengthCheck>(pub &'b mut [u8], pub PhantomData<T>);
 
 impl<'de, 'b, T: LengthCheck> Visitor<'de> for StrIntoBufVisitor<'b, T> {
-    type Value = ();
+    type Value = &'b [u8];
 
     fn expecting(&self, formatter: &mut fmt::Formatter<'_>) -> fmt::Result {
         T::expecting(formatter, "a string", self.0.len() * 2)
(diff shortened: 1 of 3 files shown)


serdect/src/common.rs
```diff
@@ -84,13 +84,15 @@ impl<'de, 'b, T: LengthCheck> Visitor<'de> for StrIntoBufVisitor<'b, T> {
...
-        // TODO: Map `base16ct::Error::InvalidLength` to `Error::invalid_length`.
-        base16ct::mixed::decode(v, self.0)
-            .map(|_| ())
-            .map_err(E::custom)
+        base16ct::mixed::decode(v, self.0).map_err(|err| match err {
+            base16ct::Error::InvalidLength => {
+                Error::invalid_length(v.len(), &"an even number of hex digits")
+            }
+            base16ct::Error::InvalidEncoding => Error::invalid_value(
+                Unexpected::Str(v),
```
fjarri, line 92: Actually, can this be secret? Should we not return it?

fjarri: Replaced with a placeholder.

tarcieri (approved):

PR #1521, opened 2024-09-16
Allow PBES2 encoding without NULL value for the algId prf parameter

ancwrd1:
Rationale: some applications and libraries produce PBES2/PBKDF2 structures without the terminating NULL value specified in the prf algorithm field.
This PR relaxes the decoding requirements.

According to [RFC4231](https://datatracker.ietf.org/doc/html/rfc4231#section-3.1) the NULL parameter should be present but is not an absolute requirement.

See for example this issue: https://github.com/ancwrd1/p12-keystore/issues/2


tarcieri:
> some applications and libraries produce PBES2/PBKDF2 structures without the terminating NULL value specified in the prf algorithm field.

Which ones? Can you include a reference as to how the test vector you're providing was generated, at least in the comments?

ancwrd1:
Specifically, this: https://github.com/smallstep/cli
It produces PKCS#12 file without the algorithm parameter value. It works fine with e.g. openssl or Java KeyStore API, but fails with my p12-keystore crate which is based on RustCrypto.

I have added a note in the comments.

tarcieri (approved):

PR #1522, opened 2024-09-18
const-oid: fix large arc handling in encoder

### LINKED ISSUE #1520: Incorrect Object Identifier DER Encoding
The encoding seems to be incorrect for certain object identifiers at the boundary of arc length. For example:

```
// decoded: 1.2.0
println!("decoded: {:?}",  ObjectIdentifier::from_der(&ObjectIdentifier::new_unwrap("1.2.128").to_der().unwrap()));

// decoded: 1.2.0.0
println!("decoded: {:?}",  ObjectIdentifier::from_der(&ObjectIdentifier::new_unwrap("1.2.16384").to_der().unwrap()));
```

I'm using this crate:
```
der = { version = "0.7.9", features = [ "alloc", "oid" ] }
```

tarcieri:
BER encodings were being miscomputed for certain large arcs.

The previous method was a bit wacky (in addition to being buggy) and attempted to encode each arc backwards within the BER output buffer.

This switches to a new method which splits the upper 7 bits from an arc and encodes that as a byte, continuing until all bytes of the arc have been encoded, which is much more straightforward.

The problematic cases which were reported have now been corrected.

Fixes #1520

cc @zhengyao-lin

### DIFF
--- a/const-oid/src/encoder.rs
+++ b/const-oid/src/encoder.rs
@@ -73,27 +73,9 @@ impl<const MAX_SIZE: usize> Encoder<MAX_SIZE> {
                 self.cursor = 1;
                 Ok(self)
             }
-            // TODO(tarcieri): finer-grained overflow safety / checked arithmetic
-            #[allow(clippy::arithmetic_side_effects)]
             State::Body => {
-                // Total number of bytes in encoded arc - 1
                 let nbytes = base128_len(arc);
-
-                // Shouldn't overflow on any 16-bit+ architectures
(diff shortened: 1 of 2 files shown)


tarcieri:
Note: with `const_mut_refs` now stable (rust-lang/rust#57349) and shipping in Rust 1.83, we will be able to completely rewrite all of the parsing/encoding logic in a much more straightforward way before the next release.

This PR isn't great, but at least stops the bleeding.

PR #1534, opened 2024-09-27
x509-cert: re-export `spki` types without `*Owned` suffixes

tarcieri:
Re-exports the following with shorter names:
- `spki::AlgorithmIdentifierOwned` => `x509_cert::AlgorithmIdentifier`
- `spki::SubjectPublicKeyInfoOwned` => `x509_cert::SubjectPublicKeyInfo`

These names are already quite long to begin with even without the `*Owned` suffix.

The `x509-cert` crate has a hard dependency on `alloc` and always uses the `*Owned` types, so this allows for more convenient names in this scenario.


tarcieri:
Thought shorter names would be nice, but it does risk being a bit confusing in regard to `x509_cert::AlgorithmIdentifier` not being the same as `spki::AlgorithmIdentifier`, especially when accessed as `x509_cert::spki`.

baloo (approved):

Issue #1559, opened 2024-10-08
der: Custom error types in derive macros.


turbocool3r:
Currently `der_derive` only provides macros that implement `DecodeValue` trait with `Error` associated type set to `::der::Error`. It would be nice to have support for custom error types via an attribute as discussed in #1053.

~~# Current blocker~~

~~The problem is `der::Reader` lacks support for error types in its [`sequence`](https://docs.rs/der/latest/der/ trait.Reader.html#method.sequence), [`read_nested`](https://docs.rs/der/latest/der/trait.Reader.html#method.read_nested), [`decode`](https://docs.rs/der/latest/der/trait.Reader.html#method.decode) and [`context_specific`](https://docs.rs/der/latest/der/trait.Reader.html#method.context_specific) methods. I think each of these should gain an additional generic parameter for error type.~~

I have a patch almost ready so if this is ok, I will move on to create a PR.

Update: after some review I see generic parameters are already implemented in my previous PR 😄

tarcieri:
Seems like we can close this now that #1560 is merged?

turbocool3r:
Yes, closing it

PR #1560, opened 2024-10-08
der: Custom error types in derive macros.

turbocool3r:
Add support for specifying a custom decode error type for `Sequence`, `Enumerated` and `Choice` derive macros using the `#[asn1(error = Ty)]` attribute.

This addresses #1559.


baloo (approved):
That looks good, the only thing that comes to mind would be to use a type to hold the attribute instead.

I've made a PR of what I had in mind in https://github.com/turbocool3r/formats/pull/1.

But it's mostly up to you.

turbocool3r:
Thanks, just merged, that's a nice way to reduce that code duplication.

One last question is should the attribute be named something like `decode_error_type` to make its purpose more clear?

baloo:
It's already `asn1(error=...)`, I think it's good enough.

der_derive/src/enumerated.rs
```diff
@@ -24,6 +23,9 @@ pub(crate) struct DeriveEnumerated {
 
     /// Variants of this enum.
     variants: Vec<EnumeratedVariant>,
+
+    /// Error type for `DecodeValue` implementation.
+    error: Option<Path>,
```
baloo, line 28: Did I forget this one?

baloo: https://github.com/turbocool3r/formats/pull/2

baloo:
I can merge that later tonight and make an rc.1 release for it.

turbocool3r:
Merged the second PR, also added docs. I think it's ready for merge.

PR #1576, opened 2024-10-16
CI: disable `security-audit` cache

tarcieri:
We're encountering errors like:

      /home/runner/.cargo/bin/cargo-audit: /lib/x86_64-linux-gnu/libc.so.6: version `GLIBC_2.39' not found (required by /home/runner/.cargo/bin/cargo-audit)

These are occurring even with freshly-built binaries.

To avoid blocking PRs, temporarily disable the build cache until we have time to properly investigate.

### DIFF
--- a/.github/workflows/security-audit.yml
+++ b/.github/workflows/security-audit.yml
@@ -21,10 +21,11 @@ jobs:
       - uses: dtolnay/rust-toolchain@master
         with:
           toolchain: stable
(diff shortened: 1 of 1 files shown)


baloo:
I think we got caught in the ubuntu-latest upgrade:
https://github.blog/changelog/2024-09-25-actions-new-images-and-ubuntu-latest-changes/

PR #1579, opened 2024-10-20
cms: ECC KeyAgreementRecipientInfo initial support

### LINKED ISSUE #1544: ECC in CMS support
Hello,

I believe that the current code base does not support `EnvelopedData` with Elliptic Curve Cryptography (ECC) for CMS.
Would you be interested in a code contribution that would bring partial CMS ECC support as per [rfc5753](https://datatracker.ietf.org/doc/html/rfc5753.html)?

I could contribute some code that would essentially implement `KeyAgreeRecipientInfoBuilder` for `KeyAgreeRecipientInfo` (Kari).

It would not implement all the KeyAgreement  algorithms mentioned in the RFC as I think that RustCrypto does not currently have all the primitive to fully support other KeyAgreement schemes (namely `'Co-factor' ECDH` and `1-Pass ECMQV`). As such it would mainly focus on [EnvelopedData Using (ephemeral-static) ECDH](https://datatracker.ietf.org/doc/html/rfc5753.html#section-3.1) with `'Standard' ECDH` 

A few initial prerequisite would be:
- ANSI-X9.63-KDF support - (~~Pending - I already opened a PR in KDFs~~ Done in https://github.com/RustCrypto/KDFs/issues/101)
- Missing OIDs from rfc5753 -  ~~I can open another issue here if there is interest~~ (~~Pending~~ Done in https://github.com/RustCrypto/formats/issues/1546).

Thank you and let me know your thoughts,

Best,

nemynm:
This PR intends to bring initial support for Elliptic Curve Cryptography (ECC) for CMS - addressing https://github.com/RustCrypto/formats/issues/1544.

- Following [rfc5753](https://datatracker.ietf.org/doc/html/rfc5753), it implements `KeyAgreementRecipientInfoBuilder` for ECC. For now only `EnvelopedData` using `(ephemeral-static) ECDH` is supported.
- It does not include reader/decryption logic (it could be part of another PR if there is interest).
- It does not support `SHA1` and `3DES` schemes (For SHA1, it could maybe be introduced and gated behind a feature to parse/read/decrypt older/legacy incoming CMS message that use them. However RustCrypto does not seem to support 3DES key wrap)

1. Utils functions:
    - KDF and key-wrapping operation are hosted in their own submodules.
    - Key-derivation uses an internal `HashDigest` enum to select the hash function to use during key-derivation
    - `KeyWrapper` is a struct that aims at abstracting the key-wrapping logic for different AES algorithm and  different incoming key size.

2. Key Agreement Recipient Info:
    - `KeyAgreementRecipientInfoBuilder` is hosted in its own submodule.
    - Supported schemes and algorithm are represented using enums. At least three are needed for the builder:
       - `EcKeyEncryptionInfo` - recipient public key (generic over [RustCrypto elliptic-curve](https://github.com/RustCrypto/traits/tree/master/elliptic-curve)).
           While `EcKeyEncryptionInfo` is essentially the ECC equivalent of the existing `KeyEncryptionInfo`, it has been introduced to limit API breakage - as it introduces a generic over the chosen elliptic-curve.
       - `KeyAgreementAlgorithm` - key agreement as per RFC (SHA1 schemes are not supported on purpose, can be amended if needed).
       - `KeyWrapAlgorithm` - key-wrap algorithm to use.
           For the sake of simplicity `From<ContentEncryptionAlgorithm> for KeyWrapAlgorithm` trait has been implemented.

3. Testing:
    Unit tests have been written in the relevant files. One integration test showcasing `KeyAgreeRecipientInfoBuilder` is available, leveraging the existing test P256 key material. Obtained message can be decrypted using:
   ```bash
   openssl cms -decrypt -inkey cms/tests/examples/p256-priv.der -inform PEM
   ```
   
Fixes #1544


cms/Cargo.toml
```diff
@@ -22,9 +22,12 @@ x509-cert = { version = "=0.3.0-pre.0", default-features = false }
 
 # optional dependencies
 aes = { version = "=0.9.0-pre.2", optional = true }
+aes-kw = { version ="0.2.1", optional = true }
```
baloo, line 25: This duplicates the dependency on aes and cipher, we should bump aes-kw to use "aes 0.9.0-pre.2" & "cipher 0.5.0-pre.7".

cms/src/builder/kari.rs
```diff
@@ -0,0 +1,446 @@
...
+/// - dhSinglePass-stdDH-sha224kdf-scheme
+/// - dhSinglePass-stdDH-sha256kdf-scheme
+/// - dhSinglePass-stdDH-sha384kdf-scheme
+/// - dhSinglePass-stdDH-sha512kdf-scheme
+///
+/// [RFC 5753 Section 7.1.4]: https://datatracker.ietf.org/doc/html/rfc5753#section-7.1.4
+/// [RFC 5753 Section 8]: https://datatracker.ietf.org/doc/html/rfc5753#section-8
+#[allow(clippy::enum_variant_names)]
+#[derive(Clone, Copy)]
+pub enum KeyAgreementAlgorithm {
```
baloo, line 101: Could we convert that to a trait instead:
```
pub trait KeyAgreementAlgorithm: AssociatedOid {
    fn kdf(secret: &[u8]);
}
```

And then make a struct like:
```
struct DhSinglePassStdDhKdf<D> where D:Digest + {
    digest: PhantomData<D>,
}

impl<D> KeyAgreementAlgorithm for DhSinglePassStdDhKdf<D> where D:Digest {
    fn kdf(secret: &[u8]) {
        ansi_x963_kdf::derive_key_into::<D>(secret)
    }
}

impl AssociatedOid for DhSinglePassStdDhKdf<sha2::Sha224> {
   const OID: ObjectIdentifier = rfc5753::DH_SINGLE_PASS_STD_DH_SHA_224_KDF_SCHEME;
}
```

cms/src/builder/utils/kw.rs
```diff
@@ -0,0 +1,571 @@
...
+    ///
+    /// In all cases the parameters field MUST be absent.
+    /// ```
+    ///
+    /// [RFC 3565 Section 2.3.2]: https://datatracker.ietf.org/doc/html/rfc3565#section-2.3.2
+    fn oid(&self) -> ObjectIdentifier {
+        match self {
+            Self::Aes128 => const_oid::db::rfc5911::ID_AES_128_WRAP,
+            Self::Aes192 => const_oid::db::rfc5911::ID_AES_192_WRAP,
+            Self::Aes256 => const_oid::db::rfc5911::ID_AES_256_WRAP,
```
baloo, line 71: TODO: We should implement AssociatedOid in https://github.com/RustCrypto/key-wraps/tree/master/aes-kw

cms/Cargo.toml
baloo: https://github.com/RustCrypto/key-wraps/pull/34

cms/src/builder/utils/kw.rs
baloo: https://github.com/RustCrypto/key-wraps/pull/35

cms/Cargo.toml
baloo: https://github.com/nemynm/formats/pull/1

cms/src/builder/kari.rs
baloo: I had something like that in mind:
https://github.com/nemynm/formats/pull/2

cms/Cargo.toml
nemynm: Thanks, I'll merge once it's ready

cms/src/builder/utils/kw.rs
nemynm: Thanks will merge once its merged upstream

cms/src/builder/kari.rs
nemynm: Thanks!
Indeed I was not particularly convinced by the current KDF handling, but I thought it was a fair solution that was easily extensible for other schemes/algo (maybe with the help of a quick macro to derive `try_ansi_x963_kdf`).

At first I actually went the trait route in my code (close-ish to what you proposed), but then settled on the enum and dispatch function. Rationales were:
1. This snippet in `EnvelopedDataBuilder` at [builder.rs#L890](https://github.com/RustCrypto/formats/blob/master/cms/src/builder.rs#L890) which caught my eye:
```
    // TODO bk Not good to offer both, `content_encryptor` and `content_encryption_algorithm`.
    // We should
    // (1) either derive `content_encryption_algorithm` from `content_encryptor` (but this is not
    //            yet supported by RustCrypto),
    // (2) or     pass `content_encryption_algorithm` and create an encryptor for it.
    // In the first case, we might need a new trait here, e.g. `DynEncryptionAlgorithmIdentifier` in
    // analogy to `DynSignatureAlgorithmIdentifier`.
    // Going for (2)
    //  content_encryptor: E,
    content_encryption_algorithm: ContentEncryptionAlgorithm,
```

Even if its not the same situation, I interprated it as "work with enum rather than trait", at least until some is finalized. That was somewhat re-inforced by the current [get_hasher()](https://github.com/RustCrypto/formats/blob/master/cms/src/builder.rs#L1061) function.

2. Consequently, just passing an enum looked closer to the current API for other builders (having to specify the KA generic in `KeyAgreeRecipientInfoBuilder` seemed a tad less user-friendly).

3. I thought that if we ended up supporting the other key agreement algorithm (e.g. cofactor ECDH and 1-Pass ECMQV ), we may want to have a more complete trait like:
```
pub trait KeyAgreementAlgorithm: AssociatedOid {
    fn shared_secret()
    fn try_kdf(secret: &[u8], other_info: &[u8], key: &mut impl AsMut<[u8]>) -> Result<()>;
}
```
-> So essentially folding most of the `build()` logic behind a trait - so to speak.

However not exactly knowing yet:
- what `shared_secret()` signature should look like for all.
- the input to `try_kdf()` would be with other schemes (would that always be a `SharedSecret<C>` or something more `generic` like `&[u8]`).
- if we should have three structs (`DhSinglePassStdDhKdf`, `DhSinglePassCoFactorDhKdf`, `MqvSinglePassKdf`) or maybe just one generic for the three schemes.
 
I figured that a trait was maybe not the best solution yet.

Anyway as you can see, these are more generic thoughts than strong reasons not to, so I am also ok to go the trait route now too.

cms/src/builder/utils/kw.rs
baloo: we will need to move that to a generic parameter as well

nemynm: I could add a trait for Key Wrapping too. Do you want the trait to only implement AssociateOid, or the key wrapping logic itself (a keen to kdf for KeyAgreement trait)?

cms/Cargo.toml
nemynm: Merged

cms/src/builder/utils/kw.rs
nemynm: > we will need to move that to a generic parameter as well

@baloo, I came up with a trait-based approach for key-wrapping, a first draft is available here https://github.com/nemynm/formats/pull/3, let me know your thoughts and if this is the kind of things you had in mind.

nemynm:
> @nemynm can you rebase?

Sure will do, I'll also mark it as draft untill changes done during initial review by @baloo have been included.

baloo (approved):

baloo:
I can't merge this before we re-integrate x509 and cms in the workspace. I'll need to revisit once that's done.
Sorry for the delay.

nemynm:
No worries, in the meantime I'll rebase the PR regularly.

baloo:
I don't know if rebase is something you can do until we re-unite the workspace.

baloo:
Could you either merge master or rebase? Thanks a lot!

nemynm:
@baloo rebased it is.
I had a few issues solving the dependency tree:
- I had to patch `async-signature` from git to bring it in sync with `signature`
- patching `ansi-x963-kdf` from its specific `master` branch does not seem necessary after the fact, but somehow helped cargo figuring it out, so I left it for now.

Also some adjustments with latest `CryptoRng` changes.

Let me know any other needed adjustments.

baloo:
async-signature should be deprecated now (it moved to `signature`)

Could you try this patch:
<details>

```
diff --git a/Cargo.lock b/Cargo.lock
index c0bbcbbe35..2029b3663e 100644
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -77,7 +77,7 @@
 [[package]]
 name = "ansi-x963-kdf"
 version = "0.0.1"
-source = "git+https://github.com/RustCrypto/KDFs.git?branch=master#b1d7fe67b3053deef498563adcf415ec631d1cd8"
+source = "git+https://github.com/RustCrypto/KDFs.git#b1d7fe67b3053deef498563adcf415ec631d1cd8"
 dependencies = [
  "digest",
 ]
@@ -98,14 +98,6 @@
 ]

 [[package]]
-name = "async-signature"
-version = "0.6.0-pre.4"
-source = "git+https://github.com/RustCrypto/traits.git#2dc47f8d1461a2a7a22b68e2afafeb4b59e13420"
-dependencies = [
- "signature",
-]
-
-[[package]]
 name = "autocfg"
 version = "1.4.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -309,7 +301,6 @@
  "aes",
  "aes-kw",
  "ansi-x963-kdf",
- "async-signature",
  "cbc",
  "cipher",
  "const-oid",
diff --git a/Cargo.toml b/Cargo.toml
index 87074a1173..490f768d86 100644
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -72,17 +72,14 @@
 # https://github.com/RustCrypto/key-wraps/pull/39
 aes-kw = { git = "https://github.com/RustCrypto/key-wraps.git" }

-
 # https://github.com/RustCrypto/KDFs/pull/102
-ansi-x963-kdf = { git = "https://github.com/RustCrypto/KDFs.git", branch = "master" }
-
+ansi-x963-kdf = { git = "https://github.com/RustCrypto/KDFs.git" }

 # https://github.com/RustCrypto/traits/pull/1777
 crypto-common  = { git = "https://github.com/RustCrypto/traits.git" }
 elliptic-curve = { git = "https://github.com/RustCrypto/traits.git" }
 signature      = { git = "https://github.com/RustCrypto/traits.git" }
 aead           = { git = "https://github.com/RustCrypto/traits.git" }
-async-signature = { git = "https://github.com/RustCrypto/traits.git" }

 # https://github.com/RustCrypto/RSA/pull/478
 # https://github.com/RustCrypto/RSA/pull/504
diff --git a/cms/Cargo.toml b/cms/Cargo.toml
index 57d54a337f..257ebb6791 100644
--- a/cms/Cargo.toml
+++ b/cms/Cargo.toml
@@ -24,7 +24,6 @@
 aes = { version = "=0.9.0-pre.3", optional = true }
 aes-kw = { version ="=0.3.0-pre", optional = true }
 ansi-x963-kdf = { version = "0.0.1", optional = true }
-async-signature = { version = "=0.6.0-pre.4", features = ["digest", "rand_core"], optional = true }
 cbc = { version = "=0.2.0-pre.2", optional = true }
 cipher = { version = "=0.5.0-pre.8", features = ["alloc", "block-padding", "rand_core"], optional = true }
 digest = { version = "0.11.0-pre.10", optional = true }
@@ -56,7 +55,6 @@
     "dep:aes",
     "dep:aes-kw",
     "dep:ansi-x963-kdf",
-    "dep:async-signature",
     "dep:cbc",
     "dep:cipher",
     "dep:digest",
```

</details>

cms/src/builder/utils/kw.rs
```diff
@@ -0,0 +1,444 @@
...
+            .map_err(|_| Error::Builder(String::from("could not wrap key")))?;
+        if res.len() != out.len() {
+            return Err(Error::Builder(String::from("output buffer invalid size")));
+        }
+        Ok(())
+    }
+}
+
+/*
+#[cfg(test)]
```
baloo, line 169: Can you remove those too? I think they don't apply anymore?

baloo (approved):

cms/src/builder/utils/kw.rs
nemynm: I think we still need it to wrap the key - we call it on the KW generic here:

https://github.com/nemynm/formats/blob/519646018bb92e6590a0f4e1ba74b904b4da8edc/cms/src/builder/kari.rs#L364

```
                // Wrap the Content Encryption Key (CEK) with the KEK
                KW::try_wrap(&kek, content_encryption_key, wrapped.as_mut())?;
```

nemynm: ah sorry you meant the tests yes will do

Issue #675, opened 2022-06-17
tai64: wrong results -- 10s is not right offset and doesn't account for past or future leap seconds in SystemTime conversion 


programmerjake:
The offset between UTC time and TAI time isn't currently 10s, it's 37s and may change to something other than 37s in the future.

The current code assumes the offset is always 10s.

programmerjake:
comparison on my system (after installing ntp so linux has the correct offset):
note how the result from linux and `Tai64N::now()` differ -- linux: `1655618465` vs. tai64: `1655618438` -- `SystemTime`: `1655618428`
code:
```rust
use tai64::Tai64N;
use std::time::SystemTime;
use libc::{clock_gettime, CLOCK_TAI, timespec};

fn now_tai() -> std::io::Result<timespec> {
    let mut retval = timespec { tv_sec: 0, tv_nsec: 0 };
    let r = unsafe { clock_gettime(CLOCK_TAI, &mut retval) };
    if r != 0 {
        Err(std::io::Error::last_os_error())
    } else {
        Ok(retval)
    }
}

fn main() {
    let t = dbg!(Tai64N::now());
    let t = (t.0.0 as i64).wrapping_sub(1 << 62);
    dbg!(t);
    dbg!(SystemTime::now());
    let tai = now_tai().unwrap();
    dbg!(tai.tv_sec);
    dbg!(tai.tv_nsec);
}
```
output:
```
[src/main.rs:16] Tai64N::now() = Tai64N(
    Tai64(
        4611686020083006342,
    ),
    259280636,
)
[src/main.rs:18] t = 1655618438
[src/main.rs:19] SystemTime::now() = SystemTime {
    tv_sec: 1655618428,
    tv_nsec: 259298279,
}
[src/main.rs:21] tai.tv_sec = 1655618465
[src/main.rs:22] tai.tv_nsec = 259305372
```

programmerjake:
See also:
https://lists.zx2c4.com/pipermail/wireguard/2022-June/007657.html

tarcieri:
The implementation is indeed naive/broken. I thought we had an open tracking issue for this, but apparently not.

I'm not sure the best way to handle this, especially given the existing published crates.

programmerjake:
how about downloading the official leap-second table and converting it to rust in a build.rs? you could have that enabled by an enabled-by-default feature.

[...]

PR #1583, opened 2024-10-22
Add new action to update leap seconds automatically

AurelienFT:
Resolves : https://github.com/RustCrypto/formats/issues/675

This PR adds a github actions that runs each week, to check for the leap seconds change in a reputable source and creates a PR to update if necessary.

The PR doesn't contains the process to release automatically as the process could change and maybe we don't want to directly change the version and make the release.

This current PR also include the changes to start with the current value 37 seconds.
Here is some tests I did on a fork : https://github.com/AurelienFT/formats/actions/runs/11470018305/job/31918401654 / https://github.com/AurelienFT/formats/pull/5/files

It could be nice to release a new major this time to have the correct time available for everyone.

### DIFF
--- a/.github/workflows/tai64_update_leap_seconds.yml
+++ b/.github/workflows/tai64_update_leap_seconds.yml
@@ -1,0 +1,27 @@
+name: Update Leap Seconds
+on:
+  schedule:
+    - cron: 0 0 * * 1
+
+jobs:
+  update:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+        with:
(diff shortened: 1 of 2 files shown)


tarcieri (approved):

AurelienFT:
@tarcieri Thanks a lot for the fast merge. Can we expect a release for the first update that has been done in this PR ? :p

AurelienFT:
@tarcieri Hello, any news about a possible release for this fix ?

tarcieri:
#1591 is open but it's been a bit tricky because of how long it's been since a release

Issue #1624, opened 2024-12-20
Type annotations needed when using serde_json


quexten:
Hi!

When building pkcs5 (in my case pkcs8, which brings pkcs5 as a dependency), and chacha20poly1305, for wasm, the build fails with the following error:

```
   Compiling pkcs5 v0.7.1
error[E0283]: type annotations needed
   --> /Users/quexten/.cargo/registry/src/index.crates.io-6f17d22bba15001f/pkcs5-0.7.1/src/pbes2/encryption.rs:105:32
    |
105 |             if key_size != len.into() {
    |                         --     ^^^^
    |                         |
    |                         type must be known at this point
    |
    = note: multiple `impl`s satisfying `usize: PartialEq<_>` found in the following crates: `core`, `serde_json`:
            - impl PartialEq for usize;
            - impl PartialEq<serde_json::value::Value> for usize;
help: try using a fully qualified path to specify the expected types
    |
105 |             if key_size != <u16 as Into<T>>::into(len) {
    |                            +++++++++++++++++++++++   ~

For more information about this error, try `rustc --explain E0283`.
error: could not compile `pkcs5` (lib) due to 1 previous error
```

Building for non-wasm, or disabling the chacha dependency fixes the build.

Let me know if there is any further info I can provide; I also might look into writing a PR. Though, I don't know why these crates influence each other.

tarcieri:
The error appears to occur due to an inference ambiguity caused by `serde_json`, FWIW:

```
note: multiple `impl`s satisfying `usize: PartialEq<_>` found in the following crates: `core`, `serde_json`:
            - impl PartialEq for usize;
            - impl PartialEq<serde_json::value::Value> for usize;
```

quexten:
Not sure why the only the above combination of build environment and dependencies causes this issue in my case, but I renamed this issue to reflect that it's an incompatibility with serde-json. Sounds like it's the same as this bug: https://github.com/serde-rs/json/issues/958

PR #1625, opened 2024-12-20
pkcs5: fix type annotation required error

quexten:
Fixes https://github.com/RustCrypto/formats/issues/1624

Without digging into why the error occurs, this seems to fix the error, and it seems safe to do the conversion in this instance.

### DIFF
--- a/pkcs5/src/pbes2/encryption.rs
+++ b/pkcs5/src/pbes2/encryption.rs
@@ -178,7 +178,7 @@ impl EncryptionKey {
     pub fn derive_from_password(password: &[u8], kdf: &Kdf, key_size: usize) -> Result<Self> {
         // if the kdf params defined a key length, ensure it matches the required key size
         if let Some(len) = kdf.key_length() {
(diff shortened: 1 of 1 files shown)


pkcs5/src/pbes2/encryption.rs
```diff
@@ -178,7 +178,7 @@ impl EncryptionKey {
     pub fn derive_from_password(password: &[u8], kdf: &Kdf, key_size: usize) -> Result<Self> {
         // if the kdf params defined a key length, ensure it matches the required key size
         if let Some(len) = kdf.key_length() {
-            if key_size != len.into() {
+            if key_size != len as usize {
```
tarcieri, line 181:
```suggestion
            if key_size != usize::from(len) {
```

I'm surprised `clippy` didn't catch that

quexten: Thank you for the lightning quick review!

tarcieri (approved):

PR #1643, opened 2025-02-01
x509-cert: enable and fix clippy `core`/`alloc`/`std` lints

tarcieri:
Enables the following clippy lints:

- `clippy::alloc_instead_of_core`
- `clippy::std_instead_of_alloc`
- `clippy::std_instead_of_core`

These check that imports are coming from the correct package in the standard library, enabling more features to be used in `no_std` environments.


tarcieri:
cc @juliankrieger

PR #1649, opened 2025-02-07
Add support for 64-bit moduli

dvdplm:
The `tiny-curve` crate uses 64-bit [`FieldBytesSize`](https://github.com/fjarri/tiny-curve/blob/master/src/curve64.rs#L50), and when used in combination with the [`FromEncodedPoint`](https://docs.rs/elliptic-curve/0.13.8/elliptic_curve/sec1/trait.FromEncodedPoint.html) trait, there's a problem with a missing impl of the `ModulusSize` trait. 

Adding `U8` to the list of impl'd types should fix this.

### DIFF
--- a/sec1/src/point.rs
+++ b/sec1/src/point.rs
@@ -15,7 +15,7 @@ use core::{
     str,
 };
 use hybrid_array::{
-    typenum::{U1, U24, U28, U32, U48, U66},
+    typenum::{U1, U24, U28, U32, U48, U66, U8},
     Array, ArraySize,
 };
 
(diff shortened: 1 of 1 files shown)


tarcieri:
Ugh, not super wild about this. Generally the whole `elliptic-curve` stack has a minimum curve size of 192-bits by design.

fjarri:
> Ugh, not super wild about this. Generally the whole elliptic-curve stack has a minimum field size of 192-bits by design.

Is this kind of hand-holding really needed? It's still pretty easy to make an unsafe curve even with 192-bit elements.

fjarri:
Also, in case this misuse protection was not the primary purpose of `ModulusSize`, a viable alternative is to remove the bound on `ArrayLength` and instead add an associated type:
```rust
pub trait Sec1Sizes { // or whatever name
    type ModulusSize: ArrayLength<u8>;
    type CompressedPointSize: ArrayLength<u8>;
    ...
}
```
I don't think it will restrict any current upstream usage, but it will allow one to define their own `ModulusSize` if needed. Currently it's not possible since `ArrayLength` implies `typenum::Unsigned`, which is sealed.

tarcieri:
@fjarri I'd be fine with some way of letting crates define their own downstream SEC1 size combinations if it can be made to work with the `elliptic-curve` crate

fjarri:
Seems that `Sec1Sizes` approach will cause cause quite a bit of changes, so as a minimum-change option I made an experimental PR that just implements `ModulusSize` for all `typenum` types: https://github.com/RustCrypto/formats/pull/1650.

tarcieri:
Let's go with #1650

PR #1650, opened 2025-02-08
sec1: blanket impl ModulusSize for all typenums

fjarri:
Instead of `impl_modulus_size!` for a number of types, impl `ModulusSize` for all types. This is useful for testing, see https://github.com/RustCrypto/formats/pull/1649

Note: the bound `UntaggedPointSize: Add<U1, Output = Self::UncompressedPointSize>` caused issues because there's another bound with `Output = Self::UncompressedPointSize`, and I don't think Rust's type system allows one to assert that two associated types are equal.

Removing that bound did not cause compilation issues, but it left the tag size unconstrained. So instead I added the `Add<U1, Output = Self::CompressedPointSize>` bound on `ModulusSize` itself.

### DIFF
--- a/sec1/src/point.rs
+++ b/sec1/src/point.rs
@@ -14,10 +14,7 @@ use core::{
     ops::{Add, Sub},
     str,
 };
-use hybrid_array::{
-    typenum::{U1, U24, U28, U32, U48, U66},
-    Array, ArraySize,
-};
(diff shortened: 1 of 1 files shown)


tarcieri (approved):
This seems like it should be fine, although it needs to be integration tested against https://github.com/RustCrypto/elliptic-curves to know for sure

PR #1652, opened 2025-02-09
Helper function for BER to DER transcoding

dwhjames:
Issue #779 has had some progress (PR #1321). However, this helper function is intended to provide an escape hatch for limited support of some cases of BER, while not waiting for general support or needing API changes.

Specifically, the transcoding of occurrences of the constructed, indefinite-length method into the constructed, definite-length method. This is likely sufficient to address the examples in the wild, reported in Issue #779 and elsewhere.

The goal is not to support all possible violations of DER that are still following valid BER. Examples are non-canonical encodings of lengths (which would require an alternative `Length`), or constructed string types (which could be handled by further work on this function, but usage may not be well motivated).


dwhjames:
I originally developed this as a recursive function, in fact, a set of co-recursive functions. To preserve stack safety for arbitrary depth encodings, I rewrote it as a loop + stack; however, this is certainly far harder code to read.

I used the recursive approach as a reference implementation to test the iterative implementation. Although I haven’t included it within the PR, so as to keep the PR size more reasonable. As the unit and integration tests give effectively 100% code coverage (except for some unreachable error cases), I’m hoping that is sufficient assurance, even without the more comprehensible reference implementation.

tarcieri:
@dwhjames sorry for not commenting on this. it's a potentially interesting stopgap, although it would be a lot better to actually support indefinite lengths

dwhjames:
Closing because
- https://github.com/RustCrypto/formats/issues/779#issuecomment-2902975628

dwhjames:
> although it would be a lot better to actually support indefinite lengths

definitely ;)

Issue #1539, opened 2024-10-01
der: Use #[asn1(type = "OCTET STRING")] with Vec<u8>


dzil123:
I'm trying to use the `der` and `der_derive` crates to implement encoding and encoding a struct containing a `Vec<u8>` as an `OCTET STRING`. I need my struct to be owned, not borrowed from the input DER bytes.

Using `OctetString` in the struct definition is inconvenient because I must use the fallible `new()` method when manually creating an instance of the struct. I would rather the struct use `Vec<u8>` directly, and keep the der-specific types and encoding error in the encode and decode steps.

I thought `#[asn1(type = "OCTET STRING")]` would implement this, as the documentation states:
> performs an intermediate conversion to `der::asn1::OctetString`

This does not work. Contrary to the documentation, the macro performs an intermediate conversion to `OctetStringRef`, which does not have a `TryFrom` and `Into` implemented for `Vec<u8>`.
```rust
use der::Sequence;

#[derive(Sequence)]
struct Foo {
    #[asn1(type = "OCTET STRING")]
    bar: Vec<u8>,
}
```
```
error[E0277]: the trait bound `Vec<u8>: TryFrom<OctetStringRef<'_>>` is not satisfied
 --> src/lib.rs:3:10
  |
3 | #[derive(Sequence)]
  |          ^^^^^^^^ the trait `From<OctetStringRef<'_>>` is not implemented for `Vec<u8>`, which is required by `OctetStringRef<'_>: TryInto<_>`
  |
  = help: the following other types implement trait `From<T>`:
            <Vec<u8> as From<&str>>
            <Vec<u8> as From<CString>>
            <Vec<u8> as From<String>>
  = note: required for `OctetStringRef<'_>` to implement `Into<Vec<u8>>`
  = note: required for `Vec<u8>` to implement `TryFrom<OctetStringRef<'_>>`
  = note: required for `OctetStringRef<'_>` to implement `TryInto<Vec<u8>>`
  = note: this error originates in the derive macro `Sequence` (in Nightly builds, run with -Z macro-backtrace for more info)

error[E0308]: mismatched types
  --> src/lib.rs:3:10
   |
3  | #[derive(Sequence)]
   |          ^^^^^^^^
   |          |
   |          expected `&[u8]`, found `Vec<u8>`
   |          arguments to this function are incorrect
   |
   = note: expected reference `&[u8]`
                 found struct `Vec<u8>`
note: associated function defined here
[...]

PR #1680, opened 2025-02-25
der: fix derive(Sequence) on field: Vec<u8>

dishmaker:
Fixes: 
- #1539

Adds attribute:
- `deref = "true"`

```rust
#[derive(Sequence)]
pub struct MySequence {
    /// Without deref = "true" macro generates an error:
    ///
    /// the trait `From<Vec<u8>>` is not implemented for `BitStringRef<'_>`
    #[asn1(type = "OCTET STRING", deref = "true")]
    pub owned_bytes: Vec<u8>,
}
```

Is there any better way to add `&` ?

### DIFF
--- a/der/src/asn1/bit_string.rs
+++ b/der/src/asn1/bit_string.rs
@@ -347,6 +347,36 @@ mod allocating {
         }
     }
 
+    // /// Hack for simplifying the custom derive use case.
+    // impl<'a> TryFrom<Vec<u8>> for BitStringRef<'a> {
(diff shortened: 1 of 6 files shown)


baloo (approved):

baloo:
I wish there was a better way to achieve that, like a `impl<'a, S> TryFrom<&'a S> for BitStringRef<'a> where S:AsRef<[u8]>` but that conflicts with other blanket implementations because we also need the From<BitString>

PR #1692, opened 2025-03-03
const-oid: Add OIDs from rfc9688.

flihp:
This includes OIDs for the use of sha3 digests in various contexts.


tarcieri (approved):

baloo:
For what it's worth, you should already have oid support in sha3.

```
[dependencies]
sha3 = { features = ["oid"] }
```

They are locally defined in the sha3 crate and should be attached to the signer (if not it's a bug).

PR #1694, opened 2025-03-05
base64ct: update changelog entry for v1.7.0 with yanked tag

newpavlov:


newpavlov:
I am fine with either.

newpavlov:
Opened #1695 to discuss what to do with v1.7.0.

newpavlov:
I will merge it for now. It can be reverted later in the case of v1.7.0 un-yanking.

PR #1702, opened 2025-03-10
base64ct: downgrade edition to 2021; MSRV 1.81

### LINKED ISSUE #1695: base64ct: un-yank v1.7.0 or publish v2.0.0?
In the v1.7.0 release we have bumped MSRV to 1.85. Technically, it was done according to our policy.

But, personally, I considered MSRV bumps breaking changes and it was one of the reasons why I did not want to release v1-s until MSRV-aware lands and we update MSRV of our crates accordingly. In other words, for pre-v1 crates the claimed MSRV policy means that MSRV bumps are considered breaking.

The v1.7.0 release has caused issues for downstream users (see #1684 for more information), which led to the yanking. But we can argue that the fault lies with the users who did not use the `--locked` flag while installing binaries or did not commit Cargo.lock to their repositories.

We have two options: un-yank v1.7.0 or publish v2.0.0. Personally, I am fine with either with a very slight leaning towards the latter option.

tarcieri:
We had some complaints about the steep MSRV upgrade from 1.60 to 1.85 which occurred in the 1.7.0 release (#1695).

This lowers the MSRV to just what we need for `core::error::Error` support, downgrading the edition in the process.

We can bump edition in a future release, once 1.85 has been out longer.

Closes #1695


tarcieri:
Aah that's fun, need to move it out of the workspace too

PR #1710, opened 2025-03-12
Make pem a optional feature in pkcs12 crate

balderk:


pkcs12/Cargo.toml
```diff
@@ -35,6 +35,7 @@ whirlpool = "=0.11.0-pre.4"
 
 [features]
 kdf = ["dep:digest", "dep:zeroize"]
```
baloo, line 37:
```suggestion
default = ["pem"]
kdf = ["dep:digest", "dep:zeroize"]
```

PR #1713, opened 2025-03-13
der: add const ::new to Length, BytesRef, AnyRef

### LINKED ISSUE #1706: Make AnyRef::new() const
The use of AnyRef is very likely to take a reference to a const slice. Makeing `AnyRef::new()` const would mean less runtime code.

dishmaker:
Fixes #1706


tarcieri:
@dishmaker this looks generally fine... want to remove draft?

PR #1714, opened 2025-03-13
Revert "base64ct: reject zero-length decode requests (#1387)"

tarcieri:
This reverts commit a375cbf5bd0abd5159d4b3ec203d8b0a9166d1d6 from #1387.

This was a breaking change which is causing issues in the `der` crate.

See #1711


baloo:
Ha thanks for that!

baloo:
I think we could have retained the bugfix @mpalmer  intended to fix and also not introduce the regression. I'll send a followup

PR #1718, opened 2025-03-14
der: const GeneralizedTime::from_date_time

### LINKED ISSUE #1704: Need a const way to create GeneralizedTime/DateTime
This could probably be done in a const way to reduce runtime instructions. This is particularly true for embedded contexts where infinite certs will be generated due to the lack of a clock.

dishmaker:
Fixes #1704

### DIFF
--- a/der/src/datetime.rs
+++ b/der/src/datetime.rs
@@ -11,6 +11,7 @@ use core::{fmt, str::FromStr, time::Duration};
 #[cfg(feature = "std")]
 use std::time::{SystemTime, UNIX_EPOCH};
 
+use const_range::const_contains_u8;
 #[cfg(feature = "time")]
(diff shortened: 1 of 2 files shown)


der/src/datetime.rs
```diff
@@ -11,6 +11,7 @@ use core::{fmt, str::FromStr, time::Duration};
 #[cfg(feature = "std")]
 use std::time::{SystemTime, UNIX_EPOCH};
 
+use const_range::const_contains_u8;
```
baloo, line 14: I'm confused, where does this come from?

baloo: nevermind it's below

baloo (approved):

baloo:
Thanks!

Issue #1705, opened 2025-03-11
Create separate derives for SequenceEncode/SequenceDecode


npmccallum:
Currently, we can `derive(Sequence)`. But this implies ownership of values. Where `*Ref` types are used, `derive(Sequence)` cannot be. Perhaps we could split Sequence into SequenceEncode/SequenceDecode? I'm open to other solutions.

npmccallum:
Something would also need to derive `FixedTag`. I think that's why `Sequence` was done as a single derive.

tarcieri:
`FixedTag` impls are trivial to write by hand

baloo:
Should be fixed by #1722 (if not my bad, please re-open)

PR #1722, opened 2025-03-15
der_derive: add DecodeValue, EncodeValue macros

dishmaker:
Fixes
- #1705

Adds DecodeValue macro:
```rust
#[derive(DecodeValue)]
struct MyDecodeOnly {
    field: u8,
}
impl FixedTag for MyDecodeOnly {
    const TAG: Tag = Tag::Sequence;
}
```

and EncodeValue:
```rust
#[derive(EncodeValue)]
struct MyEncodeOnly {
    field: u8,
}
impl FixedTag for MyEncodeOnly {
    const TAG: Tag = Tag::Sequence;
}

```

### DIFF
--- a/der/src/lib.rs
+++ b/der/src/lib.rs
@@ -385,9 +385,7 @@ pub use crate::{
 pub use crate::{asn1::Any, document::Document};
 
 #[cfg(feature = "derive")]
-pub use der_derive::{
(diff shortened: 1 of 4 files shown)


tarcieri:
There aren't corresponding traits for these (and I don't think there should be).

Instead I would suggest making it possible to derive `DecodeValue` and `EncodeValue`

baloo (approved):

baloo:
Thanks a lot!

PR #1724, opened 2025-03-15
der: remove TagNumber::N0..N30 consts

dishmaker:
Also fixes der_derive for 32-bit tags.


der_derive/src/tag.rs
```diff
@@ -117,62 +117,23 @@ impl Display for TagMode {
...
+        quote!(::der::TagNumber(#num))
     }
 }
 
 impl FromStr for TagNumber {
     type Err = ParseError;
 
     fn from_str(s: &str) -> Result<Self, ParseError> {
-        let n = s.parse::<u8>().map_err(|_| ParseError)?;
+        let n = s.parse::<u32>().map_err(|_| ParseError)?;
```
tarcieri, line 134: This could just be:

```suggestion
        s.parse::<u32>().map(Self).map_err(|_| ParseError)
```

der_derive/src/sequence/field.rs
```diff
@@ -348,11 +348,11 @@ mod tests {
             quote! {
                 let implicit_field = ::der::asn1::ContextSpecific::<>::decode_implicit(
                         reader,
-                        ::der::TagNumber::N0
+                        ::der::TagNumber(0u32)
```
tarcieri, line 351: Shouldn't these be...

```suggestion
                        ::der::TagNumber::new(0u32)
```

dishmaker: We can't use fn in match

```
error[E0164]: expected tuple struct or tuple variant, found associated function `::der::TagNumber::new`
   --> der/tests/derive.rs:121:18
    |
121 |         #[derive(Choice, Debug, Eq, PartialEq)]
    |                  ^^^^^^ `fn` calls are not allowed in patterns
```

tarcieri:
Aah, if `TagNumber` is just a `pub u32` newtype (didn't notice that changed yet), it seems like `new` isn't even necessary?

dishmaker:
Let's leave `::new` everywhere except `der_derive`.
We can pretend it's not `pub`, just like [pub Pin.__pointer](https://doc.rust-lang.org/src/core/pin.rs.html#1103)

tarcieri:
I'm not sure it makes sense to retain when it enforces no invariant. However, we can follow up on that separately.

Issue #1278, opened 2023-12-13
der: Provide from_der that doesn't error on trailing data


NWPlayer123:
Hello, I'm currently trying to implement parsing of a certificate blob that Panda3D uses, and it's proving rather difficult with RustCrypto's implementation.

Basically, the format doesn't store the length so I have no way of determining how long each certificate in the blob is without actually parsing it, and the default way that from_der works is it calls SliceReader::finish which returns an error about TrailingData instead of the finished certificate and all leftover data, which means I can't make a nice loop to parse through like I could with d2i_X509 (which is what the [original implementation uses](https://github.com/panda3d/panda3d/blob/master/panda/src/express/multifile.cxx#L2420)).

For now, I've implemented it by creating my own Certificate that just keeps the remaining length as part of the struct:
```rs
use x509_cert::certificate::{CertificateInner, Rfc5280};
use der::{Decode, Reader, Result, SliceReader, Length};

#[derive(Debug)]
pub struct Certificate {
    pub certificate: CertificateInner<Rfc5280>,
    pub remaining_len: Length,
}

impl<'a> Decode<'a> for Certificate {
    fn decode<R: Reader<'a>>(reader: &mut R) -> Result<Self> {
        let inner = CertificateInner::<Rfc5280>::decode(reader)?;
        Ok(Certificate { certificate: inner, remaining_len: Length::new(0) })
    }

    fn from_der(bytes: &'a [u8]) -> Result<Self> {
        let mut reader = SliceReader::new(bytes)?;
        let mut result = Self::decode(&mut reader)?;
        result.remaining_len = reader.remaining_len();
        Ok(result)
    }
}
```

dishmaker:
Not an issue.

Just use:
```rust
let mut reader = SliceReader::new(bytes)?;
while reader.remaining_len() > 0 {
    let cert = CertificateInner::<Rfc5280>::decode(&mut reader)?;
}
```

NWPlayer123:
That's basically what I ended up with but it doesn't solve that it should be a supported method
https://github.com/NWPlayer123/Orthrus/blob/main/crates/core/src/certificate.rs

dishmaker:
Looks like you don't use that `Certificate` so issue could be closed.
```rust
for _ in 0..cert_count {
    let (_, remaining) = cert::read_certificate(&cert_blob.remaining_slice()?).unwrap();

    let length = cert_blob.len()?;
    cert_blob.set_position(length - remaining as u64)?;
}
```

[...]

PR #1725, opened 2025-03-15
der: add from_der_partial on Decode trait

dishmaker:
Fixes:
- #1278

### DIFF
--- a/der/src/decode.rs
+++ b/der/src/decode.rs
@@ -34,10 +34,23 @@ pub trait Decode<'a>: Sized + 'a {
     }
 
     /// Parse `Self` from the provided DER-encoded byte slice.
+    ///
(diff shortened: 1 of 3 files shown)


der/src/decode.rs
```diff
@@ -34,11 +34,24 @@ pub trait Decode<'a>: Sized + 'a {
     }
 
     /// Parse `Self` from the provided DER-encoded byte slice.
+    ///
+    /// Returns [`ErrorKind::TrailingData`] if message is incomplete.
```
tarcieri, line 38: `ErrorKind` isn't in scope. Maybe this?

```suggestion
    /// Returns [`crate::ErrorKind::TrailingData`] if message is incomplete.
```

PR #1726, opened 2025-03-17
der: remove 256 MiB limit on `Length`

turbocool3r:
`Length` type was previously limited to 256 MiB. This PR removes the limit and changes the internal representation to `usize` instead of `u32`.

### DIFF
--- a/der/src/length.rs
+++ b/der/src/length.rs
@@ -7,9 +7,6 @@ use core::{
     ops::{Add, Sub},
 };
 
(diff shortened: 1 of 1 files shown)


tarcieri:
You seem to have started from an out-of-date copy of the repo.

This also makes significantly more changes than what's in the title of the issue. I would suggest starting by simply removing the limit.

turbocool3r:
@tarcieri sure, will change back to `u32`.

tarcieri (approved):

PR #1727, opened 2025-03-18
der: deprecate `TagNumber::new`

tarcieri:
After #1651 the inner `u32` of `TagNumber` is now `pub` and the `TagNumber::new` function no longer maintains an invariant (which it previously did by panicking when tag numbers are out-of-range).

With the field now `pub` there is no longer a need to call a function to construct a `TagNumber`, instead it can be done directly.


baloo:
you'll need to exclude changes made to `spki`, `x509-cert` and all other crates currently excluded from the workspace.
We can bring those back in when https://github.com/RustCrypto/formats/pull/1689 merges.

Issue #1730, opened 2025-03-21
der: derive(Sequence) with [u8; N] field


dishmaker:
`[u8; N]` arrays should be allowed as `OCTET STRING` fields.

```rust
#[derive(Sequence)]
pub struct MySequence {
    #[asn1(context_specific = "2", type = "OCTET STRING")]
    pub authority_reference: [u8; 8],
}
```

```
error[E0277]: the trait bound `der::asn1::OctetStringRef<'_>: TryFrom<[u8; 8]>` is not satisfied
  --> example.rs:68:10
   |
68 | #[derive(Sequence)]
   |          ^^^^^^^^ the trait `From<[u8; 8]>` is not implemented for `der::asn1::OctetStringRef<'_>`
   |
   = help: the following other types implement trait `From<T>`:
             `der::asn1::OctetStringRef<'_>` implements `From<&der::asn1::OctetString>`
             `der::asn1::OctetStringRef<'_>` implements `From<&der::asn1::OctetStringRef<'_>>`
   = note: required for `[u8; 8]` to implement `Into<der::asn1::OctetStringRef<'_>>`
   = note: required for `der::asn1::OctetStringRef<'_>` to implement `TryFrom<[u8; 8]>`
   = note: this error originates in the derive macro `Sequence`
```

does not work with `deref ="true"` too.
```
the trait `From<&[u8; 8]>` is not implemented for `OctetStringRef<'_>`
```

<details>
  <summary>[Spoiler] real use case</summary>

```rust
/// EU Tachograph certificate body
#[derive(Sequence, Debug, Clone)]
#[asn1(tag_mode = "IMPLICIT")]

pub struct TachographCertificateBodyRaw {
    /// primitive, 1 byte
    #[asn1(application = "41", type = "OCTET STRING")]
    pub profile_identifier: [u8; 1],

    /// primitive, 8 bytes
    #[asn1(application = "2", type = "OCTET STRING")]
    pub authority_reference: [u8; 8],

    /// primitive, 7 bytes
    #[asn1(application = "76", type = "OCTET STRING")]
    pub holder_authorisation: [u8; 7],

    /// constructed
    #[asn1(application = "73")]
    pub public_key: CertificatePublicKeyRaw,

    /// primitive, 8 bytes
    #[asn1(application = "32", type = "OCTET STRING")]
    pub holder_reference: [u8; 8],

    /// primitive, 4 bytes
    #[asn1(application = "37", type = "OCTET STRING")]
[...]

PR #1731, opened 2025-03-21
der: conversions between `BitStringRef`/`OctetStringRef` and `[u8; N]`

dishmaker:
Fixes: 
- #1730

```rust
#[derive(Sequence)]
pub struct MyStackAllocSequence {
    #[asn1(type = "OCTET STRING", deref = "true")]
    pub array_bytes: [u8; 2],
}
```

### DIFF
--- a/der/src/asn1/octet_string.rs
+++ b/der/src/asn1/octet_string.rs
@@ -125,13 +125,10 @@ impl<'a, const N: usize> TryFrom<OctetStringRef<'a>> for [u8; N] {
     type Error = Error;
 
     fn try_from(octet_string: OctetStringRef<'a>) -> Result<Self, Self::Error> {
-        let expected_len = Length::try_from(N)?;
-        octet_string.as_bytes().try_into().map_err(|_| {
-            Error::from_kind(ErrorKind::Incomplete {
-                expected_len,
-                actual_len: octet_string.len(),
-            })
-        })
+        octet_string
+            .as_bytes()
+            .try_into()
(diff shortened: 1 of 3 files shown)


der/src/asn1/octet_string.rs
```diff
@@ -113,6 +113,30 @@ impl<'a> TryFrom<&&'a [u8]> for OctetStringRef<'a> {
     }
 }
 
+/// Hack for simplifying the custom derive use case.
```
tarcieri, line 116: I don't think this is a "hack", it's a reasonable conversion to allow encoding/decoding byte arrays as `OCTET STRING`

der/src/asn1/octet_string.rs
```diff
@@ -113,6 +113,30 @@ impl<'a> TryFrom<&&'a [u8]> for OctetStringRef<'a> {
...
+}
+
+/// Hack for simplifying the custom derive use case.
+impl<'a, const N: usize> TryFrom<OctetStringRef<'a>> for [u8; N] {
+    type Error = Error;
+
+    fn try_from(octet_string: OctetStringRef<'a>) -> Result<Self, Self::Error> {
+        let expected_len = Length::new_usize(N)?;
+        octet_string.as_bytes().try_into().map_err(|_| {
+            Error::from_kind(ErrorKind::Incomplete {
```
tarcieri, line 132: I think `ErrorKind::Length` makes more sense here, especially if there is too much data to fit in the array

tarcieri (approved):

PR #1740, opened 2025-03-25
der: use strict context-specific skipping condition (equal tag numbers only)

### LINKED ISSUE #1723: der: Correct context-specific skipping algorithm
Looks like currently tag numbers lower than requested are skipped.

Decoder breaks only on higher than expected tag numbers:
`if !tag.is_context_specific() || tag.number() > tag_number { break }`

Commit history:
https://github.com/RustCrypto/formats/commit/448d5fb186c1819b0b63bec233094bd551a344d0?diff=unified#diff-7bb94c63666ed02733ba218c24281358f540200bd845b247f1a84a0f95021234R90

https://github.com/RustCrypto/formats/blame/d87ed96b253e1af95533234804a224cbb408099d/der/src/decoder.rs#L195

However, python `asn1tools` would not skip lower tags when decoding.

In my example the decoder just stops on tag `[1]` and does not decode `[4]`.

`foo.asn`
```asn1
Foo DEFINITIONS ::= BEGIN

    SequenceA ::= SEQUENCE {
        question  IA5String,
        context1  [1] IA5String OPTIONAL,
        context4  [4] IA5String OPTIONAL
    }

    SequenceB ::= SEQUENCE {
        question  IA5String,
        context4  [4] IA5String OPTIONAL
    }

END
```

```python
import asn1tools
foo = asn1tools.compile_files('foo.asn')
encoded = foo.encode('SequenceA', {
    'question': '0',
    'context1': '1',
    'context4': '4',
    })
print(encoded.hex(' '))

decoded = foo.decode('SequenceB', encoded)
print(decoded)
```
```shell
$ python test_foo.py 
30 0d 16 01 30 a1 03 16 01 31 a4 03 16 01 34
{'question': '0'}
```

And btw it works when `SequenceA` is decoded as `SequenceA`:
```shell
$ python test_foo.py 
30 0d 16 01 30 a1 03 16 01 31 a4 03 16 01 34
{'question': '0', 'context1': '1', 'context4': '4'}
```
So there are 2 possible algorithms:
- `tag.number() > tag_number`
- `tag.number() != tag_number`

Which one is correct?

dishmaker:
Fixes #1723

I simplified fn `decode_with` into `peek_tag_matches`.

Depends on:
- #1783

### DIFF
--- a/der/src/asn1/context_specific.rs
+++ b/der/src/asn1/context_specific.rs
@@ -185,6 +185,8 @@ where
             // 8.14.4 If implicit tagging was used in the definition of the type, then:
             // a) the encoding shall be constructed if the base encoding is constructed, and shall be primitive otherwise; and
             // b) the contents octets shall be the same as the contents octets [EncodeValue] of the base encoding.
+            //
+            // TODO(dishmaker): use IsConstructed trait for IMPLICIT
             TagMode::Implicit => self.value.tag().is_constructed(),
         };
 
--- a/der/src/asn1/context_specific.rs
+++ b/der/src/asn1/context_specific.rs
@@ -1,7 +1,7 @@
 //! Context-specific field.
 
 use crate::{
-    Choice, Decode, DecodeValue, DerOrd, Encode, EncodeValue, EncodeValueRef, Error, Header,
+    Choice, Class, Decode, DecodeValue, DerOrd, Encode, EncodeValue, EncodeValueRef, Error, Header,
     Length, Reader, Tag, TagMode, TagNumber, Tagged, ValueOrd, Writer, asn1::AnyRef,
     tag::IsConstructed,
 };
@@ -47,7 +47,10 @@ impl<T> ContextSpecific<T> {
     where
         T: Decode<'a>,
     {
-        Self::decode_with(reader, tag_number, |reader| Self::decode(reader))
+        if !peek_tag_matches(reader, Class::ContextSpecific, tag_number)? {
+            return Ok(None);
+        }
+        Ok(Some(Self::decode(reader)?))
     }
 
     /// Attempt to decode an `IMPLICIT` ASN.1 `CONTEXT-SPECIFIC` field with the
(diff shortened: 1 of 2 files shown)


dishmaker:
Looks like I broke something in `pkcs8`

```
     Running tests/builder.rs (target/debug/deps/builder-c42694f58bf84285)

running 9 tests
test dynamic_signer ... ok
test certificate_request_attributes ... ok
test certificate_request ... ok
test async_builder ... ok
test leaf_certificate has been running for over 60 seconds
test pss_certificate has been running for over 60 seconds
test root_ca_certificate has been running for over 60 seconds
test root_ca_certificate_ecdsa has been running for over 60 seconds
test sub_ca_certificate has been running for over 60 seconds
```

dishmaker:
Weird
image

pkcs8/src/private_key_info.rs
```diff
@@ -199,6 +202,9 @@ where
             let version = Version::decode(reader)?;
             let algorithm = reader.decode()?;
             let private_key = Key::decode(reader)?;
+            let _attributes =
+                reader.context_specific::<SequenceRef<'_>>(ATTRIBUTES_TAG, TagMode::Implicit)?;
```
dishmaker, line 206: Is this ok?

tarcieri: That's an `OPTIONAL` field

tarcieri:
What was causing the infinite loop? That seems bad

dishmaker:
zlint - htop screenshot above

pkcs8/src/private_key_info.rs
dishmaker: `context_specific()` returns `Option<T>`
https://github.com/RustCrypto/formats/blob/53d2e3b34c38920d09c266896688ee006edd3571/der/src/reader.rs#L49-L53

tarcieri: Yes that should be fine then

dishmaker:
For some reason today it doesn't get stuck: `cargo test --all --all-features`

```
   Running tests/builder.rs (target/debug/deps/builder-c42694f58bf84285)

running 9 tests
test dynamic_signer ... ok
test certificate_request_attributes ... ok
test certificate_request ... ok
test async_builder ... ok
test root_ca_certificate_ecdsa ... ok
test sub_ca_certificate ... ok
test leaf_certificate ... ok
test root_ca_certificate ... ok
test pss_certificate ... ok
```

tarcieri:
@dishmaker can you rebase?

dishmaker:
When this merges, I can add `Private<T>` and `Application<T>` as-is.

Similar to https://github.com/RustCrypto/formats/commit/1f43a08b401bf9db7c81516c974570b17a9dfde2#diff-360e6ab129e770a109e6cf7f57b6340d480bb2fd17a7f75952ef65aab6a442eb

tarcieri (approved):

Issue #1746, opened 2025-03-28
der: drop `Encode`, `Decode`, `Tagged` traits requirements for `IMPLICIT`


dishmaker:
The problem is that `Encode` and `Decode` traits are needed for everything encoded with `IMPLICIT` wrapper.

Related:
- #1745

Issue #1745, opened 2025-03-28
der: split `IMPLICIT` and `EXPLICIT` ContextSpecific wrapper


dishmaker:
`IMPLICIT` wrapper should need only `EncodeValue + DecodeValue + IsConstructed`.

Currently it needs all of the traits.

PR #1747, opened 2025-03-28
der: add `AnyCustomClassExplicit`/`AnyCustomClassImplicit`

dishmaker:
Potential fix for:
- #1745 
- #1746

Not the best solution, but a proof-of-concept of internal API that `ContextSpecific` wrapper would use.

### DIFF
--- a/der/src/asn1/any_custom_class.rs
+++ b/der/src/asn1/any_custom_class.rs
@@ -1,0 +1,252 @@
+//! Less strict Context-specific, Application or Private field.
+
+use crate::{
+    Class, Decode, DecodeValue, Encode, EncodeValue, Error, Header, Length, Reader, Tag, TagNumber,
+    Tagged, Writer,
+};
+
+use super::AnyRef;
+
+/// `APPLICATION`, `CONTEXT-SPECIFIC` or `PRIVATE` tagged value.
+///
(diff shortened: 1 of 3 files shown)


der/src/asn1/any_custom_class.rs
```diff
@@ -0,0 +1,252 @@
...
+            .custom_class(self.class, true)
+            .expect("class not to be Universal")
+    }
+}
+
+impl<T> Tagged for AnyCustomClassImplicit<T> {
+    fn tag(&self) -> Tag {
+        self.tag_number
+            .custom_class(self.class, self.constructed)
+            .expect("class not to be Universal")
```
dishmaker, line 209: I don't know how to get rid of this `unwrap`/`expect`

der/src/asn1/any_custom_class.rs
```diff
@@ -0,0 +1,252 @@
...
+        } else {
+            let expected = tag_number.custom_class(class, true);
+            Err(any_explicit.tag().unexpected_error(expected).into())
+        }
+    }
+}
+
+impl<'a, T> AnyCustomClassImplicit<T>
+where
+    T: Tagged + DecodeValue<'a> + 'a,
```
dishmaker, line 87: TODO replace with `IsConstructed` from another PR

PR #1765, opened 2025-04-16
Backport #1760

daxpedda:
This is a request to backport #1760 to `der` v0.7 and a consequent point release.

Context: some parts of PKCS#11 require supplying ASN.1 encoded values. I spent quite some time until I figured out something is wrong with the encoding :sweat_smile:.


baloo:
I will release tomorrow

PR #1776, opened 2025-04-22
x509-cert: fix test: stuck zlint stdout reader

### LINKED ISSUE #1775: x509-cert: test sometimes freezes when reading zlint stdout
```
     Running tests/builder.rs (target/debug/deps/builder-c42694f58bf84285)

running 9 tests
test dynamic_signer ... ok
test certificate_request_attributes ... ok
test certificate_request ... ok
test async_builder ... ok
test leaf_certificate has been running for over 60 seconds
test pss_certificate has been running for over 60 seconds
test root_ca_certificate has been running for over 60 seconds
test root_ca_certificate_ecdsa has been running for over 60 seconds
test sub_ca_certificate has been running for over 60 seconds
```

Image

dishmaker:
Fixes #1775

### DIFF
--- a/x509-cert/test-support/src/zlint.rs
+++ b/x509-cert/test-support/src/zlint.rs
@@ -8,6 +8,7 @@ use std::{
     fs::File,
     io::{self, Read, Write},
     process::{Command, Stdio},
+    thread,
 };
 use tempfile::tempdir;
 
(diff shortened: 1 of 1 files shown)


x509-cert/test-support/src/zlint.rs
```diff
@@ -167,17 +171,26 @@ pub fn check_certificate(pem: &[u8], ignored: &[&str]) {
...
-        .read_to_end(&mut output_buf)
-        .expect("read zlint output");
 
+    let output_buf = output_handle.join().expect("stdout thread to join");
     let output: LintResult = serde_json::from_slice(&output_buf).expect("parse zlint output");
 
     assert!(output.check_lints(ignored));
+
+    std::fs::remove_file(config_path).expect("tmp file to be removed");
+    std::fs::remove_file(cert_path).expect("tmp file to be removed");
```
baloo, line 193: That shouldn't be necessary, when `tmpdir` gets out of scope the files inside should be collected.

I guess there could have been behavior on your platform where the open FDs to config or certificate could have stop collection, but this should be solved by the scoping you added above.

x509-cert/test-support/src/zlint.rs
```diff
@@ -167,17 +171,26 @@ pub fn check_certificate(pem: &[u8], ignored: &[&str]) {
         });
 
     let mut stdout = child.stdout.take().unwrap();
+
+    // read stdout in another thread
```
baloo, line 175: It should not be necessary, moving the `child.wait` down should be enough. There is no need to start a new thread.

Read the output until the end, then wait. The wait was backward, I guess I never noticed because the buffer was enough on my setup, or the output grew and I (and CI) were using an older revision of zlint.

baloo (approved):

PR #1778, opened 2025-04-22
wip: add `der::Cow` enum

dishmaker:


dishmaker:
I have a skill issue
image

baloo:
I don't know how I feel about this one. I think I would rather duplicate the types to add a `*Ref<'a>` rather than rely on `Cow`.

(not a suggestion, or a request to change this PR at this moment, just a vibe)

tarcieri:
I'd agree this probably doesn't make sense, and would suggest potentially investigating trying to impl the relevant traits to use `alloc::borrow::Cow` (though I'm not sure `&'a TypeRef<'a>` would actually work for e.g. `Borrow`?)

PR #1779, opened 2025-04-22
gss-api: change `NegHints::hint_name` to GeneralStringRef type

dishmaker:
Better than using `AnyRef` and `.value()[2..]`

### DIFF
--- a/der/src/asn1/general_string.rs
+++ b/der/src/asn1/general_string.rs
@@ -1,0 +1,32 @@
+use crate::{DecodeValue, EncodeValue, FixedTag, Header, Length, Reader, Tag, Writer};
+
+use super::OctetStringRef;
+
+/// This is currently `OctetStringRef` as `GeneralString` is not part of the `der` crate
+#[derive(Clone, Copy, Debug, Eq, PartialEq)]
+pub struct GeneralStringRef<'a> {
+    /// Raw contents, unchecked
+    #[doc(hidden)]
+    pub __contents: OctetStringRef<'a>,
+}
+impl FixedTag for GeneralStringRef<'_> {
+    const TAG: Tag = Tag::GeneralString;
+}
+impl<'a> DecodeValue<'a> for GeneralStringRef<'a> {
+    type Error = crate::Error;
+
+    fn decode_value<R: Reader<'a>>(reader: &mut R, header: Header) -> Result<Self, Self::Error> {
+        Ok(Self {
+            __contents: OctetStringRef::decode_value(reader, header)?,
+        })
+    }
+}
+impl EncodeValue for GeneralStringRef<'_> {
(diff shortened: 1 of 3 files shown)


gss-api/src/negotiation.rs
```diff
@@ -295,14 +295,8 @@ pub enum NegState {
...
-    /// This is currently `AnyRef` as `GeneralString` is not part of the `der` crate
-    #[asn1(
-        context_specific = "0",
-        optional = "true",
-        tag_mode = "IMPLICIT",
-        constructed = "true"
-    )]
-    pub hint_name: Option<AnyRef<'a>>, // TODO: GeneralString
+    #[asn1(context_specific = "0", optional = "true")]
+    pub hint_name: Option<GeneralStringRef<'a>>, // TODO: der GeneralString
```
baloo, line 299:
```suggestion
    pub hint_name: Option<GeneralStringRef<'a>>,
```

tarcieri:
If you made this type opaque instead of exposing its `contents`, we could probably add something like `GeneralStringRef` directly to `der`, even if it provided little or no functionality (at least to start with)

der/src/asn1/general_string.rs
```diff
@@ -0,0 +1,32 @@
+use crate::{DecodeValue, EncodeValue, FixedTag, Header, Length, Reader, Tag, Writer};
+
+use super::OctetStringRef;
+
+/// This is currently `OctetStringRef` as `GeneralString` is not part of the `der` crate
+#[derive(Clone, Copy, Debug, Eq, PartialEq)]
+pub struct GeneralStringRef<'a> {
+    /// Raw contents, unchecked
+    #[doc(hidden)]
+    pub __contents: OctetStringRef<'a>,
```
tarcieri, line 10: Why is this pub? Also seems like it could just be a byte slice

dishmaker: It's used in `gss-api` test

tarcieri: Perhaps add a public accessor method for its contents?

der/src/asn1/general_string.rs
```diff
@@ -0,0 +1,30 @@
+use crate::{BytesRef, DecodeValue, EncodeValue, FixedTag, Header, Length, Reader, Tag, Writer};
+
+/// This is currently `OctetStringRef` internally, as `GeneralString` is not fully implemented yet
+#[derive(Clone, Copy, Debug, Eq, PartialEq)]
+pub struct GeneralStringRef<'a> {
+    /// Raw contents, unchecked
+    #[doc(hidden)]
+    pub __contents: &'a [u8],
```
tarcieri, line 8: The other ASN.1 string types use `BytesRef` for this, which has a precomputed `Length`, FWIW

tarcieri (approved):

PR #1788, opened 2025-05-01
wip: base64ct: test issue 1609

dishmaker:
Can't reproduce it.

Demo of issue:
- #1609


tarcieri:
We could potentially turn this into some kind of regression test, although I think that would make more sense if we could've reproduced the issue in the first place

PR #1797, opened 2025-05-03
der: const `Any::to_ref`, `BytesOwned::to_ref`

dishmaker:
Removes `expect`/`unwrap` in `AnyRef::from`

### DIFF
--- a/der/src/asn1/any.rs
+++ b/der/src/asn1/any.rs
@@ -47,6 +47,15 @@ impl<'a> AnyRef<'a> {
         }
     }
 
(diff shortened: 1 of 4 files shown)


der/src/asn1/any.rs
```diff
@@ -47,6 +47,15 @@ impl<'a> AnyRef<'a> {
         }
     }
 
+    /// Create a new [`AnyRef`] from the provided [`Any`] owned tag and bytes.
+    #[cfg(feature = "alloc")]
+    pub const fn from_owned(owned: &'a Any) -> Self {
```
tarcieri, line 52: Perhaps put this on `Any` and call it `to_ref`?

PR #1820, opened 2025-05-06
der: add indefinite length

dishmaker:
We can use `u32::MAX` as indefinite variant


tarcieri:
I've been trying to work on indefinite lengths but in a way that avoids the need to have dummy lengths or an `IndefiniteLength` type

dishmaker:
I tried a trick with `NonZero<u32>` but really what we need is `NonMax<u32>`

tarcieri:
@dishmaker that enum looks a lot like `IndefiniteLength` (which I just removed in #1807) though we still need to support definite lengths of zero

dishmaker:
Yeah, so we need `NonMax<u32>` and change `Length` to a 4-byte enum (to allow compiler optimizations)

tarcieri:
@dishmaker I definitely don't want to pull in an extra dependency for that. `der` is otherwise dependency-free.

The newtype can enforce its own maximum like it did before #1726

cms/tests/tests_from_pkcs7_crate.rs
```diff
@@ -141,3 +141,9 @@ fn cms_decode_signed_der() {
     // should match the original
     assert_eq!(reencoded_der_signed_data_in_ci, der_signed_data_in_ci)
 }
+
+#[test]
+fn cms_decode_ber() {
+    let ber_pkcs7 = include_bytes!("../tests/examples/ber_pkcs7.bin");
```
tarcieri, line 147: There's already a checked-in example: `cms_ber.bin`, which is what `cms_der.bin` is derived from

PR #1821, opened 2025-05-06
der: BER decoding support: add constructed octet string

dishmaker:
Inspired by:
- #810

### DIFF
--- a/der/src/tag.rs
+++ b/der/src/tag.rs
@@ -436,10 +436,6 @@ impl Encode for Tag {
 
         let number = self.number().value();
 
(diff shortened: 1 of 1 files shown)


der/src/tag.rs
```diff
@@ -433,6 +440,10 @@ impl Encode for Tag {
 
         let number = self.number().value();
 
+        // if writer.encoding_rules() == EncodingRules::Der {
```
tarcieri, line 443: I would really prefer not to add support for encoding non-DER-conforming BER productions to this library. I'm reluctant to even add BER support but it's necessary to parse formats we support

Issue #1603, opened 2024-11-03
Support parsing X509 certificates with no allocations


DemiMarie:
WebPKI supports this, as does my own barebones-x509 crate.

tarcieri:
See ample related discussion in #689 for the encoding side of things.

We pretty much went in the opposite direction in #765 and leaned further into the existing hard dependency on `alloc`.

tarcieri:
In some other crates, notably `spki` and the latest prereleases of `pkcs8`, we do abstract over the owned and borrowed forms using generics:

https://docs.rs/pkcs8/0.11.0-rc.1/pkcs8/struct.PrivateKeyInfo.html

This has the following forms, one of which has a lifetime and the other does not:

- https://docs.rs/pkcs8/0.11.0-rc.1/pkcs8/type.PrivateKeyInfoRef.html
- https://docs.rs/pkcs8/0.11.0-rc.1/pkcs8/type.PrivateKeyInfoOwned.html

We could potentially do something like that with `x509-cert` as well, though we already use generics to carry a profile ZST, so it might become a lot of generic parameters, which makes the API more cumbersome.

tarcieri:
Curious what that benchmark actually looks like. Other implementations may be deferring processing of things `x509-cert` processes eagerly.

I'd also be curious what the hotspots in `x509-cert`.

dishmaker:
Benchmark code:
https://github.com/librasn/rasn/blob/0a65d9a5699037ec19a706ce1f776ecc87d94b90/benches/criterion.rs#L42-L46

pyca `cryptography-x509`:
https://github.com/pyca/cryptography/blob/d20fcd0b439c6140772413aae589e3c617b39ac1/src/rust/cryptography-x509/src/certificate.rs#L40

xyzzyz:
I saw other discussion on this topic, and I understand that the current design is crucial to making builder implementation simple and consistent. 

What do you think about having a separate `CertificateRef<'a>` type, though, that only support parsing from existing DER bytes, and *does not* support building using a builder (or, really, any other way of creating instances)? On embedded devices, where this `no_std, no_alloc` scenario would be most useful, one rarely needs to create new certificates anyway, typically one only needs to verify a cert chain against trusted CA, and all of those are obtained by parsing DERs.

[...]

PR #1826, opened 2025-05-09
x509-cert: draft: `TbsCertificateInner` generic over `DerMemory` trait

dishmaker:
Main idea: have a trait that describes a few types in owned/`*Ref` form.
```rust
pub trait DerMemory {
    type BITSTRING<'m>: FixedTag
        + DecodeValue<'m, Error = der::Error>
        + EncodeValue
        + DerOrd
        + Clone
        + Debug
        + Eq
        + PartialEq
        + AsBitStringRef<'m>;
    // ... and more
}
```

Im stuck on `invariant` lifetime.

```
lifetime may not live long enough
requirement occurs because of the type `CertificateInner<'_, AllocDerMemory, P>`, which makes the generic argument `'_` invariant
the struct `CertificateInner<'m, MEM, P>` is invariant over the parameter `'m`
see <https://doc.rust-lang.org/nomicon/subtyping.html> for more information about variance

anchor.rs(94, 6): lifetime `'__der_lifetime` defined here
```

Adding `der::Cow` looks easier, i think :thinking:

### DIFF
--- a/der/src/asn1.rs
+++ b/der/src/asn1.rs
@@ -36,7 +36,7 @@ mod videotex_string;
 pub use self::{
     any::AnyRef,
     application::{Application, ApplicationRef},
-    bit_string::{BitStringIter, BitStringRef},
+    bit_string::{AsBitStringRef, BitStringIter, BitStringRef},
     choice::Choice,
     context_specific::{ContextSpecific, ContextSpecificRef},
     general_string::GeneralStringRef,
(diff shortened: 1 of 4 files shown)


tarcieri:
I'm not sure what you're trying to accomplish with this

dishmaker:
I want to make `Ref` and `Owned` types from `TbsCertificateInner`.

https://github.com/RustCrypto/formats/blob/907a76a817f674ae9c61114951264aa05060d085/pkcs8/src/private_key_info.rs#L373-L378

tarcieri:
It would probably be good to flesh out a design in #1603.

Note that elsewhere we have avoided adding lifetimes except for the reference types, which can be done with generic parameters (as opposed to what you're trying to do with `'static` in this PR)

dishmaker:
Closing, because Tony merged #1921  :smile:

PR #1830, opened 2025-05-10
der: fix panic in `value_cmp`: add `Iterator::size_hint` + tests

dishmaker:
The test was failing because:

```
 failures:

---- asn1::set_of::tests::setof_valueord stdout ----

thread 'asn1::set_of::tests::setof_valueord' panicked at core/src/iter/traits/exact_size.rs:122:9:
assertion `left == right` failed
  left: None
 right: Some(0)
```

### DIFF
--- a/der/src/asn1/bit_string.rs
+++ b/der/src/asn1/bit_string.rs
@@ -573,6 +573,8 @@ where
 #[cfg(test)]
 #[allow(clippy::unwrap_used)]
 mod tests {
+    use core::cmp::Ordering;
+
     use super::{BitStringRef, Result, Tag};
     use crate::asn1::AnyRef;
     use hex_literal::hex;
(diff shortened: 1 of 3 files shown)


der/src/asn1/bit_string.rs
```diff
@@ -472,6 +472,10 @@ impl Iterator for BitStringIter<'_> {
         self.position = self.position.checked_add(1)?;
         Some(byte & bit != 0)
     }
+    fn size_hint(&self) -> (usize, Option<usize>) {
```
baloo, line 475:
```suggestion

    fn size_hint(&self) -> (usize, Option<usize>) {
```

der/src/asn1/sequence_of.rs
```diff
@@ -125,6 +125,9 @@ impl<'a, T> Iterator for SequenceOfIter<'a, T> {
     fn next(&mut self) -> Option<&'a T> {
         self.inner.next()
     }
+    fn size_hint(&self) -> (usize, Option<usize>) {
```
baloo, line 128:
```suggestion

    fn size_hint(&self) -> (usize, Option<usize>) {
```

baloo (approved):

PR #1839, opened 2025-05-15
base64ct: bump edition to 2024 (MSRV 1.85); add back to workspace

tarcieri:
We kept `base64ct` out of the toplevel workspace because we can't run tests on it with its previous MSRV of 1.81 because that compiler can't recognize the newer edition on all of the other crates in the workspace.

We originally attempted this upgrade in #1670 but reverted it in #1702 due to complaints we received in #1695.

Now that the needed MSRV of 1.85 is `N-2` of the current 1.87 stable, we can hopefully perform this upgrade without receiving as many complaints. Several other crates have bumped edition to 2024 already, and edition upgrades are a time when MSRV bumps like this should be expected.


0x009922:
Okay, I see: https://github.com/rust-lang/api-guidelines/discussions/231

PR #1853, opened 2025-06-02
[WIP] der: use `*Ref` types based on fat pointers

tarcieri:
This is a demonstration of how the `*Ref` types can be converted into `repr(transparent)` newtypes of underlying byte slices which can impl `ToOwned` and thus work with `alloc::borrow::Cow`

cc @dishmaker


tarcieri:
Closing in favor of #1921

PR #1855, opened 2025-06-03
mcf: Modular Crypt Format implementation

tarcieri:
Initial implementation of the MCF pseudo-format as described at: https://passlib.readthedocs.io/en/stable/modular_crypt_format.html

Note that MCF is less a specification than an ad hoc set of conventions.

As MCF lacks the kind of rigidly specified sizes of its various components, a heapless `no_std` implementation isn't really possible like it is with the PHC format, and so this implementation has a hard dependency on `alloc` (though it should be possible to extract an `alloc`-free reference type which borrows from its input).

### DIFF
--- a/mcf/src/lib.rs
+++ b/mcf/src/lib.rs
@@ -1,0 +1,217 @@
+#![no_std]
+#![cfg_attr(docsrs, feature(doc_auto_cfg))]
+#![doc = include_str!("../README.md")]
+#![doc(
+    html_logo_url = "https://raw.githubusercontent.com/RustCrypto/media/6ee8e381/logo.svg",
+    html_favicon_url = "https://raw.githubusercontent.com/RustCrypto/media/6ee8e381/logo.svg"
+)]
+#![forbid(unsafe_code)]
+#![warn(
+    clippy::mod_module_files,
+    clippy::unwrap_used,
+    missing_docs,
+    unused_qualifications
+)]
+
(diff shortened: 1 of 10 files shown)


mcf/src/lib.rs
```diff
@@ -0,0 +1,199 @@
...
+        &self.0
+    }
+
+    /// Get the algorithm identifier for this MCF hash.
+    pub fn id(&self) -> &str {
+        self.fields().next().expect("ensured valid by constructor")
+    }
+
+    /// Get an iterator over the parts of the password hash as delimited by `$`, including the
+    /// initial identifier.
```
tarcieri, line 59: Perhaps it shouldn't include the initial identifier?

tarcieri: Done in 0f4d1221

carloskiki:
What is the motivation for this? The document linked says this format is deprecated since 2016 (almost 10 years ago).

tarcieri:
@carloskiki MCF is deprecated for use in new password hashes, which should use wthe PHC format (FWIW I'm the current maintainer of the PHC format spec).

However, password hashes in this format are still exceedingly common. We support them using a bespoke implementation in e.g. the [`sha-crypt`](https://github.com/RustCrypto/password-hashes/tree/master/sha-crypt) crate which could migrate to the `mcf` crate.

PR #1877, opened 2025-06-20
der: add `Reader::read_value`, auto-nest `DecodeValue`

tarcieri:
Previously `der` lacked an abstraction for automatically setting up nested readers when decoding values, so every `DecodeValue` impl ended up handling its own nesting.

This adds a new provided method to the `Reader` trait which automatically handles nesting.

This change is backwards compatible in that if a `DecodeValue` impl does its own `reader.read_nested`, it will just create another nested reader of the same length, which is redundant but doesn't break anything. However, we should remove all of the `read_nested` calls from `DecodeValue` impls with this approach anyway as they're redundant.

Using a provided method for this opens up the possibility of reader-specific handling of field decoding, which would be useful for addressing indefinite length handling for BER decoding (#779).

### DIFF
--- a/der/src/asn1/internal_macros.rs
+++ b/der/src/asn1/internal_macros.rs
@@ -151,8 +151,8 @@ macro_rules! impl_custom_class {
                     return Err(header.tag.non_canonical_error().into());
                 }
 
-                // read_nested checks if header matches decoded length
-                let value = reader.read_nested(header.length, |reader| {
+                // read_value checks if header matches decoded length
+                let value = reader.read_value(header, |reader| {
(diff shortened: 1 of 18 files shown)


tarcieri:
cc @dishmaker

tarcieri:
Note: this pretty much eliminates any direct calls to `read_nested`, although I see there are a few lingering places it's used (e.g. `ecdsa`)

dishmaker:
Looks good. Removes boilerplate :+1: 

Maybe I will add negative tests for bad nesting (i.e. when read_nested should fail).

PR #1882, opened 2025-06-21
der: add `Reader::peek`

tarcieri:
Adds a method to the `Reader` trait for spawning a sub-reader that starts at the current position but doesn't affect the parent reader's state.

Implementing indefinite lengths will involve scanning large parts of the message (possibly all the way to the end), so this is a building block to making that work.

### DIFF
--- a/der/src/reader/pem.rs
+++ b/der/src/reader/pem.rs
@@ -5,7 +5,6 @@ use crate::{EncodingRules, Error, ErrorKind, Length, Result};
 use pem_rfc7468::Decoder;
 
 /// `Reader` type which decodes PEM on-the-fly.
-#[cfg(feature = "pem")]
 #[derive(Clone)]
 pub struct PemReader<'i> {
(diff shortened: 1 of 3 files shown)


der/src/reader/pem.rs
```diff
@@ -51,9 +48,11 @@ impl<'i> Reader<'i> for PemReader<'i> {
...
 
-    fn peek_into(&self, buf: &mut [u8]) -> Result<()> {
-        self.clone().read_into(buf)?;
-        Ok(())
+    fn peek<T, F>(&self, f: F) -> Result<T>
+    where
+        F: FnOnce(&mut Self) -> Result<T>,
+    {
+        f(&mut self.clone())
     }
```
tarcieri, line 56: Really this is wrapping up cloning the `Reader`. We could potentially just bound `Reader` on `Clone` instead.

tarcieri:
Closing in favor of #1883

PR #1884, opened 2025-06-21
der: preliminary indefinite length support

tarcieri:
Adds initial support to the `Length` type for decoding indefinite lengths when `EncodingRules::Ber` are in use.

Note that this only adds the needed support to the `Length` type and handling of the constructed form of various universal tags (which may require a special `Reader` implementation for BER).


tarcieri:
cc @dishmaker

PR #1886, opened 2025-06-21
der: simplify `Header::peek` and `Tag::peek`

tarcieri:
We can rely on the ability to clone the `Reader` to use the `Decode` implementation to implement peeking, so we don't need a separate bespoke implementation.

This also avoids losing the `Reader` context while peeking.


dishmaker:
Ah, so `peek_optional` on EOF just simplifies to `reader.is_finished()`

PR #1888, opened 2025-06-21
[WIP] cms: test decoding of `cms_ber.bin`

tarcieri:
The files `cms_ber.bin` and `cms_der.bin` already checked into `cms/tests` contain test vectors showing a BER encoding with indefinite lengths and its DER equivalent.

The decoder now seems to be handling indefinite lengths, but can't handle decoding a constructed `OCTET STRING` with indefinite length (currently failing on the tag).

### DIFF
--- a/cms/tests/tests_from_pkcs7_crate.rs
+++ b/cms/tests/tests_from_pkcs7_crate.rs
@@ -141,3 +141,13 @@ fn cms_decode_signed_der() {
     // should match the original
     assert_eq!(reencoded_der_signed_data_in_ci, der_signed_data_in_ci)
 }
+
+#[test]
+fn cms_decode_signed_ber() {
+    let cms_ber = include_bytes!("../tests/examples/cms_ber.bin");
+    let _ci_ber = ContentInfo::from_ber(cms_ber).unwrap();
+
+    // let cms_der = include_bytes!("../tests/examples/cms_der.bin");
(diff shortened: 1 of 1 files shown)


tarcieri:
@dishmaker looks like we now need something along the lines of #1821

tarcieri:
I added basic handling of constructed `OBJECT STRING` tags which doesn't actually decode them but at least allows the existing implementation to get almost all of the way through the document:

```
Error { kind: Incomplete { expected_len: Length(10131), actual_len: Length(10130) }, position: Some(Length(10130)) }
```

I'm not sure why that would be happening exactly. It seems like some length is being miscomputed and it's trying to read too far into the message.

tarcieri:
I tried effectively reverting #1885. That gave the following error:

```
Error { kind: TrailingData { decoded: Length(10910), remaining: Length(2) }, position: Some(Length(10910)) }
```

That's a little confusing because it seems to be longer than the total message length (10902)

dishmaker:
My implementation too almost go to the end. Im not an expert in BER, but the example payload might be wrong.

tarcieri:
@dishmaker I think the test vector is fine and the problem is the EOC markers aren’t being consumed.

My understanding is `cms_ber.bin` was created by Apple tooling, `cms_der.bin` is the same message converted from BER to DER, and `cms` has existing tests that can parse `cms_der.bin`

tarcieri:
These tests were added in #1898

PR #1895, opened 2025-07-02
der: (re-)add `Reader::read_value` with EOC support

tarcieri:
One of the blockers for addressing indefinite length handling (#779) has been where to consume the EOC tag.

This commit (re)introduces `Reader::read_value` (added in #1877, removed in #1887) and handles decoding the EOC there. This gives us a single place where EOC can be handled for all constructed messages.

With these changes, the decoder is able to parse `cms_ber.bin` from `cms/tests`, from which the `cms_der.bin` file has been translated. This file provides a real-world example of nested indefinite lengths from CMS.

Note, however, that the example contains a constructed `Any` which isn't yet being correctly handled. A `TODO` for ensuring the BER and DER decode identically has been added.

cc @dishmaker

### DIFF
--- a/cms/tests/tests_from_pkcs7_crate.rs
+++ b/cms/tests/tests_from_pkcs7_crate.rs
@@ -141,3 +141,14 @@ fn cms_decode_signed_der() {
     // should match the original
     assert_eq!(reencoded_der_signed_data_in_ci, der_signed_data_in_ci)
 }
+
+#[test]
+fn cms_decode_signed_ber() {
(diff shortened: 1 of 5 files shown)


dishmaker:
Note: because read_nested is not needed for primitive types, I suggest to introduce `decode_value_nested`.

Therefore call chain would be:
decode -> read_value -> decode_value_nested -> read_nested -> decode_value

tarcieri:
@dishmaker I would love to find somewhere better to put this logic, but it ends up being quite difficult given the current usage

PR #1899, opened 2025-07-03
der: constructed `OctetString` support

tarcieri:
Converts `OctetString` to use `BytesOwned` internally and adds an initial decoder for constructed indefinite length data to `BytesOwned`.

This should be reusable for other constructed string types as well, although they'll need to receive a similar treatment.

### DIFF
--- a/der/src/bytes_owned.rs
+++ b/der/src/bytes_owned.rs
@@ -2,8 +2,8 @@
 //! library-level length limitation i.e. `Length::max()`.
 
 use crate::{
-    BytesRef, DecodeValue, DerOrd, EncodeValue, Error, Header, Length, Reader, Result, StrRef,
-    Writer, referenced::OwnedToRef,
+    BytesRef, Decode, DecodeValue, DerOrd, EncodeValue, Error, ErrorKind, Header, Length, Reader,
+    Result, StrRef, Writer, referenced::OwnedToRef,
 };
 use alloc::{boxed::Box, vec::Vec};
 use core::cmp::Ordering;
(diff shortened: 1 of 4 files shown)


der/src/bytes_owned.rs
```diff
@@ -64,6 +64,29 @@ impl<'a> DecodeValue<'a> for BytesOwned {
...
+                    return Err(reader.error(ErrorKind::IndefiniteLength));
+                }
+
+                // TODO(tarcieri): more efficient reader
+                let mut chunk = reader.read_vec(h.length)?;
+                bytes.append(&mut chunk);
+            }
+
+            return Self::new(bytes);
+        }
```
tarcieri, line 88: This is a pretty janky implementation that ideally would be both made more abstract and more efficient by reducing allocations

der/src/bytes_owned.rs
```diff
@@ -64,6 +64,29 @@ impl<'a> DecodeValue<'a> for BytesOwned {
     type Error = Error;
 
     fn decode_value<R: Reader<'a>>(reader: &mut R, header: Header) -> Result<Self> {
+        // Reassemble indefinite length string types
+        if header.length.is_indefinite() && !header.tag.is_constructed() {
+            // Parse constructed indefinite length data
```
dwhjames, line 69: @tarcieri, I’ve been following your recent work related to
- https://github.com/RustCrypto/formats/issues/779

I wasn’t following this change… the comment says ‘constructed’ and the goal is constructed OctetString support, but the condition says `! is_constructed`.
I must be missing something.

tarcieri: It's a bit confusing. The `Tag` type is designed for DER where the constructed encoding of e.g. string types is disallowed.

So that's essentially asking "would this type of data be encoded with the constructed bit in DER?". If it is (for example, a `SEQUENCE`), then it avoids trying to decode it.

PR #1916, opened 2025-07-07
der: auto `read_nested` in trait: `DecodeValue::decode_value_nested`

dishmaker:
This PR is a step towards removing overhead of `read_nested` for primitive types, that was introduced in #1877.

### DIFF
--- a/der/src/asn1/internal_macros.rs
+++ b/der/src/asn1/internal_macros.rs
@@ -130,7 +130,7 @@ macro_rules! impl_custom_class {
             where
                 T: Decode<'a>
             {
-                reader.read_nested(header.length, |r| T::decode(r))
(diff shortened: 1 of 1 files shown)


tarcieri:
I'm not sure what the goal of this PR is. I think there's some potential merit to this, but it also seems unrelated to this PR:

> I think reader.encoding_rules() should be generic, so BER optimizes out during compile time.

I played with that in [`der/ber-reader`](https://github.com/RustCrypto/formats/tree/der/ber-reader), however it does preclude runtime selection of DER vs BER. That said, I still think it's an option worth considering.

dishmaker:
The main goal is to optimize read_nested where it isn't necessary (e.g. primitive octet string).

But that probably wouldn't happen, because encoding_rules is a runtime value.

PR #1921, opened 2025-07-07
der: improved internal ref types

tarcieri:
Changes `BytesRef` and `StringRef` from types with an explicit lifetime which act as newtypes for inner `&'a [u8]` and `&'a str` types to being newtypes for `[u8]` and `str`.

This requires a very small amount of `unsafe` casting, but makes it possible to implement reference conversions entirely in terms of standard traits like `AsRef`, `Borrow`, `Deref`, and `ToOwned` as opposed to using our custom `RefToOwned` and OwnedToRef` traits. It makes it possible to use e.g. `Cow` as well.

This is the first step towards moving the entire crate to this pattern and eliminating the `OwnedToRef` and `RefToOwned` raits, replacing them instead with the standard `AsRef` and `ToOwned` traits.

### DIFF
--- a/der/src/asn1/any.rs
+++ b/der/src/asn1/any.rs
@@ -29,14 +29,14 @@ pub struct AnyRef<'a> {
     tag: Tag,
 
     /// Inner value encoded as bytes.
-    value: BytesRef<'a>,
+    value: &'a BytesRef,
 }
 
 impl<'a> AnyRef<'a> {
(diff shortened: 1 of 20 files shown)


tarcieri:
cc @dishmaker 

I'm going to try to do this conversion incrementally a bit at a time. That also makes it potentially possible to split up the work.

PR #1922, opened 2025-07-08
der: fix BER `IMPLICIT` constructed octet string

dishmaker:
Ignores `the encoding shall be constructed if the base encoding is constructed` when using BER.

Issue reported in #779

### DIFF
--- a/der/src/bytes.rs
+++ b/der/src/bytes.rs
@@ -156,7 +156,7 @@ pub(crate) mod allocating {
                 && header.length.is_indefinite()
                 && !inner_tag.is_constructed()
             {
-                return Self::new(read_constructed_vec(reader, header, inner_tag)?);
+                return Self::new(read_constructed_vec(reader, header.length, inner_tag)?);
             }
 
             Self::decode_value(reader, header)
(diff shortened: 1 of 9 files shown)


dishmaker:
@dwhjames is this correct? I added `inner_tag: Tag` (that would be `OctetString`) 
and is different from `header.tag: Tag` (which would be `ContextSpecific`)

der/src/length/indefinite.rs
```diff
@@ -81,6 +84,7 @@ pub(crate) fn read_eoc<'a>(reader: &mut impl Reader<'a>) -> crate::Result<()> {
 pub(crate) fn read_constructed_vec<'r, R: Reader<'r>>(
     reader: &mut R,
     header: Header,
```
tarcieri, line 86: Since `inner_tag` has taken over the role of `header.tag`, perhaps this should just be the length:

```suggestion
    length: Length,
```

der/src/bytes.rs
```diff
@@ -145,6 +145,22 @@ pub(crate) mod allocating {
                 inner,
             })
         }
+        /// Decodes [`BytesOwned`] as DER, or from parts, when using a BER reader.
+        pub fn decode_value_parts<'a, R: Reader<'a>>(
```
tarcieri, line 149: Not wild about this name, but I can't think of a better one for now

tarcieri (approved):

PR #1930, opened 2025-07-09
der: negative test case for recursively constructed octet string

dwhjames:
ASN.1 BER specification valid, but chosen as unsupported.

---

Follow up to
- https://github.com/RustCrypto/formats/issues/779#issuecomment-3049589340


tarcieri:
@dwhjames can you rebase?

dwhjames:
@tarcieri done in https://github.com/RustCrypto/formats/pull/1930/commits/f9d29492d1d509809703446b597033bc1ec6fabb

PR #1931, opened 2025-07-09
Negative test case for constructed, definite-length octet string

dwhjames:
Ensure octet strings with constructed, definite-length method return an error, rather than an incorrect decoding.

### DIFF
--- a/der/src/header.rs
+++ b/der/src/header.rs
@@ -10,19 +10,51 @@ use core::cmp::Ordering;
 #[derive(Copy, Clone, Debug, Eq, PartialEq)]
 pub struct Header {
     /// Tag representing the type of the encoded value
-    pub tag: Tag,
+    tag: Tag,
 
     /// Length of the encoded value
-    pub length: Length,
+    length: Length,
+
+    /// True if value is constructed, rather than primitive
+    constructed: bool,
 }
 
 impl Header {
-    /// Create a new [`Header`] from a [`Tag`] and a specified length.
-    ///
-    /// Returns an error if the length exceeds the limits of [`Length`].
-    pub fn new(tag: Tag, length: impl TryInto<Length>) -> Result<Self> {
-        let length = length.try_into().map_err(|_| ErrorKind::Overflow)?;
-        Ok(Self { tag, length })
+    /// Create a new [`Header`] from a [`Tag`] and a [`Length`].
+    pub fn new(tag: Tag, length: Length) -> Self {
+        #[cfg(feature = "ber")]
+        let constructed = tag.is_constructed() || length.is_indefinite();
+        #[cfg(not(feature = "ber"))]
+        let constructed = tag.is_constructed();
+        Self {
+            tag,
+            length,
+            constructed,
+        }
+    }
+
+    /// [`Tag`] of this header.
+    pub fn tag(&self) -> Tag {
+        self.tag
+    }
+
+    /// [`Length`] of this header.
+    pub fn length(&self) -> Length {
+        self.length
+    }
+
+    /// True if the [`Tag`] of this header has its constructed bit set.
+    pub fn is_constructed(&self) -> bool {
+        self.constructed
+    }
+
+    /// Copy of header with adjusted length.
+    pub fn with_length(&self, length: Length) -> Self {
+        Self {
+            tag: self.tag,
+            length,
+            constructed: self.constructed,
+        }
     }
 
     /// Peek forward in the reader, attempting to decode a [`Header`] at the current position.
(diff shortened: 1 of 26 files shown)


tarcieri:
Propagating the original `constructed` bit through the `Header` sounds good to me

der/src/header.rs
```diff
@@ -13,6 +13,8 @@ pub struct Header {
 
     /// Length of the encoded value
     pub length: Length,
+
+    constructed: bool,
```
tarcieri, line 17: Hmm, not sure how I feel about the mixed visibility here. Maybe make it `pub` for now and perhaps we could have a followup to make all of the fields private? (I imagine that would be a bit more invasive)

der/src/header.rs
```diff
@@ -21,7 +23,26 @@ impl Header {
...
+
+    /// TODO.
+    pub fn copy_with_length(&self, length: impl TryInto<Length>) -> Result<Self> {
+        let length = length.try_into().map_err(|_| ErrorKind::Overflow)?;
+        Ok(Self {
+            tag: self.tag,
+            length,
+            constructed: self.constructed,
+        })
+    }
```
tarcieri, line 41: If all fields are `pub` you don't need to worry about this

der/src/header.rs
dwhjames: Yep, definitely ugly.

Revised in https://github.com/RustCrypto/formats/pull/1931/commits/544e39783df87e5353a3691a37eb83928d87e400

---

> and perhaps we could have a followup to make all of the fields private? (I imagine that would be a bit more invasive)

I counted 24 occurrences in the der crate; for now I also access accessor methods so that new code has the option to use them in the event that the fields become private.

dwhjames: I went through with it: https://github.com/RustCrypto/formats/pull/1931/commits/0c5789e98ae2f42cdd07cb7378bc23c7b464d738

However, tests are failing due to at least
https://github.com/RustCrypto/signatures/blob/1ff759e67a1675d759e901fdb5903a07a1d7caec/ecdsa/src/der.rs#L206-L210

It looks like this change would have to be phased in?

Should I
1. roll back all the changes for using accessor methods rather than directly accessing struct fields, OR
2. roll back making the fields private but keep the pattern of using accessor methods rather than direct field access
?

der/src/header.rs
```diff
@@ -9,19 +9,55 @@ use core::cmp::Ordering;
...
+        self.length
+    }
+
+    /// True if the [`Tag`] of this header has its constructed bit set.
+    pub fn is_constructed(&self) -> bool {
+        self.constructed
+    }
+
+    /// Copy of header with adjusted length.
+    pub fn copy_with_length(&self, length: Length) -> Self {
```
tarcieri, line 55: Maybe just `with_length`?

der/src/header.rs
tarcieri: Looks great so far! Easier than I expected, actually

You can open a PR to https://github.com/RustCrypto/signatures which makes the necessary changes, using a `[patch.crates-io]` directive to pull in this `git` branch.

After that, in a bit of an ouroboros you can then update this PR with a `[patch.crates-io]` directive to pull in `ecdsa` from your branch on https://github.com/RustCrypto/signatures. That should get the tests passing.

der/src/header.rs
dwhjames: Sure; changed to `with_length` in https://github.com/RustCrypto/formats/pull/1931/commits/8a42c51c85fb45ee009af027fd3677e713b4eb30

der/src/header.rs
```diff
@@ -9,19 +9,55 @@ use core::cmp::Ordering;
...
+            length,
+            constructed: length.is_indefinite(),
+        }
+    }
+
     /// Create a new [`Header`] from a [`Tag`] and a specified length.
     ///
     /// Returns an error if the length exceeds the limits of [`Length`].
-    pub fn new(tag: Tag, length: impl TryInto<Length>) -> Result<Self> {
+    pub fn try_new(tag: Tag, length: impl TryInto<Length>) -> Result<Self> {
```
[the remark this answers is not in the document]
dwhjames: It does not; I initially thought it best to preserve the flexibility of the original, but there seem to be no current/remaining cases of leveraging the `TryInto`.

Removed in https://github.com/RustCrypto/formats/pull/1931/commits/8a42c51c85fb45ee009af027fd3677e713b4eb30

der/src/header.rs
dwhjames: Ok, I think I followed your guidance

1. https://github.com/RustCrypto/signatures/compare/master...dwhjames:RustCrypto-signatures:tmp_der_header_refactor
2. https://github.com/RustCrypto/formats/pull/1931/commits/fa5700cd9ce69d52af8c2ce0d86a06a579cb4ea7

PR #1936, opened 2025-07-10
der: add `DecodeValueExt` trait with fn `from_der_value`

dishmaker:
Motivation: I didn't know about

```rust
AnyRef::new(tag, value)?.decode_as::<Self>()
```
trick, when writing https://github.com/RustCrypto/crypto-bigint/pull/851/files#diff-87fe87bbce5380197d3c659465e30daf47af5cac570bb4058f467a5b2dffa0b2R164


tarcieri:
I'm not sure the solution is a trait method exposed everywhere

PR #1941, opened 2025-07-14
cms: Added a test for BER-CMS

bkstein:
This is my real-world test of a BER encoded CMS message using indefinite lengths. I removed the content, but kept the structure.
This kind of messages is sent by EJBCA when certificates are enrolled using SCEP protocol.


bkstein:
The `EXAMPLE_BER` is not a valid CMS message. I used it to test, if my own preliminary BER-to-DER converter works on nested indefinite structures (ignoring the content).

Issue #1808, opened 2025-05-05
der: derive(Choice) needs `constructed = "true"` flag


dishmaker:
`derive(Choice)` needs `constructed = "true"` flag, 
but `derive(Sequence)` does not.

```rust

#[derive(Choice)]
pub enum ImplicitChoice<'a> {
    // This works:
    #[asn1(context_specific = "3", tag_mode = "IMPLICIT", constructed = "true")]
    SequenceOfNulls(SequenceOf<Null, 1>),

    // This won't decode:
    #[asn1(context_specific = "3", tag_mode = "IMPLICIT")]
    SequenceOfNulls(SequenceOf<Null, 1>),
}
```

PR #1945, opened 2025-07-15
der: add `ApplicationExplicit`, `ContextSpecificExplicit`, `PrivateExplicit`

dishmaker:
Partially fixes
- #1745

Derive macro support in future PR.

### DIFF
--- a/der/src/asn1.rs
+++ b/der/src/asn1.rs
@@ -35,10 +35,10 @@ mod videotex_string;
 
 pub use self::{
     any::AnyRef,
-    application::{Application, ApplicationRef},
+    application::{Application, ApplicationExplicit, ApplicationRef},
     bit_string::{BitStringIter, BitStringRef},
     choice::Choice,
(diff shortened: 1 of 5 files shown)


dishmaker:
This partially fixes #1808, because we rewrite derive(Choice) as:
```rust
match tag {
    String::TAG => {}
    // ContextSpecificExplicit knows it's constructed, without trickery
    ContextSpecificExplicit::<2, BitString>::TAG => {}
    // .. more choices
}
```

tarcieri:
I'm getting a little worried about the `internal_macros` module. It's a lot of macros. Perhaps there are some non-macro-based abstractions to explore there.

dishmaker:
So my plan is to make this PR pure generic `*Explicit`.

`CustomTagExplicit<TAG, T, CLASS>`

Meanwhile, macro code will stay the same and will be eventually deprecated (because of trait flaws).

Issue #1934, opened 2025-07-10
der: Tracking issue for deoptimiztions since `v0.7`


dishmaker:
Due to BER:
- using `ber` features makes DER reader slow (about 20% slower)
- `read_nested` now includes primitive `DecodeValue`s too 
  (added in #1877)
  (partial fix in #1916)
- ~indefinite flag in `Length`
  (#1894)~
- ~users of pure DER need to compile BER code too (maybe add feature flag?)
  (added in #1899)~
- ~constructed flag in `Header`
  (will be added in #1931)~

tarcieri:
The latest master is what to compare to. I imagine `peek_is_constructed`, which was removed, was not particularly fast.

That said I'm a little surprised the current `master` is that much slower. A feature flag might be interesting to see if it can recover all of the performance.

dishmaker:
Here is a comparison of https://github.com/RustCrypto/formats/pull/1726

500cd652e66dd8191776ccf9720bd75a2bb5b716 vs 0b3a0b2e00741e54e81ae0b742cd5fcacd640d2e

Looks like `u32::MAX` would be about 1% slower (similar to `Length::MAX`)

|          |After / no limit [ns/iter]|Before / with limit [ns/iter]|
|----------------|--------------------------|-----------------------------|
|                |219,64                    |221,78                       |
|                |217,45                    |223,65                       |
|                |215,18                    |226,07                       |
|                |219,46                    |223,19                       |
|                |217,39                    |216,55                       |
|git rebase again|                          |                             |
|                |219,15                    |216,43                       |
|                |216,06                    |220,75                       |
|                |219,33                    |215,86                       |
|                |216,37                    |214,27                       |
|                |216,26                    |215,14                       |
|                |214,06                    |224,7                        |
|                |221,29                    |224,1                        |
|                |218                       |220,31                       |
|                |220,99                    |221,03                       |
|git rebase again|                          |                             |
|                |220,21                    |214,04                       |
|                |219,93                    |218,8                        |
|                |221,31                    |223,28                       |
|                |216,87                    |227,41                       |
|                |218,95                    |222,66                       |
|                |209,68                    |218,36                       |
|                |219,15                    |222,18                       |
|average:        |217,94                    |220,50                       |

PR #1948, opened 2025-07-15
der: add `ber` feature

dishmaker:
As noted in #1934, there is noticeable performance impact of adding `indefinite: bool` field to `Length`.

This PR allows to disable BER decoder.

### DIFF
--- a/der/Cargo.toml
+++ b/der/Cargo.toml
@@ -41,6 +41,7 @@ derive = ["dep:der_derive"]
 oid = ["dep:const-oid"]
 pem = ["dep:pem-rfc7468", "alloc", "zeroize"]
 real = []
+ber = ["alloc"]
 
 [package.metadata.docs.rs]
 all-features = true
--- a/der/src/encoding_rules.rs
+++ b/der/src/encoding_rules.rs
@@ -13,6 +13,7 @@ use core::{fmt, str::FromStr};
 #[derive(Clone, Copy, Debug, Default, Eq, PartialEq, PartialOrd, Ord)]
 pub enum EncodingRules {
     /// Basic Encoding Rules.
+    #[cfg(feature = "ber")]
     Ber,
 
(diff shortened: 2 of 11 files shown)


der/src/encoding_rules.rs
```diff
@@ -13,6 +13,7 @@ use core::{fmt, str::FromStr};
 #[derive(Clone, Copy, Debug, Default, Eq, PartialEq, PartialOrd, Ord)]
 pub enum EncodingRules {
     /// Basic Encoding Rules.
+    #[cfg(feature = "ber")]
```
tarcieri, line 16: It seems like you should be able to feature gate this whole type/module, and then remove all the codepaths that use it, since without `ber` it's unnecessary

dishmaker: It's hard to feature-gate of `SliceReader::new_with_encoding_rules`

tarcieri: I think it shouldn't be too hard, but I can take care of it myself

der/Cargo.toml
```diff
@@ -41,6 +41,7 @@ derive = ["dep:der_derive"]
 oid = ["dep:const-oid"]
 pem = ["dep:pem-rfc7468", "alloc", "zeroize"]
 real = []
+ber = ["alloc"]
```
tarcieri, line 44: I missed this... why are you enabling `alloc`?

dishmaker: Reading constructed octet string requires `Vec::new`

tarcieri: Yes, but that's impl'd in the `OctetString` type, which is already gated on `alloc`

dishmaker: Oh https://github.com/RustCrypto/formats/blob/bd840638835fcf66c4ffc3baed6d7c3b727c4ef3/der/src/length/indefinite.rs#L86

PR #1949, opened 2025-07-17
der: add `EncodeValueLenCached` wrapper

dishmaker:
Sometimes, when encoding big DER structures, `value_len` gets called many times. 

In order to limit amount of recursive calls, this wrapper caches the computed length, 
until the value gets encoded.


dishmaker:
I don't have an example. This just seems to be reasonable, where PKCS#15 nests objects recursively ([example](https://github.com/monai/node-passport/blob/master/lib/pkcs15/cryptographic_information_framework.asn1#L234-L242))
Actually, this newtype can be in a separate crate.

PR #1950, opened 2025-07-17
der: don't enable `alloc` when enabling `ber`

tarcieri:


tarcieri:
Note: I'm not against `ber` having an `alloc` requirement, but it currently doesn't

PR #1952, opened 2025-07-18
cms: Implement types from RFC 6031 - Cryptographic Message Syntax (CMS) Symmetric Key Package Content Type

callumadair:
Implementation of types as defined in [RFC 6031](https://datatracker.ietf.org/doc/html/rfc6031#section-2)

### Several queries as I was implementing:
- I've used `Utf8StringRef` in the types that require an ASN.1 UTF8String, would it be preferable to have the structs own this data and use the `String` type instead?
- const-oid seems to generate relevantt oids by parsing RFC documents, should I therefore add the RFC-6031.txt file to the ones already present in oidbindgen?

### DIFF
--- a/cms/src/symmetric_key.rs
+++ b/cms/src/symmetric_key.rs
@@ -206,7 +206,7 @@ pub type PSKCKeyUsage = String;
 ///      minLength         [3] INTEGER (0..MAX) OPTIONAL,
 ///      maxLength         [4] INTEGER (0..MAX) OPTIONAL,
 ///      pinEncoding       [5] Encoding OPTIONAL }
-///
+/// ```
 ///
 /// [RFC 6031 Section 3.3.5]: https://datatracker.ietf.org/doc/html/rfc6031#section-3.3.5
 #[derive(Sequence, PartialEq, Eq)]
--- a/cms/src/symmetric_key.rs
+++ b/cms/src/symmetric_key.rs
@@ -1,8 +1,66 @@
 //! SymmetricKeyPackage types
 
 use alloc::{string::String, vec::Vec};
-use der::{Enumerated, Sequence, asn1::OctetString};
+use der::ErrorKind;
+use der::asn1::Utf8StringRef;
+use der::{
+    DecodeValue, EncodeValue, Enumerated, FixedTag, Header, Length, Reader, Sequence, Tag, Writer,
+    asn1::OctetString,
+};
 use x509_cert::attr::Attribute;
+
+macro_rules! impl_str_enum {
+    (
+        $(#[$attr:meta])* $vis: vis enum $name: ident {
+            $( $(#[$attr_item: meta])* $item: ident => $value: expr,)*
+        }
+    )=> {
+        $(#[$attr])*
+        $vis enum $name {
+            $( $(#[$attr_item])* $item, )*
+        }
+
+        impl AsRef<str> for $name {
+            fn as_ref(&self) -> &str {
+                match self {
+                    $( Self::$item => $value, )*
+                }
+            }
+        }
+
+        impl<'d> DecodeValue<'d> for $name {
+            type Error = der::Error;
+
+            fn decode_value<R: Reader<'d>>(reader: &mut R, header: Header) -> der::Result<Self> {
+                let value = Utf8StringRef::decode_value(reader, header)?;
+
+                match value.as_ref() {
+                    $( $value => Ok(Self::$item), )*
+                    _ => Err(der::Error::new(
+                        ErrorKind::Value { tag: Self::TAG },
+                        reader.position(),
+                    )),
+                }
+            }
+        }
+
+        impl EncodeValue for $name {
+            fn value_len(&self) -> der::Result<Length> {
+                Utf8StringRef::new(self.as_ref())?.value_len()
+            }
+
+            fn encode_value(&self, encoder: &mut impl Writer) -> der::Result<()> {
+                Utf8StringRef::new(self.as_ref())?.encode_value(encoder)
+            }
+        }
+
+        impl FixedTag for $name {
+            const TAG: Tag = String::TAG;
+        }
+
+
+    };
+}
 
 /// The `SymmetricKeyPackage` type is defined in [RFC 6031 Section 2.0].
 ///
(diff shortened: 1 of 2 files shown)


cms/src/symmetric_key.rs
```diff
@@ -0,0 +1,238 @@
...
+///
+/// [RFC 6031 Section 2.0]: https://datatracker.ietf.org/doc/html/rfc6031#section-2
+#[derive(Sequence, PartialEq, Eq)]
+#[allow(missing_docs)]
+pub struct SymmetricKeyPackage {
+    #[asn1(default = "Default::default")]
+    pub version: KeyPkgVersion,
+    #[asn1(context_specific = "0", tag_mode = "IMPLICIT", optional = "true")]
+    pub s_key_pkg_attrs: Option<Vec<Attribute>>,
+    pub skeys: SymmetricKeys,
```
baloo, line 28:
```suggestion
    pub s_keys: SymmetricKeys,
```

Just for consistency with the `sKeyPkgAttrs` field

cms/src/symmetric_key.rs
```diff
@@ -0,0 +1,238 @@
...
+pub type SymmetricKeys = Vec<OneSymmetricKey>;
+
+/// The `OneSymmetricKey` type is defined in [RFC 6031 Section 2.0].
+///
+/// ```text
+///        sKeyAttrs  SEQUENCE SIZE (1..MAX) OF Attribute
+///                                       {{ SKeyAttributes }}  OPTIONAL,
+///        sKey       OCTET STRING OPTIONAL }
+///        ( WITH COMPONENTS { ..., sKeyAttrs PRESENT } |
+///          WITH COMPONENTS { ..., sKey PRESENT } )
```
baloo, line 47:
```suggestion
///      OneSymmetricKey ::= SEQUENCE {
///        sKeyAttrs  SEQUENCE SIZE (1..MAX) OF Attribute
///                                       {{ SKeyAttributes }}  OPTIONAL,
///        sKey       OCTET STRING OPTIONAL }
///        ( WITH COMPONENTS { ..., sKeyAttrs PRESENT } |
///          WITH COMPONENTS { ..., sKey PRESENT } )
```

baloo:
> * I've used `Utf8StringRef` in the types that require an ASN.1 UTF8String, would it be preferable to have the structs own this data and use the `String` type instead?

Or a proper rust enum? Should we reject values with small differences from the one defined in the RFC?

tarcieri:
> I've used Utf8StringRef in the types that require an ASN.1 UTF8String, would it be preferable to have the structs own this data and use the String type instead?

As a general rule you can prefer the allocating types over the `*Ref` types for `cms`, since it has a hard `alloc` dependency

cms/src/symmetric_key.rs
```diff
@@ -0,0 +1,237 @@
...
+/// The `PSKCKeyUsage` type is defined in [RFC 6031 Section 3.3.4].
+///
+/// ```text
+///    PSKCKeyUsage ::= UTF8String ("OTP" | "CR" | "Encrypt" |
+///                     "Integrity" | "Verify" | "Unlock" | "Decrypt" |
+///                     "KeyWrap" | "Unwrap" | "Derive" | "Generate")
+/// ```
+///
+/// [RFC 6031 Section 3.3.4]: https://datatracker.ietf.org/doc/html/rfc6031#section-3.3.4
+pub type PSKCKeyUsage<'pskc_key_usage> = String;
```
baloo, line 197: the lifetime isn't used.
```suggestion
pub type PSKCKeyUsage = String;
```

cms/src/symmetric_key.rs
```diff
@@ -0,0 +1,237 @@
...
+///
+/// [RFC 6031 Section 3.2.7]: https://datatracker.ietf.org/doc/html/rfc6031#section-3.2.7
+#[derive(Sequence, PartialEq, Eq)]
+#[allow(missing_docs)]
+pub struct ChallengeFormat<'range> {
+    pub encoding: Encoding,
+    #[asn1(default = "Default::default")]
+    pub check_digit: bool,
+    pub min: der::asn1::IntRef<'range>,
+    pub max: der::asn1::IntRef<'range>,
```
baloo, line 130: You probably can use the owned version of int:
```suggestion
pub struct ChallengeFormat {
    pub encoding: Encoding,
    #[asn1(default = "Default::default")]
    pub check_digit: bool,
    pub min: der::asn1::Int,
    pub max: der::asn1::Int,
```

cms/src/symmetric_key.rs
```diff
@@ -0,0 +1,237 @@
...
+///      responseFormat   [1] ResponseFormat,
+///      ... }
+/// ```
+///
+/// [RFC 6031 Section 3.2.7]: https://datatracker.ietf.org/doc/html/rfc6031#section-3.2.7
+#[derive(Sequence, PartialEq, Eq)]
+#[allow(missing_docs)]
+pub struct PSKCAlgorithmParameters<'range> {
+    pub suite: String,
+    pub challenge_format: ChallengeFormat<'range>,
```
baloo, line 107:
```suggestion
pub struct PSKCAlgorithmParameters {
    pub suite: String,
    pub challenge_format: ChallengeFormat,
```

callumadair:
If there are no more review comments, is this able to be merged?

baloo:
> Or a proper rust enum? Should we reject values with small differences from the one defined in the RFC?

Would that be completely crazy?

```diff
diff --git a/cms/src/symmetric_key.rs b/cms/src/symmetric_key.rs
index 041162ca6c..abab33a8cf 100644
--- a/cms/src/symmetric_key.rs
+++ b/cms/src/symmetric_key.rs
@@ -1,7 +1,11 @@
 //! SymmetricKeyPackage types

 use alloc::{string::String, vec::Vec};
-use der::{Enumerated, Sequence, asn1::OctetString};
+use der::{
+    DecodeValue, EncodeValue, Enumerated, ErrorKind, FixedTag, Header, Length, Reader, Sequence,
+    Tag, Writer,
+    asn1::{OctetString, Utf8StringRef},
+};
 use x509_cert::attr::Attribute;

 /// The `SymmetricKeyPackage` type is defined in [RFC 6031 Section 2.0].
@@ -138,7 +142,64 @@
 /// ```
 ///
 /// [RFC 6031 Section 3.2.7]: https://datatracker.ietf.org/doc/html/rfc6031#section-3.2.7
-pub type Encoding = String;
+#[derive(Copy, Clone, PartialEq, Eq)]
+pub enum Encoding {
+    /// "DECIMAL"
+    Decimal,
+    /// "HEXADECIMAL"
+    Hexadecimal,
+    /// "ALPHANUMERIC"
+    Alphanumeric,
+    /// "BASE64"
+    Base64,
+    /// "BINARY"
+    Binary,
+}
+
+impl AsRef<str> for Encoding {
+    fn as_ref(&self) -> &str {
+        match self {
+            Self::Decimal => "DECIMAL",
+            Self::Hexadecimal => "HEXADECIMAL",
+            Self::Alphanumeric => "ALPHANUMERIC",
+            Self::Base64 => "BASE64",
+            Self::Binary => "BINARY",
+        }
+    }
+}
+
+impl<'d> DecodeValue<'d> for Encoding {
+    type Error = der::Error;
+
+    fn decode_value<R: Reader<'d>>(reader: &mut R, header: Header) -> der::Result<Self> {
+        let value = Utf8StringRef::decode_value(reader, header)?;
+        match value.as_ref() {
+            "DECIMAL" => Ok(Self::Decimal),
+            "HEXADECIMAL" => Ok(Self::Hexadecimal),
+            "ALPHANUMERIC" => Ok(Self::Alphanumeric),
+            "BASE64" => Ok(Self::Base64),
+            "BINARY" => Ok(Self::Binary),
+            _ => Err(der::Error::new(
+                ErrorKind::Value { tag: Self::TAG },
+                reader.position(),
+            )),
+        }
+    }
+}
+
+impl EncodeValue for Encoding {
+    fn value_len(&self) -> der::Result<Length> {
+        Utf8StringRef::new(self.as_ref())?.value_len()
+    }
+
+    fn encode_value(&self, encoder: &mut impl Writer) -> der::Result<()> {
+        Utf8StringRef::new(self.as_ref())?.encode_value(encoder)
+    }
+}
+
+impl FixedTag for Encoding {
+    const TAG: Tag = String::TAG;
+}

 /// The `ResponseFormat` type is defined in [RFC 6031 Section 3.2.7].
 ///
 ```
 
 I guess a macro could help with the boilerplate.

**Comment by callumadair** [comment]
> > Or a proper rust enum? Should we reject values with small differences from the one defined in the RFC?
> 
> Would that be completely crazy?


Ohh, I see how you mean now @baloo. I had thought you meant something like this in terms of ownnership vs Ref strings:

```rust
pub enum Utf8String<'string> {
    Owned(String),
    Ref(Utf8StringRef<'string>),
}
```

I think you're correct in that they would be perfect for an enum. Additionally a macro would  probably be beneficial as I think there are other RFCst that do have similar predefined strings as values, though I wonder if its implementation might be out of scope for this PR.
Something like the below maybe?

```rust
#[derive(EncodeAsStr)]
pub enum Encoding {
...
}
```

baloo:
I'm not aware of any other RFC where this happens, but there are many of them and I don't know them all.

I had a macro_rules in mind. I don't think we need an additional derive crate just for that. Anyway, that was more a question for others to chime in :)

tarcieri:
That enum looks like `Cow`, which we will hopefully support in the next release

tarcieri:
@baloo I was referring to this enum: https://github.com/RustCrypto/formats/pull/1952#issuecomment-3109754265

baloo (approved):

Issue #1795, opened 2025-05-02
der: use Cow for *Ref types


dishmaker:
Following https://github.com/RustCrypto/formats/pull/1778#issuecomment-2840198645, 
we need to investigate why Cow<*Ref> does not work.

One major issue is that `Clone` is implemented on `Ref` types.
https://github.com/RustCrypto/formats/blob/217b6eff7bf1b15d4fea21a3a186c2faee4a4344/der/src/asn1/ia5_string.rs#L37-L40

which collides with (if i remember correctly):
https://doc.rust-lang.org/std/borrow/trait.ToOwned.html#impl-ToOwned-for-T

tarcieri:
Yeah, we'd need to override `ToOwned`, which would require dropping `Copy`/`Clone` which seems annoying.

Part of the problem is those types are owned wrappers around references, where typically types you `Borrow` only exist as references (e.g. `str`, `[T]`, `Path`, where the references are trivially copyable), and `ToOwned` gets you back to an owned type, so it's not important to impl `Copy`/`Clone`.

One way this could potentially be fixed, with considerable effort, would be to turn the `*Ref` types into `repr(transparent)` wrappers for types like `[u8]` and `str` which are always accessed as e.g. `&'a Ia5StringRef` rather than `Ia5StringRef<'a>`, which would permit a proper `Borrow`/`ToOwned` relationship to e.g. `Ia5String`. This only works for "fat pointer" types though, and we have more complex `*Ref` types which are structs with a lifetime that have multiple fields that borrow.

dishmaker:
These types can be `repr(transparent)`. We can remove precomputed length (and ensure it's valid in all constructors).

`BytesRef`, `StrRef` as well as
`GeneralStringRef`, `OctetStringRef`, ... , `Ia5StringRef`, `Utf8StringRef`, `VideotexStringRef`

even `BitStringRef` can be `repr(transparent)` in a hacky way.

dishmaker:
We can use `#[cfg(feature = "alloc")]` on the `Owned` variant:

```rust
pub enum DerCow<'a, B>
where
    B: RefToOwned<'a> + ?Sized,
{
    /// referenced, for example `OctetStringRef`
    Borrowed(&'a B),

    #[cfg(feature = "alloc")]
    /// allocated, for example `OctetString`
    Owned(<B as RefToOwned<'a>>::Owned),
}
```

[...]

PR #1959, opened 2025-07-23
der: add `AnyCow`

dishmaker:
`AnyRef`, that allocates as `Any` with `alloc` feature

### DIFF
--- a/der/src/asn1/any.rs
+++ b/der/src/asn1/any.rs
@@ -170,6 +170,151 @@ impl<'a> TryFrom<&'a [u8]> for AnyRef<'a> {
     }
 }
 
+/// ASN.1 `ANY`: represents any explicitly tagged ASN.1 value.
+///
+/// This type provides the same functionality as [`AnyRef`] but may own the
+/// backing data.
+///
+/// [`AnyCow`] provides similar functionality to `Cow` type from standard library.
+pub enum AnyCow<'a> {
+    Borrowed(AnyRef<'a>),
+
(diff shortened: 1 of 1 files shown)


tarcieri:
I would prefer not to have any of these types that try to special case `Cow`.

We can likely do something similar to what I proposed in #1795 for `Any`/`AnyRef` using tagged pointers to store the tag for `AnyRef`

tarcieri:
Hmm, one problem with tagged pointers is that the `Tag` type can now need up to 48-bits to represent, which can't be stored in e.g. a 32-bit tagged pointer

dishmaker:
Yes, the `Tag` alongside `&'a [u8]` is a problem.

One parallel solution is:
```rust
pub struct AnyCow<'a> {
    tag: Tag,
    value: Cow<'a, BytesRef>,
}
```

but that wouldn't be `alloc`-gated on allocating variant.

tarcieri:
`BitString`/`BitStringRef` could work with `Cow`, but it would probably involve some hacks like using tagged pointers, and storing the owned `BitString` as a tagged pointer.

I don't think I can't do the same with `Any`/`AnyRef` though now that tags have been changed from `u8`, because there's too much data to put in the pointer

tarcieri:
I've changed my mind on `BitString*`: it can't work with `Cow` because we'd need a smart pointer type for a number of bits, which doesn't exist.

That said, I don't think this is the solution either.

PR #1977, opened 2025-08-02
der: add `new` and derives to SequenceRef

kamulos:
Copied some important things from `OctetStringRef`


baloo:
Is there a use-case for this? You don't generally build a Sequence from bytes.

Please reopen if you see a use-case.

Issue #1976, opened 2025-08-01
der: breaking changes in `ObjectIdentifier` and `AnyRef` with `tag_mode = IMPLICIT`


kamulos:
So I was just trying to change my application from my self-made support for private tags to the upstream support in the release candidates. (Thanks @tarcieri and @turbocool3r for following through on that).

In my code two things broke, that might not be intentional. So this example is a minimized version of my issues:

```rust
use der::{AnyRef, Decode, FixedTag, Sequence, oid::ObjectIdentifier};

#[derive(Debug, Sequence)]
pub struct SomeType<'a> {
    pub oid: ObjectIdentifier,
    #[asn1(context_specific = "0", tag_mode = "IMPLICIT")]
    pub data: AnyRef<'a>,
}

fn main() {
    let _ = ObjectIdentifier::TAG;

    let bytes = b"\x30\x12\x06\x07\x2a\x86\x48\xce\x3d\x02\x01\x80\x07\x63\x6f\x6e\x74\x65\x6e\x74";
    let decoded = SomeType::from_der(bytes).unwrap();
    println!("{:02x?}", decoded.data.value());
}
```

1. `ObjectIdentifier::TAG` does not work anymore without specifying `MAX_SIZE`. This broke in d4395cdb10f3b7ecfb2ccba4d3dd29e74f576576
2. `AnyRef` does not work anymore with `tag_mode = IMPLICIT`. That broke in be1caa3cc42b06bf01a37d1858826a4905b74904

dishmaker:
Well, 
1. `ObjectIdentifier::TAG` is equal to `Tag::ObjectIdentifier`.
   Sorry for the breaking change. https://github.com/RustCrypto/formats/pull/1851#discussion_r2249318528
2. `AnyRef` does not know if it is constructed, statically, and `IMPLICIT` en/decoder needs to know that.
    Also, tag inside of `AnyRef` looks unused and reduntant, so I recommend to use `OctetStringRef` instead of `AnyRef`.

```rust
#[asn1(context_specific = "0", tag_mode = "IMPLICIT")]
pub data: OctetStringRef<'a>,
```

or even better:
```rust
#[asn1(context_specific = "0", tag_mode = "IMPLICIT")]
pub data: &'a [u8],
```

kamulos:
1. `Tag::ObjectIdentifier` is actually way better in my opinion. So I am happy about that :) I am however wondering why this broke: shouldn't Rust take the default if it cannot infer the value?
[...]

PR #1980, opened 2025-08-04
der: add `new` and derives to SequenceRef

kamulos:
This is a reopening of #1977 which was closed because there was too little explaination. The motivation for adding the derives is that I cannot use `SequenceRef` in my structs that I want to output in the debug format or compare for equality.

The motivation for creating new `SequenceRef` is a bit harder to explain: I have something like X.509 `ContentInfo` that needs raw data access. When decoding I need the buffer data to validate a signature and I need to add a raw buffer which I have signed when compiling the data. I also want to use `der_derive` annotations to do all the validation for me.

Up until recently I was using `AnyRef` for this purpose, but this broke in the recent `0.8,0` release candidates (please see #1976 for more info). The thing is, instead of an explicit tag mode like in `ContentInfo` the format I am parsing is using the implicit tag mode and has the `constructed` bit set. `AnyRef` does not work with this anymore and I can understand the reasoning.

The next best thing we found in the linked case was (ab)using `SequenceRef` because it accepts having the implicit tag mode and the constructed bit set. And to be able to construct the type I need some kind of constructor for `SequenceRef`.

To be honest I would prefer having something that is more intentional for the purpose of raw data access, but still allows me to use the derive macro to specify the tag and tag mode implicit.

Tagging @dishmaker because we had a lengthy discussion in the linked case

Copying the example from the linked case. This used to work, but does not anymore:

```rust
use der::{AnyRef, Decode, Sequence, oid::ObjectIdentifier};

#[derive(Debug, Sequence)]
pub struct SomeType<'a> {
    pub oid: ObjectIdentifier,
    pub data: AnyRef<'a>,
}

fn main() {
    let bytes = b"\x30\x12\x06\x07\x2a\x86\x48\xce\x3d\x02\x01\xa0\x07\x63\x6f\x6e\x74\x65\x6e\x74";
    let decoded = SomeType::from_der(bytes).unwrap();
    println!("{:02x?}", decoded.data.value());
}
```

### DIFF
--- a/der/src/asn1/sequence.rs
+++ b/der/src/asn1/sequence.rs
@@ -2,8 +2,8 @@
 //! `SEQUENCE`s to Rust structs.
 
 use crate::{
-    BytesRef, DecodeValue, EncodeValue, Error, FixedTag, Header, Length, Reader, Result, Tag,
-    Writer,
+    BytesRef, DecodeValue, EncodeValue, Error, ErrorKind, FixedTag, Header, Length, Reader, Result,
+    Tag, Writer,
 };
 
 #[cfg(feature = "alloc")]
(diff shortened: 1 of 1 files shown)


der/src/asn1/sequence.rs
```diff
@@ -28,12 +28,20 @@ impl<'a, T> Sequence<'a> for Box<T> where T: Sequence<'a> {}
...
 ///
 /// This is a zero-copy reference type which borrows from the input data.
+#[derive(Copy, Clone, Debug, Eq, PartialEq, PartialOrd, Ord)]
 pub struct SequenceRef<'a> {
     /// Body of the `SEQUENCE`.
     body: &'a BytesRef,
 }
 
 impl<'a> SequenceRef<'a> {
+    /// Create a new [`SequenceRef`] from the provided DER bytes.
```
baloo, line 38: I'm missing something, why can't we rely on the `DecodeValue::decode_value` instead?

PR #1998, opened 2025-08-18
der: remove lifetime from `OctetStringRef`

tarcieri:
Following the pattern of #1921, removes the lifetime from the struct, instead changing `OctetStringRef` to a proper reference type to be used as `&OctetStringRef`.

This makes it possible to `impl Borrow<OctetStringRef> for OctetString` and `impl ToOwned for OctetStringRef`, so they can work with `Cow`.

cc @dishmaker

### DIFF
--- a/der/src/asn1/octet_string.rs
+++ b/der/src/asn1/octet_string.rs
@@ -5,27 +5,43 @@ use crate::{
     Tag, Writer, asn1::AnyRef, ord::OrdIsValueOrd,
 };
 
+// TODO(tarcieri): custom derive hack until the logic is updated to support `&'a` reference types
+#[doc(hidden)]
+pub type OctetStringRef2<'a> = &'a OctetStringRef;
+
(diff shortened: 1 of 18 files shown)


der/src/asn1/octet_string.rs
```diff
@@ -361,7 +369,23 @@ mod bytes {
...
-    use crate::asn1::{OctetStringRef, PrintableStringRef};
+    use crate::{
+        Decode,
+        asn1::{OctetStringRef, PrintableStringRef},
+    };
+    use hex_literal::hex;
+
+    #[test]
+    fn octet_string_decode() {
+        // PrintableString "hi"
```
dishmaker, line 380:
```suggestion
```

tarcieri:
Well, with a few hacks this is green. I think the hacks might be OK to merge with, but we'd ideally want to get them removed before a final release.

tarcieri:
Gonna land this. We can work through custom derive support separately. I'll make an issue about it.

PR #2011, opened 2025-08-22
der: add `DecodeValue::decode_value_nested`

dishmaker:
This PR moves some important logic out of `pub(crate) fn read_value`.

Such trait function allows us to disable `read_nested` selectively for some primitive types, like `OctetStringRef`.


tarcieri:
@dishmaker have you benchmarked what the actual impact of this is?

dishmaker:
I did not benchmark it.

PR #2022, opened 2025-09-03
tls_codec: optimize serde serialized size using serde_bytes

kkohbrok:
By default, serde implementations treat `Vec<u8>` like any other vec. `serde_bytes` allows more optimal serialization. This PR adds `serde_bytes` as a dependency and enables that optimization for `VLBytes`. Deserialization is still compatible with the old serialization format. This PR adds a test to ensure that this is the case.

This PR does not change the serialization format of the `TlsByteVec` variants. Their generic implementation via macro complicates things a bit.

See the [`serde_bytes`](https://docs.rs/serde_bytes/latest/serde_bytes/) crate for more details.

### DIFF
--- a/tls_codec/CHANGELOG.md
+++ b/tls_codec/CHANGELOG.md
@@ -7,9 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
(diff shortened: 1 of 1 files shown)


franziskuskiefer (approved):
Thanks!

PR #2034, opened 2025-09-07
x509-cert: Don't mark binary test data executable

haxtibal:
Some *.ta files in test example data were marked as executable in git. Content is der-encoded TAC objects, so this looks like an oversight and the files can be marked non-executable.


tarcieri:
Possibly added from Windows. Thanks

PR #2057, opened 2025-10-07
Derive `Default` for enums

dishmaker:
nightly `cargo clippy --all --all-features --tests --fix`

Similar to https://github.com/RustCrypto/traits/pull/2035

### DIFF
--- a/tls_codec/derive/tests/decode.rs
+++ b/tls_codec/derive/tests/decode.rs
@@ -9,7 +9,9 @@ use tls_codec_derive::{TlsDeserialize, TlsDeserializeBytes, TlsSerialize, TlsSiz
     TlsDeserialize, TlsDeserializeBytes, Debug, PartialEq, Clone, Copy, TlsSize, TlsSerialize,
 )]
 #[repr(u16)]
+#[derive(Default)]
 pub enum ExtensionType {
+    #[default]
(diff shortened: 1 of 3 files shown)


tls_codec/derive/tests/decode.rs
```diff
@@ -9,7 +9,9 @@ use tls_codec_derive::{TlsDeserialize, TlsDeserializeBytes, TlsSerialize, TlsSiz
     TlsDeserialize, TlsDeserializeBytes, Debug, PartialEq, Clone, Copy, TlsSize, TlsSerialize,
 )]
 #[repr(u16)]
+#[derive(Default)]
 pub enum ExtensionType {
+    #[default]
```
baloo, line 14: That seems like a bad idea to default on `Reserved`

baloo (commented):
Not every enum should have a default. I think the clippy is still in nightly and should not be given too much weight.

dishmaker:
But these enums had `impl Default` before

baloo:
Ha you mean I also need to read the *left* side of the diff too?! Sorry I didn't have a coffee just yet.

Yeah, this is a straight replacement which doesn't change the behavior. I'm still convinced we should at least change the Version one to default to V3.

Issue #2062, opened 2025-10-09
der: Decode error thrown without Some(position)


dishmaker:
After PR:
- https://github.com/RustCrypto/formats/pull/1889

most cases now return error position, but few untreated errors remain, for example:

https://github.com/RustCrypto/formats/blob/3ee0028ef967b0831df6fc6338e2620ea03efd43/der/src/asn1/generalized_time.rs#L112
https://github.com/RustCrypto/formats/blob/858f3af9df8398cb456ebe05b41f82e0335e66d8/der/src/asn1/integer/uint.rs#L330

https://github.com/RustCrypto/formats/blob/858f3af9df8398cb456ebe05b41f82e0335e66d8/der/src/asn1/oid.rs#L19

PR #2080, opened 2025-10-28
der: track reader error position in `ObjectIdentifier::decode_value`

dishmaker:
Partial fix for:

- #2062

### DIFF
--- a/der/src/asn1/oid.rs
+++ b/der/src/asn1/oid.rs
@@ -21,12 +21,9 @@ impl<'a, const MAX_SIZE: usize> DecodeValue<'a> for ObjectIdentifier<MAX_SIZE> {
         let actual_len = reader.read_into(slice)?.len();
         debug_assert_eq!(actual_len, header.length().try_into()?);
 
-        let oid_ref = ObjectIdentifierRef::from_bytes(slice)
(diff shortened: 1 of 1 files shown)


der/src/asn1/oid.rs
```diff
@@ -16,11 +16,16 @@ impl<'a, const MAX_SIZE: usize> DecodeValue<'a> for ObjectIdentifier<MAX_SIZE> {
...
 
         let actual_len = reader.read_into(slice)?.len();
         debug_assert_eq!(actual_len, header.length().try_into()?);
-        Ok(ObjectIdentifierRef::from_bytes(slice)?.try_into()?)
+
+        let oid_ref = ObjectIdentifierRef::from_bytes(slice)
+            .map_err(|oid_err| reader.error(Error::from(oid_err).kind()))?;
+        Ok(oid_ref
+            .try_into()
+            .map_err(|oid_err| reader.error(Error::from(oid_err).kind()))?)
```
tarcieri: Seems like you could consolidate these with some combinators. I'm not sure this quite works but the general idea should:

```suggestion
        ObjectIdentifierRef::from_bytes(slice)
            .and_then(TryInto::try_into)
            .map_err(|err| reader.error(Error::from(err).kind()))
```

PR #2085, opened 2025-11-06
base64ct, base32ct: proptests for decoder equivalence with base32/64

nmathewson:
For the base64 case, this MR adds proptests to ensure that base64 and base64ct accept the same inputs when decoding, and produce the same outputs when decoding them.

For the base32 case, that property isn't actually true.  Instead, the proptest makes sure that whenever base32 and base32ct both accept an input when decoding, they produce the same output.

Based on discussion at #2084.

### DIFF
--- a/base32ct/tests/proptests.rs
+++ b/base32ct/tests/proptests.rs
@@ -34,4 +34,35 @@ proptest! {
         let expected = base32::encode(RFC4648_PADDED, &bytes).to_lowercase();
         prop_assert_eq!(actual, expected);
     }
+
+    /// Make sure that, if base32ct and base32 _both_ decode a value
+    /// when expecting padded inputs, they give the same output.
+    ///
+    /// TODO: It might be desirable to ensure that they both decode the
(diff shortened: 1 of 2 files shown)


base32ct/tests/proptests.rs
```diff
@@ -34,4 +34,35 @@ proptest! {
...
+    /// which inputs they accept.  But first, we should verify that
+    /// `base32`'s behavior is actually what we want.
+    #[test]
+    fn decode_arbitrary_padded(string in string_regex("[a-z0-9]{0,32}={0,8}").unwrap()) {
+        let actual = Base32Ct::decode_vec(&string);
+        let expected = base32::decode(RFC4648_PADDED, &string);
+    // assert_eq!(actual.ok(), expected);
+    if let (Ok(a), Some(b)) = (actual, expected) {
+        assert_eq!(a, b);
+    }
```
tarcieri: I'm a little confused why this isn't causing a rustfmt failure

tarcieri (approved):

PR #2094, opened 2025-11-07
der: have `PemReader` decode to `Cow::Owned`

tarcieri:
Adds an associated `CAN_READ_SLICE` boolean flag to the `Reader` trait.

When false, `Cow` decodes the owned type rather than the borrow one.

This makes it possible to use `Cow` with `PemReader`, where it will automatically decode the owned type.

### DIFF
--- a/der/src/decode.rs
+++ b/der/src/decode.rs
@@ -212,14 +212,20 @@ where
 }
 
 #[cfg(feature = "alloc")]
-impl<'a, T> DecodeValue<'a> for Cow<'a, T>
+impl<'a, T, E> DecodeValue<'a> for Cow<'a, T>
 where
     T: ToOwned + ?Sized,
-    &'a T: DecodeValue<'a>,
+    &'a T: DecodeValue<'a, Error = E>,
+    T::Owned: for<'b> DecodeValue<'b, Error = E>,
+    E: From<Error> + 'static,
 {
-    type Error = <&'a T as DecodeValue<'a>>::Error;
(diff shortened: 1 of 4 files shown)


der/src/decode.rs
```diff
@@ -212,14 +212,20 @@ where
 }
 
 #[cfg(feature = "alloc")]
-impl<'a, T> DecodeValue<'a> for Cow<'a, T>
+impl<'a, T, E> DecodeValue<'a> for Cow<'a, T>
```
tarcieri: This still somewhat annoyingly ties the lifetime of the `Cow` to the lifetime of the `Reader`, which for `PemReader` is still the lifetime of the input buffer.

tarcieri: Aha, that's easily fixable by making it impl `Reader<'static>` instead

der/src/decode.rs
```diff
@@ -212,14 +212,20 @@ where
...
 where
     T: ToOwned + ?Sized,
-    &'a T: DecodeValue<'a>,
+    &'a T: DecodeValue<'a, Error = E>,
+    T::Owned: for<'b> DecodeValue<'b, Error = E>,
```
dishmaker: Oh no, it's HRTB

tarcieri: It's really necessary here to show that the output lifetime is independent of the input lifetimes, which should let `Reader<'static>` work when decoding `Cow<'static, T>`

PR #2096, opened 2025-11-08
pem-rfc7468 v1.0.0

tarcieri:
### Added
- `detect_base64_line_width` and `new_detect_wrap` ([#1464])

### Changed
- Use `core::error::Error` ([#1571])
- Upgrade to Rust 2024 edition; MSRV 1.85 ([#1670])

[#1464]: https://github.com/RustCrypto/formats/pull/1464
[#1571]: https://github.com/RustCrypto/formats/pull/1571
[#1670]: https://github.com/RustCrypto/formats/pull/1670

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -896,7 +896,7 @@ dependencies = [
 
 [[package]]
 name = "pem-rfc7468"
-version = "1.0.0-rc.3"
+version = "1.0.0"
 dependencies = [
  "base64ct",
 ]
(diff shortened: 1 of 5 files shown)


tarcieri:
To give a short stabilization report:

`pem-rfc7468` is now over 4 years old and has not seen real breaking changes since the v0.6.0 release in April of 2022.

The only real "breaking" changes since then have been to MSRV, which was previously considered breaking under our old MSRV policy. The policy has since been updated to denote that MSRV changes can be made in patch releases.

The crate is effectively finished and feature-complete.

PR #2120, opened 2025-12-11
refactor(ext)!: improve extension API flexibility and ergonomics

npmccallum:
Redesigned the extension trait interface to support more flexible ownership patterns and use cases. The trait is renamed from AsExtension to ToExtension and now consumes values by default, with implementations for references and tuples like (OID, bool, &T) to allow direct control over criticality and OID. This makes the API more ergonomic for common scenarios while enabling advanced use cases that weren't previously possible.

BREAKING CHANGE: The AsExtension trait has been renamed to ToExtension, and to_extension now takes self by value. Users must update trait bounds from AsExtension to ToExtension and may need to adjust how extensions are passed to add_extension methods.

### DIFF
--- a/x509-cert/src/ext.rs
+++ b/x509-cert/src/ext.rs
@@ -35,6 +35,18 @@ pub struct Extension {
     pub extn_value: OctetString,
 }
 
+impl ToExtension for Extension {
+    type Error = der::Error;
+
+    fn to_extension(
+        self,
+        _subject: &crate::name::Name,
+        _extensions: &[Extension],
+    ) -> Result<Extension, Self::Error> {
+        Ok(self)
+    }
+}
+
 /// Extensions as defined in [RFC 5280 Section 4.1.2.9].
 ///
 /// ```text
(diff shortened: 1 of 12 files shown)


npmccallum:
@tarcieri This resolves your complaint about the interface being named AsExtension.

npmccallum:
Tests all pass after the transient test breakage has resolved.

x509-cert/src/ext.rs
```diff
@@ -91,44 +103,68 @@ pub trait Criticality {
...
 /// ```
-pub trait AsExtension {
+pub trait ToExtension {
     /// The error type returned when encoding the extension.
     type Error;
 
     /// Returns the Extension with the content encoded.
     fn to_extension(
-        &self,
+        self,
```
tarcieri: As it were, the convention for methods that consume self is `into_*`:

https://rust-lang.github.io/api-guidelines/naming.html#ad-hoc-conversions-follow-as_-to_-into_-conventions-c-conv

So how about `IntoExtension::into_extension`?

tarcieri: I'm going to merge this PR, but followup with the change to `IntoExtension`

PR #2127, opened 2025-12-18
Remove `bincode` dev-dependency

tarcieri:
`bincode` is unmaintained. See https://crates.io/crates/bincode/3.0.0

This adds `postcard` as an alternative for rustdocs


baloo:
postcard <3

PR #2149, opened 2026-01-03
sec1: replace `subtle` with `ctutils`

tarcieri:
This follows upstream changes in the `crypto-bigint` and `elliptic-curve` crates: RustCrypto/traits#2153


tarcieri:
Hmm, this is going to need to be done in an additive manner due to circular dependencies

Issue #2196, opened 2026-01-29
cmpv2: PollRepContent is incorrectly encoded


insecure:
Hey all.
I found a bug when trying to parse a `pollRep` message.

[RFC4210 Section 5.3.22](https://datatracker.ietf.org/doc/html/rfc4210#section-5.3.22) `PollRepContent` is defined as follows:
```
    PollRepContent ::= SEQUENCE OF SEQUENCE {
        certReqId    INTEGER,
        checkAfter   INTEGER,  -- time in seconds
        reason       PKIFreeText OPTIONAL }
```
Encoding a `PollRepContent` struct with cmpv2 results in a `SEQUENCE {...}` not `SEQUENCE OF SEQUENCE {...}`.

To reproduce this, I have written the following test in `poll.rs`:
```rust
#[cfg(test)]
mod test_generate_pollrep {
    use der::{
        Encode,
        asn1::{Ia5String, Int},
    };

    use crate::{
        body::PkiBody,
        header::{PkiHeader, Pvno},
        message::PkiMessage,
    };

    #[test]
    fn test_seq_of_seq() {
        let expected_der = [
            48, 36, 48, 22, 2, 1, 2, 130, 6, 115, 101, 110, 100, 101, 114, 130, 9, 114, 101, 99,
            105, 112, 105, 101, 110, 116, 186, 10, 48, 8, 48, 6, 2, 1, 0, 2, 1, 60,
        ];
        let header = PkiHeader {
            pvno: Pvno::Cmp2000,
            sender: x509_cert::ext::pkix::name::GeneralName::DnsName(
                Ia5String::new("sender").unwrap(),
            ),
            recipient: x509_cert::ext::pkix::name::GeneralName::DnsName(
                Ia5String::new("recipient").unwrap(),
            ),
            message_time: None,
            protection_alg: None,
            sender_kid: None,
            recip_kid: None,
            trans_id: None,
            sender_nonce: None,
            recip_nonce: None,
            free_text: None,
            general_info: None,
        };
        let body = PkiBody::PollRep(super::PollRepContent {
            cert_req_id: Int::new(&[0]).unwrap(),
            check_after: 60,
            reason: None,
        });
        let msg = PkiMessage {
            header,
            body,
            protection: None,
            extra_certs: None,
        };

        assert_eq!(msg.to_der().unwrap(), expected_der);
    }
}
```
[...]

PR #2197, opened 2026-02-02
cmpv2: fix bad DER encoding of PollRepContent

insecure:
These changes fix the wrong DER encoding of PollRepContent as described in #2196.


insecure:
Okay, tested it again and it works fine for me now. We now have a proper struct for `PollRepContent` and the basic derives. I also added some simple tests.

PR #2267, opened 2026-03-27
tls_codec: add transparent VLBytes version

kkohbrok:
This PR adds a version of VLBytes called `VLBytesFlat` (I'm open to suggestions for a better name).

In contrast to `VLBytes`, `VLBytesFlat` is serde transparent. It can be used as a drop-in replacement for VLBytes and will deserialize instances of VLBytes correctly. As such, it can be used as a backwards-compatible drop-in.

I didn't make the change directly to `VLBytes` to not break applications such as OpenMLS that use `VLBytes` as keys for lookups.

### DIFF
--- a/tls_codec/src/lib.rs
+++ b/tls_codec/src/lib.rs
@@ -48,9 +48,9 @@ pub use tls_vec::{
     TlsVecU24, TlsVecU32,
 };
 
+pub use quic_vec::{Bytes, VLByteSlice, VLBytes};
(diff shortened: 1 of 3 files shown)


franziskuskiefer:
@kkohbrok did you want to pick this up again at some point?

kkohbrok:
Thanks for the reminder! This has effectively been resolved by #2322.

PR #2280, opened 2026-04-10
pkcs12: add pbeWithSHAAnd3-KeyTripleDES-CBC decryption

MarkAtwood:
## Summary

Adds `EncryptedPrivateKeyInfo::decrypt_3des_cbc`, implementing `pbeWithSHAAnd3-KeyTripleDES-CBC` (OID `1.2.840.113549.1.12.1.3`) decryption for `pkcs8ShroudedKeyBag` entries in legacy PKCS#12 files. This is the scheme used by OpenSSL, pyca/cryptography, and most other tooling when generating PKCS#12 files with the `--legacy` flag.

**New functionality:**
- `src/decrypt.rs`: `EncryptedPrivateKeyInfo::decrypt_3des_cbc(password: &str) -> der::Result<Zeroizing<Vec<u8>>>`
  - Validates OID before touching any key material
  - Derives 24-byte 3DES key (RFC 7292 §B.2, SHA-1, ID=1) and 8-byte IV (ID=2) via the existing KDF
  - Decrypts with 3DES-EDE3-CBC, PKCS#7 unpads
  - All key material and output wrapped in `Zeroizing`
  - Implementation-defined iteration cap (1,000,000) with rationale comment to prevent DoS from crafted .p12 files

**Also fixed:**
- `Pkcs12KeyType` now derives `Copy`, `Clone`, `Debug`, `PartialEq`, `Eq`
- `derive_key_utf8` returns `Err` for `rounds <= 0` (was silent wrong output — inner loop is `for _ in 1..rounds`, which is empty for non-positive values)
- `derive_key` doc example was calling `derive_key_utf8` — replaced with an actual `derive_key` call showing the expected BMP+null-terminator `pass` format
- Renamed `PKCS_12_PBEWITH_SHAAND40_BIT_RC2_CBC` → `PKCS_12_PBE_WITH_SHAAND40_BIT_RC2_CBC` (missing underscore, inconsistent with all five other PBE OID constants) — **breaking change**, happy to add a `#[deprecated]` re-export for one cycle if preferred
- Removed dead commented-out `#[asn1(context_specific = "0", tag_mode = "EXPLICIT")]` from `SafeBag`
- Added `pub` field docs to `EncryptedPrivateKeyInfo` and `Pbes2Params`; removed `#[allow(missing_docs)]` on those structs
- Fixed typos in `kdf.rs` comments: "Concateate" → "Concatenate", "on thy fly" → "on the fly"

## Notes for reviewers

**`cms` dependency pin:** The `cms = "=0.3.0-pre.2"` exact pin is intentional — I preserved the existing pin rather than touching it. It's a coordinated pre-release within this monorepo and will become a normal semver range when both crates reach stable.

**OID rename:** `PKCS_12_PBEWITH_SHAAND40_BIT_RC2_CBC` → `PKCS_12_PBE_WITH_SHAAND40_BIT_RC2_CBC` is a breaking public API rename. The name has a missing underscore that makes it inconsistent with the other five PBE OID constants (`PKCS_12_PBE_WITH_SHAAND128_BIT_RC4`, etc.). Happy to add a `#[deprecated]` re-export for one release cycle if that's preferred.

**`Pbkdf2Params::prf` DEFAULT not handled (pre-existing, not in this PR):** `Pbkdf2Params.prf` is typed `AlgorithmIdentifierOwned` (non-optional) but the PBKDF2 ASN.1 schema declares it `DEFAULT algid-hmacWithSHA1`. Decoding a PBKDF2 params blob that omits `prf` (relying on the default) will fail to parse. This is pre-existing and out of scope here — flagging it in case it isn't already tracked.

## Test plan

- `cargo test --features encryption` — 19 tests, all green
- New tests use independent external oracles:
  - SHA-256 fingerprint of decrypted `PrivateKeyInfo` DER verified against both **pyca/cryptography** and **OpenSSL** (both agree on the same hash; any regression in the KDF or cipher layer will break this)
  - **Bouncy Castle** `PBETest.java` cipher-layer isolation vector — constructs `EncryptedPrivateKeyInfo` in memory, no .p12 parsing, exercises only KDF + 3DES-CBC
  - **pyca/cryptography**-generated EC key fixture (cross-vendor interoperability; different key type and independent implementation)
  - Wrong-password error path tested (PKCS#7 unpadding rejects the bad key)
  - All three RSA fixtures (iter=1, iter=2048, iter=100000) decrypt to identical key bytes


carl-wallace:
One question regarding the changes. Why did you delete PrivateKeyInfo? I may be misremembering but I think that was added due to needing the Sequence trait elsewhere (this is what the comment suggests).

One broader question. Do we want a higher level API and a builder? I just added your test cases to a PKCS 12 builder that is here to demonstrate usage: https://github.com/carl-wallace/pkcs12_builder/blob/main/tests/decrypt_3des.rs. I just added legacy support last month (behind a "legacy" feature).

MarkAtwood:
Thanks for the review!

**On deleting `PrivateKeyInfo`:**

The local `PrivateKeyInfo` (and `Pkcs8Version`) in `safe_bag.rs` was dead code — defined but not imported or used anywhere in the crate. The comment on it acknowledged it duplicated `pkcs8::PrivateKeyInfo`, added only because the upstream type was missing `Debug`/`PartialEq`/`Eq`/`Sequence`. The new tests use `pkcs8::PrivateKeyInfoOwned` directly, which satisfies all those needs via the `alloc` feature. If you know of a specific usage that depended on it being there, I'm happy to restore it — but I'd prefer to point that usage at the pkcs8 type rather than maintain a local copy.

**On a higher-level API and builder:**

Agreed that a higher-level API and builder would be valuable — I'd welcome that as a follow-up. This PR deliberately stops at the low-level `decrypt_3des_cbc` primitive: the security warning in the docs (callers must verify the PFX MAC first) highlights that a real high-level API needs to integrate MAC verification, bag walking, cert extraction, and ideally enforce the verify-before-decrypt ordering — a substantially larger design scope.

I saw your `pkcs12_builder` crate and it's a great demonstration of what that could look like. Would you want to pursue a higher-level API as a follow-up PR to this repo? Happy to help or just stay out of the way depending on how much bandwidth you have.

carl-wallace:
Re: PrivateKeyInfo, I didn't realize it wasn't exported. If it was dead code, OK.

Re: builder crate, I'd be happy for any assistance preparing a PR. I wrote most of that crate a while ago as part of a non-public repo but moved it last month and added the legacy stuff (the lack of which had been forcing use of openssl in a few spots). I'm not particularly wed to the API but would like to retain what it does. ping me at carl <at> redhoundsoftware.com if you want to chat about how we could collaborate on that. 

@tarcieri and @baloo - if you'd rather keep P12 building external to formats please let us know.

tarcieri:
A builder sounds great to me. We have them in other format crates

tarcieri:
@carl-wallace opened #2299 to add an initial builder, so let's get that landed first

PR #2283, opened 2026-04-11
pkcs12: zeroize KDF output and intermediates

MarkAtwood:
## Summary

- Return `Zeroizing<Vec<u8>>` from `derive_key`, `derive_key_bmp`, and `derive_key_utf8` so all derived key material is erased on drop.
- Three buffers now zeroized: the output key (`out`), each per-round intermediate hash (`result`), and the stretched S||P password buffer (`init_key`).  `init_key` uses `Zeroizing` for panic-safe erasure on unwind in addition to an explicit `.zeroize()` on the normal path.
- Enable `digest/zeroize` (zeroes the digest block buffer holding input fragments of id||salt||password) and add `hybrid-array` as a direct optional dep with `features = ["zeroize"]`.  The direct dep is required because `digest/zeroize` does not propagate `hybrid-array/zeroize`, leaving `Array<u8, OutputSize>: Zeroize` unsatisfied.  If `digest` ever adds `hybrid-array/zeroize` to its own `zeroize` feature, the direct dep can be removed.  Both are gated behind the `kdf` feature.
- New tests verified against `openssl kdf PKCS12KDF`: empty UTF-8 password (BMP null-terminates to `[0x00, 0x00]`) and a 72-byte salt spanning two SHA-256 blocks.

## Breaking change

`derive_key`, `derive_key_bmp`, and `derive_key_utf8` now return `Zeroizing<Vec<u8>>` instead of `Vec<u8>`.  This is intentional and is the point of the PR.  Because `Zeroizing<Vec<u8>>` implements `Deref<Target = Vec<u8>>`, most callers that use `&result[..]` or iterate are unaffected.  Callers that stored the result as `Vec<u8>` or compared directly with `[u8; N]` will need to call `.as_slice()` or deref-coerce.  The crate is `0.2.0-pre.0` so this break is within the pre-release versioning policy.

## Dependency notes

- `cms = "=0.3.0-pre.2"` pin is intentional coordinated monorepo pre-release.
- The `hybrid-array` dep is `version = "0.4"`, matching the version already pulled in transitively by `digest`.

## Pre-existing issue (out of scope)

`rounds: i32` — passing `0` or a negative value silently behaves as `rounds = 1` (one hash iteration).  This is pre-existing behavior and is not introduced or changed by this PR.

## Test plan

- [ ] `cargo test --features kdf -p pkcs12` — 6 KDF tests pass, including new empty-password and long-salt cases
- [ ] `cargo clippy --all-features -p pkcs12 -- -D warnings` — clean
- [ ] `cargo hack check --feature-powerset --no-dev-deps -p pkcs12` — all 7 feature combinations compile
- [ ] `RUSTDOCFLAGS="--cfg docsrs -D warnings" cargo +nightly doc --no-deps --all-features -p pkcs12` — clean
- [ ] `cargo +1.85 test --features kdf -p pkcs12` — passes at MSRV

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -167,6 +167,7 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "cdd35008169921d80bc60d3d0ab416eecb028c4cd653352907921d95084790be"
 dependencies = [
  "hybrid-array",
+ "zeroize",
 ]
 
 [[package]]
@@ -547,6 +548,7 @@ dependencies = [
  "const-oid",
  "crypto-common",
  "ctutils",
+ "zeroize",
 ]
 
 [[package]]
@@ -1061,6 +1063,7 @@ dependencies = [
  "der",
  "digest",
  "hex-literal",
+ "hybrid-array",
  "pkcs5",
  "pkcs8",
  "sha2",
--- a/pkcs12/Cargo.toml
+++ b/pkcs12/Cargo.toml
@@ -23,7 +23,13 @@ const-oid = { version = "0.10", features = ["db"], default-features = false }
 cms = { version = "=0.3.0-pre.2", default-features = false }
 
 # optional dependencies
-digest = { version = "0.11", features = ["alloc"], optional = true }
+# "zeroize" feature enables zeroing of the digest's internal block buffer, which
+# holds input fragments of the id||salt||password data between Update calls.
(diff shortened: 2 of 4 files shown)


baloo:
before we proceed, I'd appreciate a reply on https://github.com/RustCrypto/formats/pull/2285#discussion_r3074619778

PR #2285, opened 2026-04-11
fix(pkcs12): validate kdf iteration count >= 1

MarkAtwood:
## Summary

- `derive_key`, `derive_key_bmp`, and `derive_key_utf8` all accept `rounds: i32` with no validation
- The iteration loop `for _ in 1..rounds` silently degenerates to a **single hash application** when `rounds <= 0`, producing wrong output with no error
- RFC 7292 Appendix C defines `iterations INTEGER (1..MAX)` — zero and negative values are out-of-spec

### Fix

Add a `rounds < 1` guard at the top of each function returning `Err(ErrorKind::Value { tag: Tag::Integer })`. `derive_key` and `derive_key_bmp` return types change from `Vec<u8>` to `der::Result<Vec<u8>>` to carry the error.

### Tests

- **`pkcs12_key_derive_rounds_one_boundary`** — verifies `rounds=1` produces the correct output for `EncryptionKey` and `Mac` key types (SHA-256, vectors generated with `openssl kdf … -kdfopt iter:1 … PKCS12KDF` and hardcoded)
- **`pkcs12_key_derive_zero_rounds_errors`** — asserts that `rounds=0`, `rounds=-1`, and `rounds=i32::MIN` all return `Err`

### Breaking change note

`derive_key` and `derive_key_bmp` are public API. Changing their return type from `Vec<u8>` to `der::Result<Vec<u8>>` is a breaking change. Given that the previous behaviour was silently incorrect for `rounds <= 0`, and `pkcs12` is still pre-release (`0.2.0-pre.0`), this seems justified. Happy to discuss alternatives (e.g. `u32` parameter, or a separate validated wrapper).

### DIFF
--- a/pkcs12/src/kdf.rs
+++ b/pkcs12/src/kdf.rs
@@ -15,7 +15,7 @@
 //! [`argon2`](https://docs.rs/argon2) crate for a state-of-the-art password-based KDF.
 
 use alloc::{vec, vec::Vec};
-use der::asn1::BmpString;
+use der::{Tag, asn1::BmpString};
 use digest::{Digest, FixedOutputReset, OutputSizeUser, Update, block_api::BlockSizeUser};
 use zeroize::{Zeroize, Zeroizing};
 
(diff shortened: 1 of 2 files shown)


baloo (approved):

pkcs12/src/kdf.rs
```diff
@@ -72,9 +82,14 @@ where
...
 /// Derives `key` of type `id` from `pass` and `salt` with length `key_len` using `rounds`
-/// iterations of the algorithm
-/// `pass` must be a unicode (utf16) byte array in big endian order without order mark and with two
-/// terminating zero bytes.
+/// iterations of the algorithm.
+///
+/// `pass` must be a Unicode (UTF-16BE) byte array in big-endian order without a byte-order mark
+/// and with two terminating zero bytes.
+///
+/// `rounds` must be ≥ 1 (RFC 7292 Appendix C defines `iterations INTEGER (1..MAX)`).
```
baloo, line 90: The change makes sense to me (and I already merged it), but I can't actually find the text referred here in the RFC7292. Am I missing something?

@MarkAtwood

Issue #2298, opened 2026-04-22
`BitStringRef` that works with `Cow`


tarcieri:
In #1998 the lifetime was removed from `OctetStringRef` by making it a slice newtype, which to simplify looks something like `pub struct OctetStringRef([u8])`, and in practice used as `&OctetStringRef`.

What's really nice about this it means we can `impl Borrow<OctetStringRef> for OctetString` and `ToOwned for OctetStringRef`, and with that we can use `Cow<OctetStringRef>`.

We discussed several times whether it would be possible to do this for `BitStringRef` too and at the time I couldn't figure out a way. I looked into trying to put the unused bits in a tagged pointer but that won't work since we need to construct them from `&[u8]`. However the other place we can stick the information is in the length part of the fat pointer by using it to store the length in bits instead of bytes.

Concretely that might look like `pub struct BitStringRef([UnsafeCell<()>])` where the length of that inner slice represents a number of bits.

Here's a POC of the idea:

https://play.rust-lang.org/?version=stable&mode=debug&edition=2024&gist=e162511e707fe74c8f9f7731972acd32

cc @baloo @dishmaker 

(Thanks to #dark-arts for some help here)

tarcieri:
Open questions: is this actually sound? I saw a suggestion here something similar may not be valid under stacked borrows: https://github.com/ferrilab/bitvec/issues/283#issuecomment-2644347268

I guess I can run my POC by Miri and `cargo careful` to start with.

dishmaker:
I think we're spending too much time and unsafe tricks on the `Cow` issue. Let's just add `BitStringCow` which could be changed to something like:
```rust
type BitStringCow<'a> = Cow<'a, BitStringRef>
```
in later versions of `der`.

And for now it could be an `enum`.

tarcieri:
I think this solves it (I believe `bitvec` already does a similar trick), and really this was the only one I couldn't previously think of a good solution for. I would definitely like to get the second opinion of some experts on the topic though regarding soundness, though `UnsafeCell<()>` is a ZST and therefore slices can't refer to invalid memory so long as the pointer (which is tied to the lifetime of the reference) is valid.

I'd really like to avoid adding bespoke `Cow`-alike types. We can change pretty much all of the string types to work with `Cow` as well.

tarcieri:
Here's a draft PR that implements the basic functionality: #2300

[...]

PR #2300, opened 2026-04-23
[WIP] der: remove lifetime from `BitStringRef` [BREAKING]

tarcieri:
Following the pattern of #1921 and what #1998 did for `OctetStringRef`, this removes the lifetime from the struct, instead changing `BitStringRef` to a proper reference type to be used as `&BitStringRef`, which can implement the `Borrow` and `ToOwned` patterns needed to work with `Cow`.

To make this work, this uses a hack described in #2298 where we abuse a fat pointer, namely a ZST slice `[UnsafeCell<()>]`, to carry the pointer to the `*const u8` buffer that backs the `BitStringRef`, and carries the bit length along as the length of the ZST slice.

To get the original byte length back we can `div_ceil(8)` and then reconstruct the original `[u8]`.

### DIFF
--- a/der/src/asn1/bit_string2.rs
+++ b/der/src/asn1/bit_string2.rs
@@ -1,0 +1,207 @@
+//! ASN.1 `BIT STRING` support.
+
+use crate::{Length, Result, Tag};
+use core::{
+    cell::UnsafeCell,
+    fmt::{self, Debug},
+    ptr, slice,
+};
+
+/// Inaccessible placeholder ZST which is sound to construct in any length.
+///
+/// Using [`UnsafeCell`] prevents the compiler from reasoning about aliasing of the unknown type.
+type Inner = UnsafeCell<()>;
+
+/// ASN.1 `BIT STRING` reference type.
+///
+/// This type contains a sequence of any number of bits.
+///
+/// This is a zero-copy reference type which borrows from the input data.
+#[repr(transparent)]
+pub struct BitStringRef2 {
+    /// Fat pointer which represents the bit string as a slice with a given number of bits.
+    inner: [Inner],
+}
+
+impl BitStringRef2 {
+    /// Create a new ASN.1 `BIT STRING` from a byte slice.
+    ///
+    /// Accepts an optional number of "unused bits" (0-7) which are omitted from the final octet.
+    /// This number is 0 if the value is octet-aligned.
+    ///
+    /// # Errors
+    /// Returns an error if any of the following occur:
+    /// - `unused_bits` is invalid
+    /// - `bytes` is too long
+    /// - an overflow occurred calculating the bit length
+    #[allow(unsafe_code)]
+    pub fn new<'a>(unused_bits: u8, bytes: &'a [u8]) -> Result<&'a Self> {
+        let bits = bit_length(unused_bits, bytes)?;
+
+        // Create a slice that stores the original pointer to `bytes` and `bits` as its length.
+        // SAFETY: `Inner` is a ZST so we can construct slices of any length so long as the pointer
+        // is valid.
+        let slice = unsafe { slice::from_raw_parts::<'a, Inner>(bytes.as_ptr().cast(), bits) };
+
+        // SAFETY: `Self` is a `repr(transparent)` newtype for `[UnsafeCell<()>]`.
+        Ok(unsafe { &*(ptr::from_ref(slice) as *const Self) })
+    }
+
+    /// Create a new ASN.1 `BIT STRING` from the given bytes.
+    ///
+    /// The "unused bits" are set to 0.
+    ///
+    /// # Errors
+    /// Has the same error cases as [`BitStringRef2::new`].
+    pub fn from_bytes(bytes: &[u8]) -> Result<&Self> {
+        Self::new(0, bytes)
+    }
+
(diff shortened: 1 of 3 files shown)


der/src/asn1/bit_string2.rs
```diff
@@ -0,0 +1,207 @@
...
+    ///
+    /// If the number of unused bits is expected to be 0, the [`BitStringRef2::as_bytes`] function
+    /// can be used instead.
+    #[must_use]
+    pub fn raw_bytes(&self) -> &[u8] {
+        // SAFETY: `byte_length` computes the original length of the byte slice this `BitStringRef`
+        // was constructed from, and `inner` contains the original pointer.
+        #[allow(unsafe_code)]
+        unsafe {
+            slice::from_raw_parts(self.inner.as_ptr().cast(), byte_length(self.inner.len()))
```
tarcieri, line 88: This one, it seems, fails under stacked borrows:

https://github.com/RustCrypto/formats/actions/runs/24847701760/job/72739437052#step:5:78

```
error: Undefined Behavior: trying to retag from <177556> for SharedReadOnly permission at alloc64685[0x0], but that tag does not exist in the borrow stack for this location
  --> der/src/asn1/bit_string2.rs:88:13
   |
88 |             slice::from_raw_parts(self.inner.as_ptr().cast(), byte_length(self.inner.len()))
   |             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ this error occurs as part of retag at alloc64685[0x0..0x3]
   |
   = help: this indicates a potential bug in the program: it performed an invalid operation, but the Stacked Borrows rules it violated are still experimental
   = help: see https://github.com/rust-lang/unsafe-code-guidelines/blob/master/wip/stacked-borrows.md for further information
help: <177556> would have been created here, but this is a zero-size retag ([0x0..0x0]) so the tag in question does not exist anywhere
```

@RalfJung I don't suppose you know a way to pull off this particular hack that would actually be compatible with stacked borrows? (trying to make a bitstring-like reference type that can be used with `Borrow` and `ToOwned` patterns so it works with `Cow`) Apparently stacked borrows doesn't like me using a ZST slice like `[UnsafeCell<()>]` so I can store the bit length in a fat pointer.

tarcieri: This issue and comment seem relevant: https://github.com/rust-lang/unsafe-code-guidelines/issues/256#issuecomment-2506767812

tarcieri:
This fails under stacked borrows, but passes with [72211d8](https://github.com/RustCrypto/formats/pull/2300/commits/72211d8ea91905d5543e050e9098d96de8b2cd1a) which enables `-Zmiri-tree-borrows`.

Given this is for a data type used to represent untrusted data that lives at the network edge, I'm disinclined to merge something with soundness that appears somewhat unclear:

- https://github.com/rust-lang/unsafe-code-guidelines/issues/134
- https://github.com/rust-lang/unsafe-code-guidelines/issues/256

It seems stacked borrows and tree borrows disagree about how to handle this scenario and it's unclear to me which one is "right": is stacked borrows being too strict, or is this an edge case in tree borrows that isn't being handled correctly?

That's further compounded by it being a breaking change.

So, I'm going to go ahead and close this, but perhaps we can reconsider something like this in the next breaking release of `der` when perhaps there will be some progress on soundness in the intervening time.

der/src/asn1/bit_string2.rs
RalfJung: I assume the thread you started [on Zulip](https://rust-lang.zulipchat.com/#narrow/channel/136281-t-opsem/topic/.28ab.29using.20ZST.20slices.20for.20references.20with.20.22tagged.22.20lengths/with/590959933) is the same question? Thanks for moving it there, that means other people can help answer when I am swamped (which is my default state these days ;).

RalfJung:
> It seems stacked borrows and tree borrows disagree about how to handle this scenario and it's unclear to me which one is "right": is stacked borrows being too strict, or is this an edge case in tree borrows that isn't being handled correctly?

This is a grey area of the language where no decision has been made yet. If you want to write code that you can be sure will work with all future versions of Rust, then you can't do tricks like this. That's why Stacked Borrows rejects the code.

If you want to write code that should work with *today's* Rust, but that is not covered by Rust's stability guarantees, then you can do tricks like this. That's why Tree Borrows accepts the code. But if you do then be aware that while the code should not be miscompiled by today's rustc, it might well be miscompiled in the future when we narrow down the spec. There's a reason Tree Borrows is not the default yet.

We will likely want to allow the `&Header` pattern in *some* cases. If you want to help make progress on this, I think a proposal which documents that `repr(C)` structs and arrays/slices do *not* have subobject provenance would make a lot of sense.

tarcieri:
I think grey areas are something we probably want to avoid. Even if it so happens to work today I can imagine we would still have to deal with people asking us why it fails under Miri.

Hopefully by the time we're ready to ship the next breaking release of this crate in a year or two there will have been some developments in Rust the language like proper custom DSTs or even just some decision being made about how this is specified so we know there won't be future miscompilation.

RalfJung:
Yeah, hopefully by then someone will step up to drive such developments forwards.

PR #2322, opened 2026-04-29
tls_codec: revert serde_bytes for VLBytes and add VLByteVec

kkohbrok:
- reverts #2022
- adds test that serde roundtrip doesn't change VLBytes variables
- deprecates `VLBytes` and `SecretVLBytes`
- adds new `VLByteVec` and `SecretVLByteVec`
  - use serde_bytes and serde(transparent)
  - are backwards compatible with serialized `VLBytes`

### DIFF
--- a/tls_codec/src/quic_vec.rs
+++ b/tls_codec/src/quic_vec.rs
@@ -11,6 +11,13 @@
 //! to 30-bit values.
 //! This is in contrast to the default behaviour defined by RFC 9000 that allows
 //! up to 62-bit length values.
+
+// `VLBytes` and `SecretVLBytes` are kept around as deprecated types. The
+// internal trait impls for them and their use as building blocks for other
+// items in this module would otherwise emit deprecation warnings at every call
+// site within the crate.
+#![allow(deprecated)]
+
 use super::alloc::vec::Vec;
 use core::fmt;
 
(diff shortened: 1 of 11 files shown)


tls_codec/src/quic_vec.rs
```diff
@@ -327,56 +334,179 @@ impl Size for &VLBytes {
...
+#[cfg_attr(feature = "serde", derive(SerdeSerialize, SerdeDeserialize))]
+#[cfg_attr(feature = "serde", serde(transparent))]
+#[derive(Clone, PartialEq, Eq, Hash, Ord, PartialOrd)]
+pub struct VLByteVec {
+    #[cfg_attr(feature = "serde", serde(serialize_with = "serde_bytes::serialize"))]
+    #[cfg_attr(
+        feature = "serde",
+        serde(deserialize_with = "serde_compat::deserialize_vlbytes_compat")
+    )]
+    vec: Vec<u8>,
```
boxdot, line 361: Nice simplification with `serde(transparent)`.

tls_codec/src/quic_vec.rs
```diff
@@ -327,56 +334,179 @@ impl Size for &VLBytes {
...
+}
+
+impl_vl_bytes_generic!(VLByteVec);
+
+#[cfg(feature = "std")]
+impl Zeroize for VLByteVec {
+    fn zeroize(&mut self) {
+        self.vec.zeroize();
+    }
+}
```
boxdot, line 386: If I am not mistaken, `Zeroize` should be only implemented for `VLSecretByteVec`, right?

boxdot: Ah no. This just allows to zeroize the vec, but does not do it automatically on drop.

boxdot (approved):
Thank you! This is a good change allowing to use the old building blocks and to use the new building blocks with the old data.

tls_codec/src/quic_vec.rs
franziskuskiefer: yep, I'd like to get all the zeroization unified at some point though to make things simpler.

franziskuskiefer (approved):
Apologies for the delay. lgtm, let's get this in.

PR #2323, opened 2026-04-29
codec and x.509 hardening

tob-scott-a:
> [!NOTE]
> This is all being submitted as one branch for discussion. If you prefer each to be individual branches / pull requests, I will decompose this into multiple. I'm submitting this as a draft PR for now. If you're happy with all 3 changes and want them in one go, I'm also happy to flip it into a ready-to-merge PR.

During the course of testing internal Claude skills I'm developing, I tasked them with finding/mitigating side-channels across RustCrypto. It yielded a lot of false positives and not much worth writing about.

Some follow-up work:

* base32ct's `remove_padding` had a TODO;  https://github.com/RustCrypto/formats/commit/4e5982c1d53a5f8f8ff9e8a117546e1c0bc65e38 aspires to address it
* ct-disasm is a new developer tool that aims to do the same kind of analysis as one of Trail of Bits' public skills, but without Python or LLMs involved; https://github.com/RustCrypto/formats/commit/b3f9c3af3de4b9099d6d77dea2cd9bfe85f01664
* https://github.com/RustCrypto/formats/commit/81804a355fb30c5fafa50372c8e596d227c58786 adds the C2SP/x509-limbo test harness to the x509-cert crate to improve test coverage

### DIFF
--- a/dev-tools/ct-disasm/README.md
+++ b/dev-tools/ct-disasm/README.md
@@ -1,0 +1,119 @@
+# ct-disasm — constant-time disassembly check (research preview)
+
+A regression-detection tool for the constant-time hot paths in `base16ct`,
+`base32ct`, and `base64ct`. Compiles a small wrapper crate that drives each
+crate's public encode/decode API on a fixed-length buffer, dumps the
+emitted assembly via `rustc --emit=asm`, and reports per-wrapper
+conditional-branch counts. Designed to be diffed against a snapshotted
+baseline as a CI gate.
+
+## Why this exists
+
+The CT-critical inner functions (`decode_nibble`, `decode_5bits`,
+`decode_6bits`, `is_pad_ct`) are `#[inline(always)]` and not exported.
+Checking them in isolation isn't useful — what matters is what they look
+like *after* LLVM has inlined them into the public API path that real
+callers use. This crate places `#[inline(never)]` `extern "C"` shims around
+those public APIs, with input/output buffers sized as compile-time
+constants, so length-dependent branches in the public API collapse and
+the asm scanner sees the actual inlined CT machinery.
+
+A constant-time-analysis audit of these crates (Trail of Bits) confirmed
+the source-level CT idioms are correct. This tool exists to catch
+**compiler regressions** — a future LLVM that decides to lower one of the
+arithmetic-mask idioms back into a branch.
+
+## Usage
+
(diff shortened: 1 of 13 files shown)


dev-tools/ct-disasm/README.md
```diff
@@ -0,0 +1,119 @@
+# ct-disasm — constant-time disassembly check (research preview)
```
tarcieri, line 1: For something like this it would probably be better to put it in https://github.com/rustcrypto/utils though it does look interesting

dev-tools/ct-disasm/README.md
```diff
@@ -0,0 +1,119 @@
...
+
+A regression-detection tool for the constant-time hot paths in `base16ct`,
+`base32ct`, and `base64ct`. Compiles a small wrapper crate that drives each
```
tarcieri, line 4: Oh hmm, if it's specialized to these I guess it can go here

baloo:
My understanding is that x509-limbo is more geared towards validation logic. x509-cert does not handle validation.

@carl-wallace made a validator (which I recommend and use) over in https://github.com/carl-wallace/rust-pki/tree/main/certval

I'd love if this could be adopted here at some point.
You'll find a branch there that does make use of x509-limbo.

I'd rather see this dropped from here for now.

dev-tools/ct-disasm/check.sh
```diff
@@ -0,0 +1,130 @@
...
+        in_body && /\.cfi_endproc/ { in_body = 0 }
+        in_body { print }
+    ' "$ASM_FILE")
+
+    if [[ -z "$BODY" ]]; then
+        echo "WARN: $W not found in asm" >&2
+        continue
+    fi
+
+    COUNT=$(printf '%s\n' "$BODY" | grep -cE "(^|[[:space:]])($BRANCH_REGEX)([[:space:]]|$)" || true)
```
baloo, line 122:
```suggestion
    COUNT=$(printf '%s\n' "$BODY" | grep -cE "^[[:space:]]*($BRANCH_REGEX)[[:space:]]*$" || true)
```
Am I reading this regexp wrong?

Surely there is a better way to read assembly than a regex anyway? That seems like it's looking for false-positives.

baloo:
thanks for wasting the time of maintainers.

Issue #2320, opened 2026-04-29
Relax `Encoding::encode_string`'s limit to `usize::MAX/4 * 3`?


tisonkun:
`base64` uses the following logic:

```rust
/// Calculate the base64 encoded length for a given input length, optionally including any
/// appropriate padding bytes.
///
/// Returns `None` if the encoded length can't be represented in `usize`. This will happen for
/// input lengths in approximately the top quarter of the range of `usize`.
pub const fn encoded_len(bytes_len: usize, padding: bool) -> Option<usize> {
    let rem = bytes_len % 3;

    let complete_input_chunks = bytes_len / 3;
    // `?` is disallowed in const, and `let Some(_) = _ else` requires 1.65.0, whereas this
    // messier syntax works on 1.48
    let complete_chunk_output =
        if let Some(complete_chunk_output) = complete_input_chunks.checked_mul(4) {
            complete_chunk_output
        } else {
            return None;
        };

    if rem > 0 {
        if padding {
            complete_chunk_output.checked_add(4)
        } else {
            let encoded_rem = match rem {
                1 => 2,
                // only other possible remainder is 2
                // can't use a separate _ => unreachable!() in const fns in ancient rust versions
                _ => 3,
            };
            complete_chunk_output.checked_add(encoded_rem)
        }
    } else {
        Some(complete_chunk_output)
    }
```

While `base64ct` does `checked_mul` at the beginning:

```rust
const fn encoded_len_inner(n: usize, padded: bool) -> Option<usize> {
    match n.checked_mul(4) {
        Some(q) => {
            if padded {
                Some(((q / 3) + 3) & !3)
            } else {
                Some((q / 3) + (q % 3 != 0) as usize)
            }
        }
        None => None,
    }
}
```

Since the Base64 encoding takes generally 4/3 length of the input, is it possible to relax the `Encoding::encode_string`'s limit by reworking calculations?

cc @tarcieri

tarcieri:
Yeah, that logic should definitely be re-evaluated, especially as `div_ceil` is now stable and `const fn`-friendly and can be used to simplify it.

The first case looks like it can just be `n.div_ceil(3).checked_mul(4)`.

Second case looks like it can be simplified too while avoiding overflow.

Edit: I think something like this works?

```rust
match n.div_ceil(3).checked_mul(4) {
    Some(len) if padded => Some(len),
    Some(len) => {
        let rem = n % 3;
[...]

PR #2325, opened 2026-05-02
base64ct: use `div_ceil` to compute encoded length

tarcieri:
The previous logic was using the idiom for `div_ceil` but now that our MSRV is sufficiently high we can now ue it directly.

This also addresses the `arithmetic_side_effects` lint.

Additionally this addresses #2320.

### DIFF
--- a/base64ct/src/encoding.rs
+++ b/base64ct/src/encoding.rs
@@ -359,17 +359,18 @@ fn is_pad_ct(input: u8) -> i16 {
     ((((PAD as i16 - 1) - input as i16) & (input as i16 - (PAD as i16 + 1))) >> 8) & 1
 }
 
-// TODO(tarcieri): explicitly checked/wrapped arithmetic
-#[allow(clippy::arithmetic_side_effects)]
-#[inline(always)]
+/// Compute the encoded length for an `n`-byte input and whether the output is `padded`.
+#[inline]
 const fn encoded_len_inner(n: usize, padded: bool) -> Option<usize> {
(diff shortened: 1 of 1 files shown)


tarcieri:
Spot checked in the playground: https://play.rust-lang.org/?version=stable&mode=debug&edition=2024&gist=1692f04e8fbaf39fa93c78897c0ee059

tisonkun (approved):
LGTM. Thanks!

newpavlov (approved):

base64ct/src/encoding.rs
```diff
@@ -359,17 +359,18 @@ fn is_pad_ct(input: u8) -> i16 {
...
+    match n.div_ceil(3).checked_mul(4) {
+        Some(len) if padded => Some(len),
+        Some(len) => {
+            let rem = n % 3;
+            len.checked_sub(if rem == 0 {
+                0
             } else {
-                Some((q / 3) + (q % 3 != 0) as usize)
-            }
+                3usize.checked_sub(rem).expect("should be in range")
```
tisonkun, line 372: nit: this can be `3 - rem` since `rem` must be 1 or 2 here?

Or even using a `match` for the whole `checked_sub` argument with `unreachable!`.

tarcieri: `3 - rem` doesn't pass the `arithmetic_side_effects` lint.

A match with `unreachable!` is basically the same thing as `expect`

PR #2336, opened 2026-05-13
base32ct: comments and constants

tisonkun:
BTW, when reading https://www.rfc-editor.org/rfc/rfc4648.html#section-6 I noticed that both Base32 encoding should use upper-case alphabet characters. But changing this would be a breaking change, any idea why we use lower-case characters for `Base32`?

### DIFF
--- a/base32ct/src/alphabet.rs
+++ b/base32ct/src/alphabet.rs
@@ -4,9 +4,9 @@ pub(crate) mod rfc4648;
 
 use core::{fmt::Debug, ops::RangeInclusive};
 
-/// Core encoder/decoder functions for a particular Base64 alphabet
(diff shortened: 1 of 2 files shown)


tarcieri:
Note that section opens with: "The Base 32 encoding is designed to represent arbitrary sequences of octets in a form that needs to be case insensitive". I think the lower case alphabet has seen more use in practice, e.g. Tor `.onion` addresses or Tahoe-LAFS capability strings.

`Base32` could potentially be renamed `Base32Lower`.  It would probably be good to then add a `Base32` type which actually implements case insensitive decoding (but rejects mixed case).

tarcieri (approved):

PR #2345, opened 2026-06-05
der: replace O(n^2) SetOf sort with O(n log n) sort

### LINKED ISSUE #2319: `SetOfVec::decode_value` uses O(N²) insertion sort — quadratic DoS on crafted SET OF input
## Summary

`der_sort()` in `der/src/asn1/set_of.rs` is a hand-rolled insertion sort with no element count cap. It is called on every `SetOfVec` decode in DER mode. On a reverse-sorted input of N elements, insertion sort performs `N(N−1)/2` comparisons and swaps — O(N²) total work.

A 246 KB crafted SET OF with 18,000 elements in reverse-sorted DER order causes a single SetOfVec::decode_value call to block for ~11 seconds on a modern machine. At N=9,000 (123 KB payload) it takes ~2.9 seconds, N=18,000 takes ~11s.

## Root cause
```
// der/src/asn1/set_of.rs
fn der_sort<T: DerOrd>(slice: &mut [T]) -> Result<(), Error> {
    for i in 0..slice.len() {
        let mut j = i;
        while j > 0 {
            if slice[j - 1].der_cmp(&slice[j])? == Ordering::Greater {
                slice.swap(j - 1, j);
                j -= 1;
            } else {
                break;
            }
        }
    }
    Ok(())
}
```
This is insertion sort. For N elements in reverse order: element at index k requires k swaps, total = N(N−1)/2 = O(N²).

## Verified timing (der 0.7.10, x509-cert 0.2.5, --release)
Payload: crafted RelativeDistinguishedName (SET OF AttributeTypeAndValue) with N ATVs in reverse-sorted DER order.

| N	| Payload|	Worst-case (ms)|	Best-case (ms)|	Ratio|
| --- |  --- |  --- |  --- |  --- |
| 300|	4.2 KB|	3.2|	0.08|	40×|
|1,000|	14 KB|	35|	0.24|	146×|
|3,000|	42 KB|	316|	0.73|	433×|
|9,000|	123 KB|	2,818|	2.2|	1,281×|
|18,000|	246 KB|	11,414|	4.3|	2,654×|

Consecutive worst-case ratios per 3×-N step are consistently ~9×, confirming O(N²). Best-case (already sorted) is O(N) as expected for insertion sort.

Note: each comparison is allocation-free — DerOrd uses the blanket impl in ord.rs which compares Header structs then delegates to the field-by-field ValueOrd derive. The O(N²) cost is pure computation (function calls + byte comparisons + struct swaps).

## Attack surface

Any SetOfVec<T> decoded from untrusted DER is affected. The two widest-deployed paths:

- x509-cert — RelativeDistinguishedName = SetOfVec<AttributeTypeAndValue>: triggered by any call to Certificate::from_der() on an attacker-supplied certificate (TLS, S/MIME, code signing, …).
- cms — DigestAlgorithmIdentifiers, CertificateSet, SignerInfos are all independent SetOfVec fields in SignedData; a single crafted CMS blob stacks the sort cost across all three.

## This should fix

Replace the insertion sort with slice::sort_unstable_by (introsort, O(N log N)):
```
fn der_sort<T: DerOrd>(slice: &mut [T]) -> Result<(), Error> {
    let mut first_err: Option<Error> = None;
    slice.sort_unstable_by(|a, b| {
        a.der_cmp(b).unwrap_or_else(|e| {
            first_err.get_or_insert(e);
            core::cmp::Ordering::Equal
        })
    });
    first_err.map_or(Ok(()), Err)
}
```
This reduces worst-case sort time from O(N²) to O(N log N) — at N=18,000 that is >100× faster. No change to accepted/rejected inputs or sort semantics. A secondary element count limit (e.g., reject SET OF with > 1024 elements at decode time) could be added as an optional hardening measure but is not required to fix the algorithmic issue.

## Affected versions
Tested in 0.7.x / 0.8.x, might affect all versions.

--- 

### One another finding (maybe feature?
In the same file, the two SET OF types have opposite behaviour on out-of-order DER input:
- SetOfVec::decode_value — silently sorts → accepts non-canonical input
- SetOfRef::from_bytes — rejects with ErrorKind::SetOrdering

Are they work as expected?

---

If I made any mistake, plz let me know. Many thanks.

arpitjain099:
Fixes #2319

## Problem

`der_sort()` in `der/src/asn1/set_of.rs` is a hand-rolled insertion sort that runs on every `SetOfVec` / heapless `SetOf` decode in DER mode. It is O(n^2), and reverse-sorted input forces `n(n-1)/2` comparisons and swaps. A crafted reverse-sorted `SET OF` (about 18k elements in 246 KB) makes a single `SetOfVec::decode_value` take roughly 11 seconds, a denial-of-service vector.

This is reachable from any `SetOfVec<T>` decoded from untrusted DER, including `RelativeDistinguishedName` in x509-cert (`Certificate::from_der`) and `DigestAlgorithmIdentifiers` / `CertificateSet` / `SignerInfos` in cms.

## Fix

Replace the insertion sort with `slice::sort_unstable_by` (in-place O(n log n) introsort), capturing the first `der_cmp` error and returning it after the sort so error bubbling is preserved.

A few correctness points worth calling out:

- `no_std`: `sort_unstable_by` is provided by `core`, so it compiles under `#![no_std]` with no `alloc`. The stable `sort_by` would require `alloc` and break the heapless no-alloc build, so unstable is the right choice here. Verified the heapless feature combination compiles and tests pass.
- Ordering is identical: two elements that compare `Equal` under `DerOrd` have byte-identical DER encodings, so an unstable sort cannot change the serialized output. This is cross-checked in a test against a reference copy of the old insertion sort on both randomized and adversarial inputs.
- Duplicates are preserved: `der_sort` never deduplicated, and that behavior is unchanged (covered by a test).
- Non-canonical input is still accepted and sorted at decode time, unchanged.

## Before / after

Decoding a reverse-sorted `SET OF INTEGER`:

- N=9,000: about 2,818 ms before, 0.35 ms after.
- N=18,000: about 11,414 ms before, 0.68 ms after (roughly a 16,000x improvement).

## Tests

Added `der_sort_matches_reference_ordering` (old insertion sort vs new sort produce identical ordering on randomized and adversarial inputs), `setofvec_decodes_large_reverse_sorted_der`, and `der_sort_preserves_duplicates`. All green across default, `alloc`, `--all-features`, and `--no-default-features --features heapless`. `cargo clippy -p der --all-features` is clean. x509-cert and cms build against the change.

## Out of scope

The issue also notes the `SetOfVec` (sorts) vs `SetOfRef` (rejects out-of-order) behavior difference, which a maintainer already commented on; that is a separate behavior question and is not touched here. I also did not add a hard element-count cap, since the algorithmic fix is the clean behavior-preserving change and a cap would reject currently-accepted inputs (a policy call for maintainers).

### DIFF
--- a/der/src/asn1/set_of.rs
+++ b/der/src/asn1/set_of.rs
@@ -740,28 +740,33 @@ fn check_der_ordering<T: DerOrd>(a: &T, b: &T) -> Result<(), Error> {
 /// Sort a mut slice according to its [`DerOrd`], returning any errors which
 /// might occur during the comparison.
 ///
-/// The algorithm is insertion sort, which should perform well when the input
-/// is mostly sorted to begin with.
+/// Uses [`slice::sort_unstable_by`] (an O(n log n) introsort) rather than the
+/// stable [`slice::sort_by`] because the latter requires `alloc`, while this
+/// function must also work on heapless `no_std` targets where only `core` is
+/// available. Sorting unstably is fine here: two elements that compare
+/// [`Ordering::Equal`] under [`DerOrd`] have identical DER encodings, so their
+/// relative order does not affect the serialized output.
 ///
-/// This function is used rather than Rust's built-in `[T]::sort_by` in order
-/// to support heapless `no_std` targets as well as to enable bubbling up
-/// sorting errors.
-#[allow(clippy::arithmetic_side_effects)]
+/// A previous implementation used a hand-rolled insertion sort. On adversarial
+/// reverse-sorted input that degrades to O(n^2), which let a crafted `SET OF`
+/// turn a single decode into a quadratic-time denial of service (see
+/// <https://github.com/RustCrypto/formats/issues/2319>).
+///
+/// `DerOrd::der_cmp` is fallible. Since the standard sort comparator must
+/// return [`Ordering`] rather than a `Result`, the first comparison error is
+/// captured and returned after the sort completes; on error the slice is left
+/// in an unspecified (but valid) order.
 fn der_sort<T: DerOrd>(slice: &mut [T]) -> Result<(), Error> {
-    for i in 0..slice.len() {
-        let mut j = i;
+    let mut first_err: Option<Error> = None;
 
-        while j > 0 {
-            if slice[j - 1].der_cmp(&slice[j])? == Ordering::Greater {
-                slice.swap(j - 1, j);
-                j -= 1;
-            } else {
-                break;
-            }
-        }
-    }
+    slice.sort_unstable_by(|a, b| {
+        a.der_cmp(b).unwrap_or_else(|err| {
+            first_err.get_or_insert(err);
+            Ordering::Equal
+        })
+    });
 
-    Ok(())
+    first_err.map_or(Ok(()), Err)
 }
 
 #[cfg(feature = "alloc")]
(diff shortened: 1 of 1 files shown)


tarcieri (approved):

der/src/asn1/set_of.rs
```diff
@@ -955,4 +960,161 @@ mod tests {
...
+        } else {
+            let len_bytes = body_len.to_be_bytes();
+            let first = len_bytes.iter().position(|&b| b != 0).unwrap();
+            let trimmed = &len_bytes[first..];
+            // `trimmed.len()` is the count of significant length bytes (<= 8),
+            // so the conversion cannot truncate.
+            der.push(0x80 | u8::try_from(trimmed.len()).expect("length byte count <= 8"));
+            der.extend_from_slice(trimmed);
+        }
+        der.extend_from_slice(&body);
```
dishmaker, line 1094: Shorter version:

```suggestion
        let mut der = elements.to_der();
        // change SEQUENCE tag 0x30 to SET 0x31
        der[0] = 0x31;
```

https://docs.rs/der/0.8.0/src/der/asn1/sequence_of.rs.html#274

tarcieri: @dishmaker want to open a PR?

arpitjain099: Nice, that's cleaner, thanks @dishmaker. Flipping the tag byte directly reads better than what I did, and it lines up with how sequence_of.rs handles it. Happy to open a quick follow-up applying this and credit you, or if you'd rather send it yourself please go ahead. Whichever you prefer.

PR #2348, opened 2026-06-15
tls_codec: use write_all instead of write

keks:
I ran into an issue when writing into a base64 encoder. Debugging this turned up that the base64 encoder does not always write the entire provided buffer at once. In order to write into this cleanly, I just use `write_all` instead of write, and track the number of written bytes manually (i.e. not by inspecting the result of `write` - we now know that the whole slice was written).

I also added a test writer struct that only writes a single byte with each invocation of write_all, and a test that uses that for ensuring this is now handled correctly.

### DIFF
--- a/tls_codec/src/primitives.rs
+++ b/tls_codec/src/primitives.rs
@@ -32,8 +32,7 @@ impl<T: Serialize> Serialize for Option<T> {
     fn tls_serialize<W: Write>(&self, writer: &mut W) -> Result<usize, Error> {
         match self {
             Some(e) => {
-                let written = writer.write(&[1])?;
-                debug_assert_eq!(written, 1);
+                writer.write_all(&[1])?;
                 e.tls_serialize(writer).map(|l| l + 1)
             }
             None => {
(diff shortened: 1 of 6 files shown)


tls_codec/src/primitives.rs
```diff
@@ -156,7 +155,9 @@ macro_rules! impl_unsigned {
             #[cfg(feature = "std")]
             #[inline]
             fn tls_serialize<W: Write>(&self, writer: &mut W) -> Result<usize, Error> {
-                let written = writer.write(&self.to_be_bytes())?;
+                let bytes = &self.to_be_bytes();
+                writer.write_all(bytes)?;
+                let written = bytes.len();
                 debug_assert_eq!(written, $bytes);
```
franziskuskiefer: This shouldn't be able to fail? But also doesn't really hurt.

franziskuskiefer (changes requested):
Thanks! lgtm with a small nit.

The `InvalidWriteLength` is unused now I think. Let's add a comment to that effect (to avoid breakage we can keep it).

tls_codec/tests/encode.rs
```diff
@@ -87,3 +87,28 @@ fn serialize_var_len_boundaries() {
...
+    }
+}
+
+#[test]
+fn serialize_into_short_writer() {
+    let mut short_writer = SingleByteWriter(vec![0u8; 32]);
+
+    let first = [0u8, 1, 2];
+
+    first.tls_serialize(&mut short_writer).unwrap();
```
franziskuskiefer, line 125: We should actually test something.
This is what's happening.

```suggestion
    assert_eq!(short_writer.0.len(), 35);
    assert_eq!(&short_writer.0[short_writer.0.len() - 3..], &[0, 1, 2]);
```

franziskuskiefer (changes requested):
Another nit in the test, then we can merge this.

franziskuskiefer (approved):

PR #2350, opened 2026-06-17
tls_codec: implement `SerializeBytes` for `VLBytes`

wysiwys:
This PR implements `SerializeBytes` for `VLBytes`. Since the `SerializeBytes::tls_serialize()` method is now implemented for `VLBytes` in addition to the existing `Serialize::tls_serialize()` method, this PR also updates the calls to `Serialize::tls_serialize()` in this module to explicitly use the method from the `Serialize` trait.

### DIFF
--- a/tls_codec/src/quic_vec.rs
+++ b/tls_codec/src/quic_vec.rs
@@ -672,7 +672,23 @@ mod rw_bytes {
     impl Serialize for &VLBytes {
         #[inline(always)]
         fn tls_serialize<W: std::io::Write>(&self, writer: &mut W) -> Result<usize, Error> {
-            (*self).tls_serialize(writer)
+            Serialize::tls_serialize(*self, writer)
+        }
+    }
+
+    impl SerializeBytes for VLBytes {
+        #[inline(always)]
+        fn tls_serialize(&self) -> Result<Vec<u8>, Error> {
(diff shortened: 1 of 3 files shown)


tls_codec/src/quic_vec.rs
```diff
@@ -672,7 +672,23 @@ mod rw_bytes {
     impl Serialize for &VLBytes {
         #[inline(always)]
         fn tls_serialize<W: std::io::Write>(&self, writer: &mut W) -> Result<usize, Error> {
-            (*self).tls_serialize(writer)
+            Serialize::tls_serialize(*self, writer)
+        }
+    }
+
+    impl SerializeBytes for VLBytes {
```
franziskuskiefer: Can you add a small test to ensure it works?

franziskuskiefer (changes requested):
Thanks.

This creates an asymmetry for `no_std`. I don't think that's necessary.
You could implement it the way `&[T]` already does it, and move both impls out to a location that is not gated by `std`.

franziskuskiefer (changes requested):
Looks like CI isn't happy.

franziskuskiefer (approved):
Thanks, let's get this in.

Just a note that ` VLByteVec` is replacing `VLBytes` (hence the deprecation).

PR #2351, opened 2026-06-18
tls_codec: impl for String

### LINKED ISSUE #2204: tls: Implement de/serialization for `String`
Right now, you can't derive the de/serialization impls for structs that contain `String` fields, which is annoying. Right now we are wrapping `String` and impl the traits on that type, but it feels like it would be simpler to have this code upstream.

keks:
This implements tls_codec de/serialization for String and &str.

Fixes #2204


tls_codec/src/string.rs
```diff
@@ -0,0 +1,224 @@
+use alloc::string::String;
```
franziskuskiefer: Let's add a doc comment on top here that this is not specified in the TLS presentation language. But that it's what's implicitly done all the time. And then say how the string is actually being serialized (just its bytes).

tls_codec/src/string.rs
```diff
@@ -0,0 +1,224 @@
...
+    }
+}
+
+#[cfg(feature = "std")]
+impl Deserialize for String {
+    fn tls_deserialize<R: std::io::Read>(bytes: &mut R) -> Result<Self, crate::Error>
+    where
+        Self: Sized,
+    {
+        let bytes = TlsByteVecU8::tls_deserialize(bytes)?;
```
franziskuskiefer: This isn't right. You serialize to variable length encoding.
This goes back to the comment on top. We need to clearly describe how strings are being encoded if we want to provider a general implementation.

tls_codec/src/string.rs
```diff
@@ -0,0 +1,224 @@
...
+
+        let buf = SerializeBytes::tls_serialize(&s).unwrap();
+        assert_eq!(buf, [2, 0xC3, 0xBC]);
+        assert_eq!(s.tls_serialized_len(), 3);
+    }
+
+    #[cfg(feature = "std")]
+    #[test]
+    fn roundtrip_deserialize() {
+        let original = String::from("roundtrip test");
```
franziskuskiefer: Using a longer string here would surface the incompatible serialize/deserialize. e.g. it will with `let original = String::from_utf8(vec![0x30; 300]).unwrap();`.

franziskuskiefer (changes requested):
Thanks.
But we need to pick a way to serialize here and describe it so that it's not used when it's not the right encoding.

tls_codec/src/string.rs
keks: Ohh, I thought this would happen by itself. Alright. It looks like VLByteSlice doesn't implement De/SerializeBytes yet, and I would like to impl those as well for nostd support. Should I do that in this PR or a separate one?

tls_codec/src/string.rs
keks: Will do!

tls_codec/src/string.rs
keks: You are right, that fails. I'll add it to ensure this is caught.

keks:
We discussed the next steps offline: 

The new module needs more documentation, especially around what the format used here is. There are multiple possible encodings, but here we are focusing on varint-length encoded buffers containing utf8 bytes. I'll open a followup issue for supporting Strings with sequences of fixed-length integers in the length field.

Strings backed by arrays (i.e. fixed-length byte buffers) will _not_ be supported. Restricting strings (esp with multibyte characters) to fixed byte lengths seems hard to work with.

tls_codec/src/varint.rs
```diff
@@ -168,6 +168,16 @@ impl Serialize for TlsVarInt {
     }
 }
 
+impl SerializeBytes for TlsVarInt {
+    #[inline]
+    fn tls_serialize(&self) -> Result<alloc::vec::Vec<u8>, Error> {
+        let mut bytes = alloc::vec![0u8; 8];
+        let len = self.write_bytes(&mut bytes)?;
+        bytes.truncate(len);
```
franziskuskiefer: If you'd use `bytes_len()` for the length you wouldn't need to truncate. I don't think it makes a real difference though.

tls_codec/src/quic_vec.rs
```diff
@@ -539,6 +539,29 @@ impl Size for VLByteSlice<'_> {
...
+impl SerializeBytes for VLByteSlice<'_> {
+    fn tls_serialize(&self) -> Result<Vec<u8>, Error> {
+        // Get the byte length of the content, make sure it's not too
+        // large and write it out.
+        let content_length = self.0.len();
+
+        let mut len_bytes =
+            SerializeBytes::tls_serialize(&ContentLength::from_usize(content_length)?)?;
+
+        let mut out = alloc::vec::Vec::with_capacity(content_length + len_bytes.len());
```
franziskuskiefer: We should make sure `content_length + len_bytes.len() <= isize::MAX`, or does that happen somewhere already?

tls_codec/src/string.rs
```diff
@@ -0,0 +1,231 @@
...
+
+use crate::{DeserializeBytes, SerializeBytes, Size, VLByteSlice, VLByteVec};
+
+impl Size for String {
+    fn tls_serialized_len(&self) -> usize {
+        self.as_bytes().tls_serialized_len()
+    }
+}
+
+impl Size for &str {
```
franziskuskiefer, line 20: Can you add implementations for `str` as well? Then we can call it on a string directly  `"hello".tls_serialize_detached()`.

franziskuskiefer (changes requested):
Thanks. Generally lgtm with a few nits.

tls_codec/src/quic_vec.rs
keks: There is a check behind an `mls` feature gate that checks something even stronger, but that's all. I added another check in here.

franziskuskiefer (approved):
Thanks :shipit:

PR #2353, opened 2026-06-19
Implement PemLabel for ContentInfo

### LINKED ISSUE #2352: `cms::content_info::ContentInfo` should implement `PemLabel`
[RFC7468](https://www.rfc-editor.org/info/rfc7468/#section-9) specifies a PEM encoding of `ContentInfo` from RFC 5652, with a label of "CMS". Accordingly, it would be helpful if `cms::content_info::ContentInfo` could implement `PemLabel` so that it can be used with `EncodePem::to_pem`and `DecodePem::from_pem`.

```rust
impl der::pem::PemLabel for ContentInfo {
    const PEM_LABEL: &str = "CMS";
}
```

richvdh:
We should be able to encode and decode ContentInfo objects encoded as PEM strings.

Fixes: https://github.com/RustCrypto/formats/issues/2352

### DIFF
--- a/cms/Cargo.toml
+++ b/cms/Cargo.toml
@@ -50,6 +50,7 @@ tokio = { version = "1.45.1", features = ["macros", "rt"] }
 x509-cert = { version = "0.3.0-rc.4", features = ["pem"] }
 
 [features]
+default = ["pem"]
 std = ["der/std", "x509-cert/std", "spki/std"]
 builder = [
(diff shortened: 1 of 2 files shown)


richvdh:
Since this requires enabling the `pem` feature of the `der` crate, I thought it best to put it behind a feature flag.

tarcieri (approved):

PR #2354, opened 2026-06-19
re-enable wasmtime tests

rajatjindal:
I tried reproducing the issue reported in https://github.com/bytecodealliance/actions/issues/21, but it seems to be working as [expected](https://github.com/rajatjindal/actions-test/actions). 

so I am opening this PR to see if this issue still happens for wasmtime. If it does, it will help me identify the root cause of the issue.

Thanks


rajatjindal:
while other workflows are still running, but it seems like the workflow re-enabled in this PR is passing now. 🎉. probably a transient error with GitHub api? 

cc @tarcieri

tarcieri:
@rajatjindal thanks!

PR #2365, opened 2026-07-03
tls_codec: 0.4.3 prep

franziskuskiefer:
Release prep for the next tls_codec release.
This also fixes a couple more issues (see changelog) and adds some more fuzzers.


tls_codec/src/quic_vec.rs
```diff
@@ -148,11 +164,13 @@ impl<T: SerializeBytes> SerializeBytes for &[T] {
...
         // This requires more computations but the other option would be to buffer
         // the entire content, which can end up requiring a lot of memory.
-        let content_length = self.iter().fold(0, |acc, e| acc + e.tls_serialized_len());
+        let content_length = self.iter().try_fold(0usize, |acc, e| {
+            crate::checked_len_add(acc, e.tls_serialized_len())
+        })?;
         let length = ContentLength::from_usize(content_length)?;
         let len_len = length.0.bytes_len();
 
-        let mut out = Vec::with_capacity(content_length + len_len);
```
wysiwys, line 155: If the total serialized content length is `isize::MAX`, this vector allocation could panic.  I think this could use a check first that `content_length + len_len` <= `isize::MAX`.

tls_codec/src/lib.rs
```diff
@@ -123,6 +123,88 @@ impl From<std::io::Error> for Error {
...
+/// Unlike `vec![0u8; len]` followed by `read_exact`, this does **not** eagerly
+/// allocate `len` bytes up front: the initial allocation is capped so that a
+/// bogus (large) length field in untrusted input can't trigger a huge
+/// allocation before any bytes are read. The vector grows as data actually
+/// arrives.
+///
+/// Returns [`Error::EndOfStream`] if the reader is exhausted before `len` bytes
+/// are read.
+#[cfg(feature = "std")]
+fn read_bytes_bounded<R: std::io::Read>(reader: &mut R, len: usize) -> Result<Vec<u8>, Error> {
```
wysiwys, line 144: A nit on the parameter name: I think `len` could be renamed to `expected_len` or similar, to be more descriptive

tls_codec/src/tls_vec.rs
wysiwys, line 45: It looks like this may panic on 16-bit platforms where `$size` is `u32`

tls_codec/src/lib.rs
```diff
@@ -123,6 +123,88 @@ impl From<std::io::Error> for Error {
...
+    const MAX_PREALLOC: usize = 4096;
+
+    // Cap the initial allocation. `read_to_end` grows the vector as data
+    // actually arrives, and `Take` bounds the reader to `len` so growth can
+    // never exceed the request.
+    let mut result = Vec::with_capacity(core::cmp::min(len, MAX_PREALLOC));
+
+    // `read_to_end` reads directly into the vector's spare capacity and retries
+    // `ErrorKind::Interrupted` internally, unlike a bare `read` loop.
+    reader.take(len as u64).read_to_end(&mut result)?;
```
wysiwys, line 156: Since this method is used for types that are internally represented as byte vectors, and the bytes are read here into a byte vector,  the size of the serialized data cannot exceed `isize::MAX`. I think that before reaching this line, an initial check that `len` does not exceed `isize::MAX` would be useful in order to rule out invalid lengths early on

tls_codec/CHANGELOG.md
```diff
@@ -7,12 +7,20 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
 - [#1656](https://github.com/RustCrypto/formats/pull/1656) Add `TlsVarInt` type for variable-length integers.
 
 ### Fixed
 - [#2348](https://github.com/RustCrypto/formats/pull/2348) Use `write_all` everywhere instead of write to prevent partial writes from going undetected. The `Error::InvalidWriteLength` variant is deprecated as it is no longer returned.
+- [#XXXX](https://github.com/RustCrypto/formats/pull/XXXX) Element-vector deserialization (`Vec<T>`, `TlsVecU*<T>`, `SecretTlsVecU*<T>`), for both the `Deserialize` (`std::io::Read`) and `DeserializeBytes` implementations, now measures actual byte consumption instead of relying on `tls_serialized_len()` and enforces the declared length exactly. This makes the two implementations agree for non-canonical inner encodings (e.g. non-minimal varint lengths) and rejects input whose elements overshoot the declared vector boundary.
```
wysiwys:
```suggestion
- [#2365](https://github.com/RustCrypto/formats/pull/2365) Element-vector deserialization (`Vec<T>`, `TlsVecU*<T>`, `SecretTlsVecU*<T>`), for both the `Deserialize` (`std::io::Read`) and `DeserializeBytes` implementations, now measures actual byte consumption instead of relying on `tls_serialized_len()` and enforces the declared length exactly. This makes the two implementations agree for non-canonical inner encodings (e.g. non-minimal varint lengths) and rejects input whose elements overshoot the declared vector boundary.
```

tls_codec/CHANGELOG.md
```diff
@@ -7,12 +7,20 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
+  - Element vectors now reject zero-length elements that would otherwise never advance the read cursor, preventing an infinite loop and unbounded allocation on malicious input.
+  - Byte-vector deserialization (`TlsByteVecU*`, `VLBytes`, `VLByteVec`) now uses `checked_add` when computing the content range, returning `Error::InvalidVectorLength` instead of overflowing `usize` on targets where the length field is as wide as the pointer width.
```
wysiwys: `checked_add()` is not yet used in the `Deserialize` impls for `VLBytes` and `VLByteVec`, right?

tls_codec/derive/src/lib.rs
```diff
@@ -761,6 +761,35 @@ fn define_discriminant_constants(
...
+/// Builds an expression that sums `base` and all `terms` into a `usize`,
+/// guarding against overflow only on targets where it can actually occur.
+///
+/// On 64-bit targets every length is bounded by addressable memory
+/// (`isize::MAX`), so the sum can't overflow `usize`; a plain, branch-free
+/// addition is emitted to keep the serialization hot path free of checks. On
+/// narrower targets the serialized form of a large structure can carry enough
+/// length-prefix / discriminant overhead to exceed `usize::MAX`, so we saturate
+/// rather than silently wrapping to a small value (a wrapped length would be
+/// written to the wire as a truncated, mismatched length prefix).
```
wysiwys, line 773: I wonder if it would be better to panic in the derive macro instead, since saturating could cause an incorrectly short length to be written. This would be caught by a debug assertion in `impl_serialize()`, but otherwise the length does not appear to be checked

wysiwys (commented):
Added some notes, mainly on checks and avoiding overflows. Mainly only relevant on platforms where the serialized struct's size is more likely to reach `usize::MAX` (but this seems like it would almost never occur in practice)

tls_codec/src/tls_vec.rs
franziskuskiefer: yes that may happen. We don't cover 16-bit.

tls_codec/derive/src/lib.rs
franziskuskiefer: Panicking is not what we want here. But I added another check in the generated serialize.

wysiwys: Ah indeed, taking another look there isn't a way to check at this stage that the length is too long

PR #2369, opened 2026-07-03
der: fix impl `EncodeValue` for `&SequenceRef`

dishmaker:
Fixes `&SequenceRef` to match `&OctetStringRef`:
https://github.com/RustCrypto/formats/blob/de2aa55dfadbe7b3c1e7a1e5baf0574cd90e6c2f/der/src/asn1/octet_string.rs#L85

Adds test for `SequenceRef` from:
- #1976

### DIFF
--- a/der/src/asn1/sequence.rs
+++ b/der/src/asn1/sequence.rs
@@ -78,7 +78,7 @@ impl<'a> DecodeValue<'a> for &'a SequenceRef {
     }
 }
 
-impl EncodeValue for SequenceRef {
+impl EncodeValue for &SequenceRef {
     fn value_len(&self) -> Result<Length> {
         Ok(self.body.len())
     }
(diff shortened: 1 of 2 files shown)


baloo:
I'm not sure about this, that feels like a bug in the der_derive to me.

dishmaker:
Even the constructor returns `&Self`:

https://github.com/RustCrypto/formats/blob/f362dac4c7128776e93c1a243ed071744fe6274f/der/src/asn1/sequence.rs#L43

https://github.com/RustCrypto/formats/blob/f362dac4c7128776e93c1a243ed071744fe6274f/der/src/asn1/octet_string.rs#L25

tarcieri:
Yeah, due to the blanket impls the ergonomics of reference types leave a bit to be desired.

That's a wart we can potentially smooth over in the next breaking release, possibly adding blanket impls for e.g. `Borrow` rather than concrete types to abstract over the discrepancy.

PR #2379, opened 2026-07-09
der: cap maximum number of nested messages at 64

tarcieri:
Via a privately disclosed report, if a `Decode` implementation decides to unboundedly recurse through nested messages, it can potentially exhaust the stack.

Note that this requires a specially written `Decode` implementation which utilizes such recursion, and isn't a general problem with the way `der` implements message decoding. This instead provides a built-in limit to ensure where such implementations exist they do not exhaust the stack.

### DIFF
--- a/der/src/reader/position.rs
+++ b/der/src/reader/position.rs
@@ -10,14 +10,21 @@ pub(super) struct Position {
 
     /// Position in the input buffer (in bytes after Base64 decoding).
     position: Length,
+
+    /// Number of nested productions encountered (i.e. depth in the call stack).
+    depth: usize,
 }
 
 impl Position {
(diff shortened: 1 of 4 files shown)


der/src/reader/position.rs
```diff
@@ -10,14 +10,21 @@ pub(super) struct Position {
...
     /// Position in the input buffer (in bytes after Base64 decoding).
     position: Length,
+
+    /// Number of nested productions encountered (i.e. depth in the call stack).
+    depth: usize,
 }
 
 impl Position {
+    /// Maximum number of nested messages we tolerate (prevents stack exhaustion).
+    const MAX_DEPTH: usize = 64;
```
tarcieri, line 20: I didn't expose this yet, but it might be a good idea

PR #2396, opened 2026-08-06
x509-cert: fix the IssuingDistributionPoint associated OID

### LINKED ISSUE #2389: IssuingDistributionPoint has the wrong AssociatedOid (id-pe-subjectInfoAccess)
`x509-cert/src/ext/pkix/crl/dp.rs`:

```rust
impl AssociatedOid for IssuingDistributionPoint {
    const OID: ObjectIdentifier = ID_PE_SUBJECT_INFO_ACCESS;   // 1.3.6.1.5.5.7.1.11
}
```

`issuingDistributionPoint` is 2.5.29.28 (RFC 5280 §5.2.5). 1.3.6.1.5.5.7.1.11 is `id-pe-subjectInfoAccess`, which `ext/pkix/access.rs` already binds to `SubjectInfoAccessSyntax`.

So a lookup by associated OID never finds an IDP. It searches for 1.3.6.1.5.5.7.1.11 while the CRL carries 2.5.29.28, and comes back with "no such extension" rather than an error. If you're reading `indirect_crl` to decide whether a CRL's entries apply to your own issuer's serial space, you'll treat an indirect CRL as a direct one.

`impl_extension!(IssuingDistributionPoint, critical = true)` uses the same constant, so the encode side writes the wrong OID too.

Present in 0.3.0 and on master at `b1e1582`.

## Repro

```rust
use const_oid::{AssociatedOid, db::rfc5280::ID_CE_ISSUING_DISTRIBUTION_POINT};
use x509_cert::ext::pkix::IssuingDistributionPoint;

assert_eq!(IssuingDistributionPoint::OID, ID_CE_ISSUING_DISTRIBUTION_POINT);   // fails
```

## Fix

```diff
-    const OID: ObjectIdentifier = ID_PE_SUBJECT_INFO_ACCESS;
+    const OID: ObjectIdentifier = ID_CE_ISSUING_DISTRIBUTION_POINT;
```

`ID_CE_ISSUING_DISTRIBUTION_POINT` is already in `const_oid::db::rfc5280`.

I have this locally with two regression tests in `tests/crl.rs`, one for the constant and one that puts a real IDP on the wire and finds it by associated OID. Both fail before, pass after, and the rest of the suite is unchanged (78 tests plus 10 doc-tests). The 5 failures in `tests/builder.rs` are `zlint: command not found` and happen on unmodified master too.

It's a behaviour change for anyone already working around it, so probably wants a CHANGELOG entry. Happy to send a PR.

eyusd:
Fixes #2389.

`IssuingDistributionPoint` was bound to `ID_PE_SUBJECT_INFO_ACCESS` (1.3.6.1.5.5.7.1.11). That is `id-pe-subjectInfoAccess`, a certificate extension, and `ext/pkix/access.rs` already binds it to `SubjectInfoAccessSyntax`. `issuingDistributionPoint` is 2.5.29.28 (RFC 5280 §5.2.5).

Two tests in `tests/crl.rs`: one on the constant, one that puts an IDP with `indirectCRL` set on the wire and locates it by associated OID. Both fail before this change and pass after.

`cargo test -p x509-cert --all-features --no-fail-fast` goes from 77 tests to 79, doc-tests unchanged at 10. The 5 failures in `tests/builder.rs` are `zlint: command not found` and are identical on unmodified master, since I don't have zlint installed.

`impl_extension!` uses the same constant, so this also changes the OID written on encode. That makes it a behaviour change for anyone who has worked around the old value.

### DIFF
--- a/x509-cert/tests/crl.rs
+++ b/x509-cert/tests/crl.rs
@@ -1,5 +1,16 @@
-use der::Decode;
-use x509_cert::{certificate::Rfc5280, crl::CertificateList};
+use const_oid::{
+    AssociatedOid,
+    db::rfc5280::{ID_CE_ISSUING_DISTRIBUTION_POINT, ID_PE_SUBJECT_INFO_ACCESS},
+};
+use der::{Decode, Encode, asn1::OctetString};
+use x509_cert::{
+    certificate::Rfc5280,
+    crl::CertificateList,
+    ext::{
+        Extension,
+        pkix::{IssuingDistributionPoint, SubjectInfoAccessSyntax},
+    },
+};
 
 #[test]
 fn decode_crl() {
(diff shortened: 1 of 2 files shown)


x509-cert/tests/crl.rs
```diff
@@ -15,3 +26,50 @@ fn decode_crl() {
...
+        extn_value: OctetString::new(idp.to_der().unwrap()).unwrap(),
+    };
+
+    let extensions = vec![ext];
+    let found = extensions
+        .iter()
+        .find(|e| e.extn_id == IssuingDistributionPoint::OID)
+        .expect("IDP extension located by IssuingDistributionPoint::OID");
+    let decoded = IssuingDistributionPoint::from_der(found.extn_value.as_bytes()).unwrap();
+    assert!(decoded.indirect_crl);
```
carl-wallace, line 74: It might be good to add a round trip encode/decode cycle and/or a test that uses an example read from a CRL. There're are a few in the PKITS data set, including https://github.com/carl-wallace/rust-pki/blob/main/certval/tests/examples/PKITS_data_2048/crls/indirectCRLCA1CRL.crl.

tarcieri:
Note this is failing clippy and rustfmt

PR #2401, opened 2026-08-11
der: error on nested trailing data

### LINKED ISSUE #2391: der: read_nested may no longer require complete consumption of nested content
While reviewing the changes between `der` 0.8.0 and 0.8.1, I noticed that `SliceReader::read_nested` in `src/reader/slice.rs` no longer appears to call `nested_reader.finish()` after the callback returns.

Previously, `finish()` ensured that the callback consumed the complete contents of the nested value. In the new implementation, the outer reader is restored after the callback, but I could not find an equivalent check for unconsumed bytes in the nested reader.

This appears to mean that a callback could return `Ok` after consuming only part of the nested value, with the remaining bytes being silently skipped. That could weaken strict DER validation and potentially allow malformed nested values containing trailing data to be accepted.

tarcieri:
This fixes a regression introduced in #2378 which would silently ignore trailing data inside of nested DER messages.

Closes #2391

### DIFF
--- a/der/src/reader/slice.rs
+++ b/der/src/reader/slice.rs
@@ -99,8 +99,20 @@ impl<'a> Reader<'a> for SliceReader<'a> {
 
         let resumption = self.position.split_nested(len)?;
         let ret = f(self);
+        let finished = self.is_finished();
+        let decoded = self.position.current();
+        let remaining = self.remaining_len();
+
         self.bytes = bytes;
         self.position.resume_nested(resumption);
+
+        if ret.is_ok() && !finished {
+            self.failed = true;
+            return Err(self
+                .error(ErrorKind::TrailingData { decoded, remaining })
+                .into());
+        };
+
         ret
     }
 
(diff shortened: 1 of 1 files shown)


der/src/reader/slice.rs
```diff
@@ -217,4 +229,26 @@ mod tests {
...
+    fn nested_trailing_data() {
+        let der = hex!("0102");
+        let mut reader = SliceReader::new(&der).unwrap();
+
+        let err: Error = reader
+            .read_nested(2u8.into(), |reader| {
+                reader.read_slice(1u8.into())?;
+                Ok(())
+            })
+            .unwrap_err();
```
dishmaker:
```suggestion
            .expect_err("read_nested should return Err when the callback did not consume the complete contents of the nested value");
```

karalabe:
Any chance of cutting a release with this one? Kind of borks the latest release.

PR #2408, opened 2026-08-24
`pkcs5`: add `belt-kwp` support, tweak iv parsing in pbes2

makavity:
Need https://github.com/RustCrypto/key-wraps/pull/98 to be merged, before this can be merged


pkcs5/src/pbes2.rs
```diff
@@ -515,40 +541,55 @@ impl<'a> Decode<'a> for EncryptionScheme {
...
+/// OCTET STRING parameters.
+fn decode_iv<const N: usize>(params: Option<AnyRef<'_>>) -> der::Result<[u8; N]> {
+    params
+        .ok_or_else(|| Tag::OctetString.value_error())?
+        .decode_as::<&OctetStringRef>()?
+        .as_bytes()
+        .try_into()
+        .map_err(|_| Tag::OctetString.value_error().into())
+}
+
```
makavity, line 554: Hey, @tarcieri!
What do you think about that approach?

makavity: Maybe I should separate it into another PR?

PR #2410, opened 2026-08-25
pkcs5: support for generic `aead` encrypt/decrypt 

makavity:
Changed `gcm_encrypt` and `gcm_decrypt` to generic AEAD version.
After https://github.com/RustCrypto/formats/pull/2408 i'm planning to add `belt-dwp` support and faced with the duplicate of `gcm_*` methods.

### DIFF
--- a/pkcs5/Cargo.toml
+++ b/pkcs5/Cargo.toml
@@ -16,6 +16,7 @@ edition = "2024"
 rust-version = "1.85"
 
 [dependencies]
+aead = { version = "0.6", default-features = false }
 der = { version = "0.8", features = ["oid"] }
 spki = "0.8"
 
--- a/pkcs5/Cargo.toml
+++ b/pkcs5/Cargo.toml
@@ -16,7 +16,7 @@ edition = "2024"
 rust-version = "1.85"
 
 [dependencies]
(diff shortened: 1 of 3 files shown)


makavity:
Also, unblocks this TODO's: https://github.com/RustCrypto/formats/blob/fe4004b7db7aae8afd3c7627dc3b894a3e353c00/pkcs5/src/pbes2.rs#L350
https://github.com/RustCrypto/formats/blob/fe4004b7db7aae8afd3c7627dc3b894a3e353c00/pkcs5/src/pbes2.rs#L522

pkcs5/Cargo.toml
```diff
@@ -16,6 +16,7 @@ edition = "2024"
 rust-version = "1.85"
 
 [dependencies]
+aead = { version = "0.6", default-features = false }
```
tarcieri: This PR looks good aside from making this a mandatory dependency. Can you feature gate its use and only pull it in when features for AEAD algorithms are enabled?

```suggestion
aead = { version = "0.6", default-features = false, optional = true }
```

makavity: Sure. Forgot, what `pbes2` is a feature.

tarcieri (approved):

Issue #2385, opened 2026-07-23
pkcs1: blockers for v0.8.0 release?


erickt:
Thanks for all the hard work on RustCrypto! I noticed that `pkcs1` has not yet cut a full `0.8.0` release, just `0.8.0-rc.4` as of writing. Are there known issues with the release candidate that we might not want to put it into production yet? Thanks so much!

tarcieri:
The description is sort of spread across #1117 and #1611, but I'd like to refactor `pkcs1` to follow the `*Owned`/`*Ref` pattern, similar to what we've done in `spki` and `pkcs8`

PR #2411, opened 2026-08-25
pkcs1: split der objects into `*Ref` and `*Owned`

dishmaker:
Generic `Ref` and `Owned` aliases of `der` objects

- [x] `pkcs1` progress:
  - [x] `OtherPrimeInfo`
  - [x] `RsaOaepParams`
  - [x] `RsaPrivateKey`
  - [x] `RsaPssParams`
  - [x] `RsaPublicKey`

Related issue:
- #2385

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -61,6 +61,10 @@ x509-tsp = { path = "./x509-tsp" }
 x509-cert = { path = "./x509-cert" }
 x509-ocsp = { path = "./x509-ocsp" }
 
+# RSA: refactor: use *Ref (or *Owned) der types from RustCrypto/formats - #706
+rsa = { git = "https://github.com/dishmaker/RSA.git", rev = "96c97ec3e8b2bedde44f5954334618bec87a19c3" }
+
+
 [workspace.lints.clippy]
 borrow_as_ptr = "warn"
 cast_lossless = "warn"
--- a/pkcs1/src/params.rs
+++ b/pkcs1/src/params.rs
@@ -1,13 +1,15 @@
 //! PKCS#1 RSA parameters.
 
 use crate::{Error, Result};
+#[cfg(feature = "alloc")]
+use der::Any;
 use der::{
     Decode, DecodeValue, Encode, EncodeValue, FixedTag, Length, Reader, Sequence, Tag, TagMode,
     TagNumber, Writer,
     asn1::{AnyRef, ContextSpecificRef, ObjectIdentifier},
     oid::AssociatedOid,
 };
-use spki::{AlgorithmIdentifier, AlgorithmIdentifierRef};
+use spki::{AlgorithmIdentifier, AlgorithmIdentifierRef, AsAlgorithmIdentifierRef};
 
 const OID_SHA_1: ObjectIdentifier = ObjectIdentifier::new_unwrap("1.3.14.3.2.26");
 const OID_MGF_1: ObjectIdentifier = ObjectIdentifier::new_unwrap("1.2.840.113549.1.1.8");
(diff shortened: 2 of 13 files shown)


tarcieri:
> Is UintType generic parameter name okay?

I generally use `U` for that sort of thing

Cargo.toml
```diff
@@ -61,6 +61,10 @@ x509-tsp = { path = "./x509-tsp" }
 x509-cert = { path = "./x509-cert" }
 x509-ocsp = { path = "./x509-ocsp" }
 
+# RSA: refactor: use RsaPublicKeyRef instead of RsaPublicKey from RustCrypto/formats - #706
+rsa = { git = "https://github.com/RustCrypto/RSA.git", rev = "dc89967d5acf570f35375a9eb9151380fabd2f28" }
```
tarcieri: Maybe point this at your repo/branch rather than pinning the commit and relying on GitHub's weird mirroring behavior. But I'm also unclear why https://github.com/RustCrypto/RSA/pull/706 is green but this PR isn't...

dishmaker: OK

dishmaker:
I had to add `UintLike` trait in `der`

tarcieri:
I'd prefer not to add additional traits in this PR, particularly ones unrelated to `pkcs1` like `IntLike`.

Any reason it can't just be `AsRef<[u8]>`? I mean, it looks like we'd need to add impls, but that seems more straightforward than trying to shoehorn multiple new traits into this PR.

If we really need to add `UintLike` I'd prefer it be its own PR.

tarcieri:
I guess the main problem is there's a divergence between `Uint(Ref)::to_bytes` and its underlying encoding due to the leading zero stripping

dishmaker:
I think `AsRef<[u8]>` for `Uint` would be unintuitive, because of leading zeros/ones.

But I can add `AsRef<[u8]>` if you'd like.

tarcieri:
No, that doesn't seem like a good solution either.

One thing that's annoying is `UintRef` didn't get the same treatment as e.g. `OctetStringRef` in that it still has a lifetime, which makes writing other bounds annoying (e.g `AsRef<UintRef>` would be nice here).

So I guess we can do `UintLike` for now, with a TODO to eventually get rid of it and replace it with `AsRef<UintRef>`. I'd probably prefer it actually exposed the full `UintRef` rather than just its bytes, and be added in its own separate PR first.

dishmaker:
Edit: I misunderstood you wanted
```rust
fn as_uint_ref<'a>(&'a self) -> &UintRef<'a>;
```

~Should I:~
- ~remove the `.len()` and `.is_empty()` from `UintRef`,~
- ~or leave it in both `UintRef` and `UintLike` trait?~

tarcieri:
I think `UintLike` should just be a helper to retrieve a `UintRef` which can be replaced by `AsRef<UintRef>` in the future when we can make a breaking change from `&UintRef<'a>` to `&'a UintRef`.

I don't think there should be a need to add/remove any other methods.

pkcs1/src/params.rs
```diff
@@ -138,65 +180,59 @@ impl<'a> RsaPssParams<'a> {
...
 }
 
-impl<'a> DecodeValue<'a> for RsaPssParams<'a> {
+impl<'a, Params> DecodeValue<'a> for RsaPssParams<Params>
+where
+    AlgorithmIdentifier<Params>: DecodeValue<'a, Error = der::Error> + FixedTag,
+    AlgorithmIdentifier<AlgorithmIdentifier<Params>>:
+        DecodeValue<'a, Error = der::Error> + FixedTag,
+    AlgorithmIdentifier<Params>: From<AlgorithmIdentifierRef<'static>>,
+    Params: 'a,
```
tarcieri, line 205: Oof

dishmaker: Im not sure about `Params: 'a` tho

tarcieri (approved):

Issue #1585, opened 2024-10-28
const-oid: panic unexpected, parser bugs


dishmaker:
`ObjectIdentifier::new` should return Err with invalid data, but returns Ok. 

Then, `Display::fmt` panics at `const-oid/src/lib.rs:257`.

```toml
# bug exists in 0.9.5 also
const_oid = { version = "0.10.0-rc.2" }
```

```rust
#[test]
fn test_oid_arc() {
    // Result<ObjectIdentifier>
    // return Ok(..)
    let oid = ObjectIdentifier::new("1.2.840.10045.4.3.4.1111111111");

    // prints oid: Ok(ObjectIdentifier(
    println!("oid: {oid:?}");
    // panics: OID malformed: ArcTooBig
    // const-oid/src/arcs.rs:116:25
}
```

dishmaker:
```rust
#[test]
fn test_oid_roundtrip() {
    let input_str = "0.1.1.1.1.222222222222";

    // debug panic: attempt to multiply with overflow
    let oid = ObjectIdentifier::new(input_str).expect("oid to parse");

    println!("oid: {oid}");
    let output_str = oid.to_string();
    
    // release bug: assertion `left == right` failed
    // left: "0.1.1.1.1.3178890126"
    // right: "0.1.1.1.1.222222222222"
    assert_eq!(output_str, input_str);
}
```

dishmaker:
```rust

#[test]
fn test_oid_dots() {
    let input_str = "0.1..1.1";
    let output_str = ObjectIdentifier::new(input_str)
        .expect("oid to parse")
        .to_string();

    // assertion `left == right` failed
    //  left: "0.1.0.1.1"
    // right: "0.1..1.1"
    assert_eq!(output_str, input_str);
}
```

dishmaker:
```rust
#[test]
fn test_oid_not_overflow() {
    let max: i32 = 128_i32.pow(4) - 1;
    // 0.1.268435455
    let input_str = format!("0.1.{max}");

    let output_str = ObjectIdentifier::new(&input_str)
        .expect("oid to parse")
        .to_string();

    assert_eq!(output_str, input_str);
}

#[test]
#[should_panic(expected = "oid should not parse")]
fn test_oid_overflow() {
    let max: i32 = 128_i32.pow(4) + 1;
    let input_str = format!("0.1.{max}");

    let output_str = ObjectIdentifier::new(&input_str)
        .expect("oid should not parse")
        .to_string();

[...]

PR #2412, opened 2026-08-26
const-oid: reject over-u32 arcs in BER decoder instead of truncating

jaideeppyne:
`Arcs::try_next` (used by `ObjectIdentifier::from_bytes`) accumulated base-128 arc digits with an unchecked `result << 7`, guarded only by `arc_bytes > ARC_MAX_BYTES`. Since a `u32` arc can span up to five base-128 bytes, that guard never fires for a five-byte arc, so the final shift can push the value past `u32::MAX` and silently truncate it.

As a result `from_bytes` accepted encodings denoting arcs greater than `u32::MAX` and returned a wrong value instead of `ArcTooBig`:

- `2A 90 80 80 80 00` (arc 2³²) decoded to `1.2.0`
- `2A 90 80 80 80 05` (arc 2³²+5) decoded to `1.2.5`

This is the byte-decode counterpart of the base-10 parse overflow fixed for #1585/#1592, which only touched `parser.rs`.

Fix: accumulate with `checked_mul`/`checked_add` and return `ArcTooBig` on overflow. The largest in-range arc (`u32::MAX`, `2A 8F FF FF FF 7F`) still decodes correctly. Adds a regression test.

### DIFF
--- a/const-oid/src/arcs.rs
+++ b/const-oid/src/arcs.rs
@@ -73,15 +73,9 @@ impl<'a> Arcs<'a> {
                         Some(byte) => {
                             arc_bytes = checked_add!(arc_bytes, 1);
 
-                            // Accumulate the base 128 digits with overflow checking.
-                            //
-                            // An arc whose value does not fit in `Arc` (`u32`) is
-                            // rejected as `ArcTooBig` rather than being silently
-                            // truncated by the `<< 7` shift. Note that a `u32` arc can
-                            // be up to five base 128 bytes long, so checking the number
-                            // of consumed bytes alone is not sufficient: the final byte
-                            // of a five-byte arc can still push the value past
-                            // `u32::MAX`.
+                            // A five byte arc can still exceed `Arc`, so the digits are
+                            // accumulated with overflow checking rather than by counting
+                            // bytes.
                             result = match result
                                 .checked_mul(0x80)
                                 .and_then(|result| result.checked_add((byte & 0b0111_1111) as Arc))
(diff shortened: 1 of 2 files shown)


const-oid/src/arcs.rs
```diff
@@ -72,27 +63,34 @@ impl<'a> Arcs<'a> {
...
-                            if byte & 0b10000000 == 0 {
+                            // Accumulate the base 128 digits with overflow checking.
+                            //
+                            // An arc whose value does not fit in `Arc` (`u32`) is
+                            // rejected as `ArcTooBig` rather than being silently
+                            // truncated by the `<< 7` shift. Note that a `u32` arc can
+                            // be up to five base 128 bytes long, so checking the number
+                            // of consumed bytes alone is not sufficient: the final byte
+                            // of a five-byte arc can still push the value past
+                            // `u32::MAX`.
```
tarcieri: This comment seems overly verbose and also references the previous bug which is somewhat irrelevant after it's fixed.

Can you whittle it down to something shorter? And no need to mention the truncation bug in a fixed implementation.

It can stay on the comment on the tests, since those are talking about desirable or undesirable behavior.

tarcieri:
Nice catch, thanks!

const-oid/src/arcs.rs
jaideeppyne: Cut down to three lines and the truncation reference is gone:

```rust
// A five byte arc can still exceed `Arc`, so the digits are
// accumulated with overflow checking rather than by counting
// bytes.
```

Left the test comments as they are, since those are describing the behaviour being asserted.

jaideeppyne:
Thanks!

PR #2415, opened 2026-08-29
der: add `AsIntRef` and `AsUintRef` traits

dishmaker:
Needed for `pkcs1` split: `Ref` and `Owned` types.

- #2411

### DIFF
--- a/der/src/asn1/integer.rs
+++ b/der/src/asn1/integer.rs
@@ -3,6 +3,9 @@
 pub(super) mod int;
 pub(super) mod uint;
 
+use int::IntRef;
+use uint::UintRef;
+
 use core::{cmp::Ordering, mem::size_of};
 
 use crate::{EncodeValue, Result, encode::encode_value_to_slice};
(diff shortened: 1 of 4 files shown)


der/src/asn1/integer.rs
```diff
@@ -30,6 +35,22 @@ where
...
+    #[must_use]
+    fn as_int_bytes(&self) -> &[u8];
+}
+
+/// Borrow the inner byte slice of [`UintRef`], or an owned `Uint`
+pub trait UintLike {
+    /// Borrow the inner byte slice which contains the least significant bytes
+    /// of a big endian integer value with all leading zeros stripped.
+    #[must_use]
+    fn as_uint_bytes(&self) -> &[u8];
```
tarcieri: I was suggesting this be:

```suggestion
    fn as_uint_ref<'a>(&'a self) -> &UintRef<'a>;
```

tarcieri: And with that as the method, perhaps `AsUintRef` might be a better name, especially with an eventual goal of it just being `AsRef<UintRef>`

dishmaker: Ah, ok

tarcieri:
This looks generally fine to me now if you want to remove draft

